PR #21, opened 2025-02-11
Add test execution timeouts by overriding `Minitest::Spec#run`

marcoroth:
This seems a little bit excessive, but I haven't found another way to make sure that Ruby can kill the process if the Lexer in the C-Library is running into an infinite loop.


marcoroth:
Also, leaving this here for context: https://elk.zone/ruby.social/@zenspider/113998994541257968

PR #109, opened 2025-06-06
Upgrade Ruby to `3.4.5`

marcoroth:
Bumps the locally used Ruby version to `3.4.4`.


marcoroth:
See https://github.com/rake-compiler/rake-compiler-dock/pull/156

marcoroth:
Also, see https://github.com/rake-compiler/rake-compiler-dock/issues/162

Issue #103, opened 2025-05-23
Herb sometimes hangs when parsing an HTML file


spect88:
Hi Marco!

When I try to parse [the default 404 page in Rails](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/public/404.html), herb works most of the time, but sometimes it hangs (stuck at 100% CPU and I have to kill it)

To reproduce:
```sh
# Try this a couple more times if it exits properly
for i in {1..10}; do herb parse 404.html > /dev/null; done
```

I use `herb (0.1.1-arm64-darwin)` on `ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [arm64-darwin23]`.

I've also seen an occasional segfault (when parsing the same file).

Another thing, which might or might not be related is that when parsing the same HTML file using the Ruby API, I _sometimes_ get empty `<style>` tag:
```
    │   │   │   │   ├── @ HTMLElementNode (location: (14:4)-(96:12))
    │   │   │   │   │   ├── open_tag:
    │   │   │   │   │   │   └── @ HTMLOpenTagNode (location: (14:4)-(14:11))
    │   │   │   │   │   │       ├── tag_opening: "<" (location: (14:4)-(14:5))
    │   │   │   │   │   │       ├── tag_name: "style" (location: (14:5)-(14:10))
    │   │   │   │   │   │       ├── tag_closing: ">" (location: (14:10)-(14:11))
    │   │   │   │   │   │       ├── children: []
    │   │   │   │   │   │       └── is_void: false
    │   │   │   │   │   │
    │   │   │   │   │   ├── tag_name: "style" (location: (14:5)-(14:10))
    │   │   │   │   │   ├── body: (1 item)
    │   │   │   │   │   │   └── @ HTMLTextNode (location: (14:11)-(96:4))
    │   │   │   │   │   │       └── content: ""
    │   │   │   │   │   │
    │   │   │   │   │   ├── close_tag:
    │   │   │   │   │   │   └── @ HTMLCloseTagNode (location: (96:4)-(96:12))
    │   │   │   │   │   │       ├── tag_opening: "</" (location: (96:4)-(96:6))
    │   │   │   │   │   │       ├── tag_name: "style" (location: (96:6)-(96:11))
    │   │   │   │   │   │       └── tag_closing: ">" (location: (96:11)-(96:12))
    │   │   │   │   │   │
    │   │   │   │   │   └── is_void: false
```

[...]

PR #118, opened 2025-06-08
Increase initial buffer capacity to 8192

marcoroth:
This pull request updates the initial capacity of the `buffer_T` structure from `1024` to `8192`. This helps with the symptoms that surfaced in #103, but obviously doesn't solve them ultimately, just makes them less likely to show up.


marcoroth:
Closing in favor of #122

PR #205, opened 2025-07-01
Linter: Implement `erb-no-output-control-flow` rule

### LINKED ISSUE #126: Linter Rule: Disallow output ERB tags with control flow
##### Description

Disallow using output ERB tags (`<%=`) for control flow statements like `if`, `unless`, `case`, `while`, etc. Control flow should be written with regular ERB tags (`<% ... %>`), since these do not produce output directly.

##### Rationale

Using `<%=` with control flow is typically a mistake or misunderstanding of ERB behavior. Output tags (`<%=`) are designed to render values into the HTML output, while control flow statements only affect execution and do not produce a value to render. This misuse can result in unexpected output, unnecessary blank spaces, or subtle bugs.

Reporting this as a warning can help developers catch likely mistakes while allowing flexibility for rare advanced cases.

###### ✅ Good

```html+erb
<% if condition %>
  Content here
<% end %>

<%= user.name %>
```

###### 🚫 Bad

```html+erb
<%= if condition %>
  Content here
<% end %>

<%= unless user.nil? %>
  Welcome!
<% end %>
```

#### References

* [Inspiration](https://x.com/specialcasedev/status/1935013470069719231)

written-fresh:
ERB no output control flow disallows rendering if/else/elsif/end blocks

In the future we could expand this rule to also catch `for`/`while`/`until` and maybe more.

Closes https://github.com/marcoroth/herb/issues/126


javascript/packages/linter/src/rules/erb-no-output-control-flow.ts
```diff
@@ -0,0 +1,70 @@
+import { Visitor, Node, ERBIfNode, ERBUnlessNode, ERBElseNode, ERBEndNode } from "@herb-tools/core"
+import { Rule, LintMessage } from "../types.js"
```
marcoroth: I personally like to keep the regular types separate from the `type` imports

```suggestion
import { Visitor } from "@herb-tools/core"

import type { Rule, LintMessage } from "../types.js"
import type { Node, ERBIfNode, ERBUnlessNode, ERBElseNode, ERBEndNode } from "@herb-tools/core"
```

javascript/packages/linter/src/rules/erb-no-output-control-flow.ts
```diff
@@ -0,0 +1,70 @@
...
+        severity: "error"
+      })
+    }
+
+    return
+  }
+
+}
+
+export class ErbNoOutputControlFlow implements Rule {
```
marcoroth: Can we uppercase `ERB`?

```suggestion
export class ERBNoOutputControlFlow implements Rule {
```

javascript/packages/linter/src/rules/erb-no-output-control-flow.ts
```diff
@@ -0,0 +1,70 @@
...
+  name = "erb-no-output-control-flow"
+  description = "Prevent block-level elements from being placed inside inline elements"
+
+  check(node: Node): LintMessage[] {
+    const visitor = new NoOutputControlFlow(this.name)
+    visitor.visit(node)
+    return visitor.getMessages()
+  }
+}
```
marcoroth: Can you make sure there is a newline at the end of the file, thanks!
```suggestion
}

```

marcoroth (commented):
Hey @drewhoffer, thanks for this pull request! 

Just a first pass-through that I saw when looking over it 🙌🏼

javascript/packages/linter/src/rules/erb-no-output-control-flow.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  visitERBEndNode(node: ERBEndNode): void {
+    this.checkOutputControlFlow(node)
+    this.visitChildNodes(node)
+  }
+  
+  getMessages(): LintMessage[] {
+    return this.messages
+  }
```
marcoroth: I think we can just remove this one, we removed the interface for this in #203

javascript/packages/linter/src/rules/erb-no-output-control-flow.ts
```diff
@@ -0,0 +1,66 @@
...
+      })
+    }
+    return
+  }
+
+}
+
+export class ERBNoOutputControlFlow implements Rule {
+  name = "erb-no-output-control-flow"
+  description = "Prevent block-level elements from being placed inside inline elements"
```
marcoroth: we also removed this description in #203

marcoroth (approved):
Awesome, thank you so much @drewhoffer! 🙌🏼

Issue #209, opened 2025-07-01
Linter Rule: `role="heading"` elements must include `aria-level`


marcoroth:
##### Description

Ensure that any element with `role="heading"` also has a valid `aria-level` attribute. The `aria-level` defines the heading level (1–6) and is required for assistive technologies to properly interpret the document structure.

##### Rationale

In HTML, semantic heading elements like `<h1>` through `<h6>` implicitly define their level. When using `role="heading"` on non-semantic elements (e.g., `<div>`, `<span>`), the level must be explicitly declared using `aria-level`, otherwise screen readers and accessibility tools may not understand the document hierarchy.

##### ✅ Good

```html+erb
<div role="heading" aria-level="2">Section Title</div>

<span role="heading" aria-level="1">Main Title</span>
```

##### 🚫 Bad

```html+erb
<div role="heading">Section Title</div>

<span role="heading">Main Title</span>
```

#### References

- [ARIA: `heading` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/heading_role)
- [ARIA: `aria-level` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-level)

written-fresh:
Hey can i work on this?

PR #219, opened 2025-07-01
Implement `html-aria-role-heading-requires-level` lint rule

drewhoffer:
- add linter rule that ensures a node with `role="heading"` also requires a aria-level
- this pr simply checks for the **presence** of `aria-level` not whether it is valid or not

Closes: #209


marcoroth (approved):
Thanks @drewhoffer!

PR #220, opened 2025-07-01
Add `html-anchor-require-href` linter rule

### LINKED ISSUE #172: Linter Rule: Disallow Link without href
##### Description

Disallow the use of anchor tags without an`href` attribute in HTML templates. Use <button> if you want to perform an action without having the user navigated to a new URL.

##### Rationale

Anchor tags without `href` are unfocusable if user is using keyboard navigation, or is unseen by screen readers. 

##### ✅ Good

```html+erb
<a href="https://alink.com">I'm a real link</a>
```

##### 🚫 Bad

```html+erb
<a data-action="click->doSomething">I'm a fake link</a>
```

#### References

Inspired by [
ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/link-href-attributes.md)

NotGrm:
Implements a new linter rule to detect `<a>` tags without `href` attributes

Closes #172


javascript/packages/linter/src/rules/html-anchor-require-href.ts
```diff
@@ -0,0 +1,41 @@
...
+        "error",
+      )
+    }
+  }
+}
+
+export class HTMLAnchorRequireHrefRule implements Rule {
+  name = "html-anchor-require-href"
+  description =
+    "Disallow the use of anchor tags without anhref attribute in HTML templates."
```
marcoroth: You can remove the description here, since it's not used
```suggestion
```

javascript/packages/linter/src/rules/html-anchor-require-href.ts
```diff
@@ -0,0 +1,41 @@
...
+  private checkATag(node: HTMLOpenTagNode): void {
+    const tagName = getTagName(node)
+
+    if (tagName !== "a") {
+      return
+    }
+
+    if (!hasAttribute(node, "href")) {
+      this.addMessage(
+        "Missing required `href` attribute on `<a>` tag.",
```
marcoroth: How do you feel about making the message more actionable?
```suggestion
        "Add an `href` attribute to `<a>` to ensure it is focusable and accessible.",
```

NotGrm: Very good idea 👍

marcoroth:
Okay, now that the CI is accurate, it looks like the test failures are legit. I think we might have to update the way we run these snapshots, because they are failing now because our linter got smarter 😅

NotGrm:
I forgot to update my tests after changing error message. It is done now and I see that all rules test are okay; but as you said snapshot is outdated 😅

marcoroth:
Would you mind updating them (using `yarn test -u`)

NotGrm:
> Would you mind updating them (using `yarn test -u`)

No problem

NotGrm:
Snapshots are updated, but locally I still have [one test](https://github.com/marcoroth/herb/blob/main/javascript/packages/linter/test/cli.test.ts#L39) that is failing.

It seems to be due to this line, but I didn't manage to fix it. I need to figure out the CLI logic
https://github.com/marcoroth/herb/blob/main/javascript/packages/linter/test/cli.test.ts#L21

marcoroth:
@NotGrm it looked like you updated the tests using an older version of the linter CLI. Maybe you need to run `yarn build` once?

Either way, I just updated them 👍🏼

NotGrm:
> @NotGrm it looked like you updated the tests using an older version of the linter CLI. Maybe you need to run `yarn build` once?
> 
> Either way, I just updated them 👍🏼

I didn't thought about rebuilding the linter CLI, I will try to remember it for next contribution. Thanks a lot

marcoroth (approved):
Thanks for contributing this rule @NotGrm! 🙌🏼

PR #221, opened 2025-07-02
Add NX affected back

marcoroth:
After removing it in #132 (but setting the whole NX workspace up for it), this pull request enables it again.


marcoroth:
We'll see how it goes and if it works out in practice

Issue #211, opened 2025-07-01
Linter Rule: Only use supported `aria-*` attributes


marcoroth:
##### Description

Disallow unknown or invalid `aria-*` attributes. Only attributes defined in the WAI-ARIA specification should be used. This rule helps catch typos (e.g. `aria-lable`), misuse, or outdated attribute names that won't be interpreted by assistive technologies.

##### Rationale

ARIA attributes are powerful accessibility tools, but **only if used correctly**. Mistyped or unsupported attributes:

- Are silently ignored by browsers and screen readers
- Fail to communicate intent
- Give a false sense of accessibility

Validating against a known list ensures you're using correct and effective ARIA patterns.

##### ✅ Good

```html+erb
<div role="button" aria-pressed="false">Toggle</div>
<span role="heading" aria-level="2">Title</span>
<input type="text" aria-label="Search" />
```

##### 🚫 Bad

```html+erb
<div role="button" aria-presed="false">Toggle</div>       <!-- typo -->
<span role="heading" aria-size="large">Title</span>       <!-- invalid -->
<input type="text" aria-lable="Search" />                 <!-- typo -->

```

#### References

- [ARIA states and properties (attributes)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes) 
- [NPM Package: `aria-attributes`](https://github.com/wooorm/aria-attributes)

written-fresh:
Can i take this?

PR #242, opened 2025-07-09
Add `html-aria-attribute-must-be-valid` linter rule

written-fresh:
- only allow valid aria attributes
- add missing anchor tags require href docs link

Closes Linter Rule: Only use supported `aria-*` attributes #211

### DIFF
--- a/javascript/packages/linter/docs/rules/html-aria-attribute-must-be-valid.md
+++ b/javascript/packages/linter/docs/rules/html-aria-attribute-must-be-valid.md
@@ -1,4 +1,4 @@
-# Linter Rule: Disallow output ERB tags with control flow
+# Linter Rule: Aria Attribute must be valid
 
 **Rule:** `aria-attribute-must-be-valid`
 
--- a/javascript/packages/linter/src/rules/html-aria-attribute-must-be-valid.ts
+++ b/javascript/packages/linter/src/rules/html-aria-attribute-must-be-valid.ts
@@ -18,17 +18,16 @@ class AriaAttributeMustBeValid extends AttributeVisitorMixin {
     attributeNode: HTMLAttributeNode,
     _parentNode: HTMLOpenTagNode | HTMLSelfCloseTagNode,
   ): void {
-    if (!attributeName.includes("aria-") return;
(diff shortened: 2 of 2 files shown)


javascript/packages/linter/src/rules/html-aria-attribute-must-be-valid.ts
```diff
@@ -0,0 +1,43 @@
...
+} from "@herb-tools/core";
+
+class AriaAttributeMustBeValid extends AttributeVisitorMixin {
+  checkAttribute(
+    attributeName: string,
+    _attributeValue: string | null,
+    attributeNode: HTMLAttributeNode,
+    _parentNode: HTMLOpenTagNode | HTMLSelfCloseTagNode,
+  ): void {
+    if (attributeName.includes("aria-")) {
```
written-fresh: Wondering if this should be `.startsWith` :thinking: ...

javascript/packages/linter/docs/rules/html-aria-attribute-must-be-valid.md
```diff
@@ -0,0 +1,48 @@
+# Linter Rule: Disallow output ERB tags with control flow
```
NotGrm: Title do not match here

javascript/packages/linter/src/rules/html-aria-attribute-must-be-valid.ts
NotGrm: What do you think of returning early instead of nesting conditionals?

```suggestion
    if (!attributeName.includes("aria-") return;
```

written-fresh: 100% always forget to clean these up! Good catch

javascript/packages/linter/docs/rules/html-aria-attribute-must-be-valid.md
written-fresh: I gotta start using that generator @marcoroth made just for this! Thanks for catching

marcoroth (approved):
Thank you, @drewhoffer! 🙌🏼

PR #263, opened 2025-07-15
Formatter: Ensure CLI output ends with trailing newline

marcoroth:
Files formatted by the CLI now properly end with a newline for better file formatting standards while keeping the formatter API unchanged.


brunoprietog:
This is also happening to me in VS Code, actually. I don't know if you use the cli under the hood? Every time I format, the formatter deletes the line at the end of the file.

marcoroth:
good point @brunoprietog, thanks for spotting that! The extension doesn't use the CLI directly, but I added the same logic to the formatting service in the language server!

Issue #280, opened 2025-07-20
Conditional HTML attributes incorrectly removed


tmaier:
**Problem:** The formatter removes conditional ERB code for HTML attributes, breaking the HTML structure.

**Original code:**
```erb
<dialog
  class="relative z-30"
  aria-describedby="<%= stimulus_identifier %>-body"
  <% if title %>
  aria-labelledby="<%= stimulus_identifier %>-title"
  <% end %>
  aria-modal="true"
  data-controller="<%= stimulus_identifier %>"
>
  test
</dialog>
```

**Formatted result:**
```erb
<dialog
  class="relative z-30"
  aria-describedby="<%= stimulus_identifier %>-body"
  aria-modal="true"
  data-controller="<%= stimulus_identifier %>"
>
  test
</dialog>
```

**Impact:** The conditional `aria-labelledby` attribute is completely removed, breaking accessibility functionality.

**Versions:**

```
yarn herb-formatter --version
Versions:
  @herb-tools/formatter@0.4.0
  @herb-tools/node-wasm@0.4.0
  @herb-tools/core@0.4.0
  libprism@1.4.0
  libherb@0.4.0 (WebAssembly)
```

marcoroth:
Thanks for reporting @tmaier! I just opened #286 to address this 🙌🏼

PR #286, opened 2025-07-22
Formatter: Preserve ERB tags inside HTML attributes

marcoroth:
Resolves https://github.com/marcoroth/herb/issues/280


tmaier:
Thank you!

PR #293, opened 2025-07-23
Fix how to instantiate specific rules in the Javascript customization example

sbastn:
This PR corrects how rules are passed to the Linter in the JavaScript customization example.


marcoroth (approved):
Thanks for spotting this, @sbastn! 🙏🏼

PR #298, opened 2025-07-25
Add a plugin for JetBrains IDEs

mwnciau:
This PR merges https://github.com/mwnciau/herb-lsp-intellij into this project in the `jetbrains/` folder.

I have changed the README to resemble the VSCode plugin, and moved the github actions to the root project (although I hadn't used them in the original repo).

This is a WIP, and there are a number of minor issues that probably need to be addressed, but I'm not sure how much time I have to work on this further. A few that I know about:

* HTML works when viewing errors in the "problems" window in the IDE, but HTML disappears (i.e. it's being parsed as HTML) when hovering over elements with problems in the IDE. I suspect this may be a bug with the JetBrains implementation of LSP.
* Currently, the plugin requires Node.js to be installed. If it cannot find Node.js, it fails silently. Ideally, there would be a prompt to install/setup Node.js like other functionality in the IDE, plus support for remote interpreters.


mwnciau:
The plugin has also been published to https://plugins.jetbrains.com/plugin/27954-herb, and I'm happy to hand control of this over.

mwnciau:
I've hopefully fixed the build task now, and the release task should also work once the relevant tokens are added to the repository.

I've also raised an issue with JetBrains about the HTML parsing problem when you hover over an error: https://youtrack.jetbrains.com/issue/IJSDK-2604/Inconsistent-parsing-of-HTML-for-LSP

marcoroth:
Hey @mwnciau, sorry for the delay here! Thanks for working on this!

I'm gonna check if can run this locally

mwnciau:
Any luck getting this running? It's been a while since I've had to setup an IDE from scratch and it can be a bit finicky.

I've managed to get the build github action working with the code in a folder. A few more paths updated, and the action requires a "QODANA_TOKEN" github secret (see the latest commit - it uses the "Github Actions" environment). You can get a key from https://qodana.cloud. I've set the action to continue if this step fails for now.

mickeytgl:
@mwnciau Hey thanks for this Plugin! Quick question, are you able to run this in Rubymine 2025.2 ? I had to install another plugin (LSP4IJ) and add it manually as a language server to get it to work. See below:

CleanShot 2025-08-23 at 17 35 01

Without that, I see only that "The Herb language server is running on other files" even if I'm in an `.html.erb` or `.erb` file :/ . Let me know if I'm missing anything

PR #320, opened 2025-07-31
Improve range formatting to preserve minimum indentation

marcoroth:
This pull request improves the range formatting functionality introduced in #319 by calculating the minimum indentation across all selected lines, then stripping the minimum indentation before formatting (to prevent over-indenting) and then re-applying the minimum indentation to all formatted lines so it stays in-place and doesn't get dedented all the way to the beginning of the line.

**Before:**

https://github.com/user-attachments/assets/d5aea24d-9090-447c-aff8-a3ae17f767b6

**After:**

https://github.com/user-attachments/assets/21001f36-f337-49fa-847e-2f59cbd1bfc1

/cc @excid3

### DIFF
--- a/javascript/packages/language-server/src/formatting_service.ts
+++ b/javascript/packages/language-server/src/formatting_service.ts
@@ -160,24 +160,47 @@ export class FormattingService {
       const formatter = new Formatter(this.project.herbBackend, options)
 
       const rangeText = document.getText(params.range)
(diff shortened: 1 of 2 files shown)


excid3:
That looks like it handles it nicely!

marcoroth:
Yeah, I think this is definitely more predictable compared to what it was doing before 🙌🏼

PR #383, opened 2025-08-15
Linter: Implement `html-no-duplicate-meta-names` linter rule

### LINKED ISSUE #381: Linter Rule: Duplicate `<meta>` name attributes are not allowed
**Rule:** `html-meta-name-must-be-unique`

## Description

Warn when multiple `<meta>` tags share the same `name` or `http-equiv` attribute within the same `<head>` block, unless they are wrapped in conditional comments.

## Rationale

In HTML, duplicate `<meta>` tags with the same `name` or `http-equiv` can cause unexpected behavior. For example, search engines or social sharing tools may use the wrong value. These duplicates are often the result of copy-paste or partial rendering logic and should be avoided for clarity and correctness.

### ✅ Good

```erb
<meta name="description" content="Welcome to our site">
<meta name="viewport" content="width=device-width, initial-scale=1.0">
```

```erb
<% if mobile? %>
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
<% else %>
  <meta name="viewport" content="width=1024">
<% end %>
```

### 🚫 Bad

```erb
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<meta name="viewport" content="width=1024">
```

```erb
<meta http-equiv="X-UA-Compatible" content="IE=edge">
<meta http-equiv="X-UA-Compatible" content="chrome=1">
```

## References

\-

marcoroth:
Resolves #381


pardeyke:
I believe there is one exception to this:

```
<meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)">
<meta name="theme-color" content="#212529" media="(prefers-color-scheme: dark)">
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/theme-color

PR #386, opened 2025-08-18
Use GitHub Actions runner

marcoroth:
This pull request is a test to see if this could be an alternative to #385 which seems to solve the issue.

This reverts commit cd79de93cbb4019afa60a31f07e9e0a59c60cfc6.


marcoroth:
Okay looks like that's not the issue, closing in favor of #385

PR #396, opened 2025-08-20
Add Parser Option to track insignificant whitespace

marcoroth:
This pull request introduces a `track_whitespace` parser options that allows to track report back insignificant white as `WhitespaceNode`'s in the syntax tree.

Having the whitespace nodes available in the syntax tree allows to parse documents without any loss of information, so we can do source -> AST -> source translations and preserve every detail.

This makes it ideal for AST-level rewrites (i.e for auto-fixes in the linter) without reformatting anything else in the document.


marcoroth:
This should allow https://github.com/marcoroth/herb/issues/397

PR #414, opened 2025-08-22
Linter: Add 7 new accessibility rules inspired by `erblint-github`

marcoroth:
This pull request adds the following 7 new linter rules inspired by [`github/erblint-github`](https://github.com/github/erblint-github):

- `html-aria-label-is-well-formatted` - `aria-label` must be well-formatted
- `html-avoid-both-disabled-and-aria-disabled` - Avoid using both `disabled` and `aria-disabled` attributes
- `html-iframe-has-title` - `iframe` elements must have a `title` attribute
- `html-navigation-has-label` - Navigation landmarks must have accessible labels
- `html-no-aria-hidden-on-focusable` - Focusable elements should not have `aria-hidden="true"`
- `html-no-positive-tab-index` - Avoid positive `tabindex` values
- `html-no-title-attribute` - Avoid using the `title` attribute

These rules should help enforce accessibility best practices and help developers create more inclusive web applications.


brunoprietog:
Hi @marcoroth!

I really appreciate you taking accessibility so seriously from the beginning! ❤️

I just wanted to add my two cents here. I think these two rules are a bit aggressive:

1. html-navigation-has-label: If there's only one nav on the entire page, it's not necessary to add an aria-label, as this becomes redundant. Marking this as an error doesn't seem appropriate to me. Redundancy is also not good for accessibility, as it could cause confusion or cause a screen reader to repeat certain information more than once.
2. html-no-title-attribute: While the title attribute is not **always** accessible, I don't think we should avoid it entirely. You can achieve something similar using aria and elements that are only visible to screen readers. So, you could have the title attribute and something with sr-only at the same time, while still being perfectly accessible. Using the title attribute doesn't imply that something is inaccessible at all. In fact, the title attribute is accessible by default when you put it in a focusable control.

Thanks!

KonnorRogers:
> html-navigation-has-label: If there's only one nav on the entire page, it's not necessary to add an aria-label, as this becomes redundant. Marking this as an error doesn't seem appropriate to me. Redundancy is also not good for accessibility, as it could cause confusion or cause a screen reader to repeat certain information more than once.

Agreed. _most `<nav>` elements implicitly are understood as site navigation.

> html-no-title-attribute: While the title attribute is not always accessible, I don't think we should avoid it entirely. You can achieve something similar using aria and elements that are only visible to screen readers. So, you could have the title attribute and something with sr-only at the same time, while still being perfectly accessible. Using the title attribute doesn't imply that something is inaccessible at all. In fact, the title attribute is accessible by default when you put it in a focusable control.

I agree with this as well. `title` attributes are one of those ones I think is hard to have a good rule for because its hard to know the author intent. Is it meant to be a tooltip? Or is it meant to add extra descriptive information?

> In fact, the title attribute is accessible by default when you put it in a focusable control.

Is this true? Last I remember the title attribute was very inconsistent across browser + screen reader combos.

Where some browsers would concatenate it, some will use it as a label, and some don't read it at all. Its possible my info is outdated though.

marcoroth:
Hey @brunoprietog, thanks for the very valuable feedback, this is very appreciated and definitely something we can incorporate.

I want to take accessibility in Herb seriously, especially since I saw that a lot of Rails developers aren't too versed on the frontend, and accessibility especially. So having a tool, like Herb, available in your editor while writing code gives these concerns more visibility or even makes people aware of these concerns.

The Linter, in general, is super aggressive on all fronts right now. But that's on purpose so it's easier to see pain points and where rules are annoying/wrong while we are figuring out the right infrastructure and ways to configure/tune rules.

We also have all rules enabled by default for now and have all of them report their offenses as errors. All of that to say that it's unbalanced on purpose for now.

In the future I think it would still make sense to have these rules enabled but maybe have them report an "info" level diagnostic so that it's just acting as a "hint" for "something to think about" instead of failing the build. Unless it's certain that there is more than one `<nav>`  on the page or that a element with a `[title]` is also backed by a `.sr-only` + `[aria-*]` attribute.

brunoprietog:
> Is this true? Last I remember the title attribute was very inconsistent across browser + screen reader combos.

Well, right you're correct @KonnorRogers, it's not 100% consistent, especially on mobile devices. Personally, I use the title attribute but in combination with an sr-only element.

Thanks, @marcoroth. I mention this because, as a blind person myself, I don't make something inaccessible because I need to read it myself, and with this PR I have accessibility errors, when in reality they are not errors and are working perfectly.

I really appreciate most of the rules, but we have to make sure we apply them correctly, otherwise we could even cause the opposite effect.

For example, while the title attribute may not be accessible on its own in all cases, it is better to use the attribute than to reinvent the functionality in an inaccessible way that never works, for example. I am a bit pragmatic in that sense as well.

I mean, we shouldn't set rules that only apply in certain cases. Because in cases where they don't apply, the result is worse.

In my case, if I wanted to avoid these errors, I would have to add an aria-label to my nav, which is less than ideal , because it adds extra noise to my screen reader that I don't want to have, and I would have to reimplement all the title attributes, even though I'm using sr-only on the child element. The only other option to avoid errors is to uninstall herb, which would be super sad 😢

marcoroth:
@brunoprietog I opened https://github.com/marcoroth/herb/pull/456 to remove these rules from the default rules. We might be able to revisit them as soon as we have a better way of configuring linter rules. Thanks for the feedback, Bruno!

Issue #313, opened 2025-07-29
Linter Rule: Prevent _path helpers to be used inside mailers


MatheusRich:
**Rule:** `actionmailer-no-path-route-helpers`

## Description

Prevent _path helpers to be used inside mailers.

## Rationale

Mailers can't use *_path route helpers and need to use the _url versions instead. I'm unsure if this is something that herb should tackle directly, as it is Rails-specific. If this is not the right project for this, LMK. Maybe this should live on `erb-lint`?

### ✅ Good

```erb
<%= link_to "Go to profile", profile_url(@profile) %>
```

### 🚫 Bad

```erb
<%= link_to "Go to profile", profile_path(@profile) %>
```

marcoroth:
Thanks for suggesting this rule @MatheusRich!

I'd say that this is totally valid to have in Herb. But I guess we would need a feature that scopes certain rules to only certain directories/files/globs.

marcoroth:
I guess we could do something similar to what RuboCop does with `Include` in the default config:

```yml
Bundler/DuplicatedGem:
  Description: 'Checks for duplicate gem entries in Gemfile.'
  Enabled: true
  Severity: warning
  VersionAdded: '0.46'
  VersionChanged: '1.40'
  Include:
    - '**/*.gemfile'
    - '**/Gemfile'
    - '**/gems.rb'
```

https://github.com/rubocop/rubocop/blob/3432c6d939312cb88ad3e18226f8e7a20214bbdb/config/default.yml#L181-L190

marcoroth:
I implemented a new `isEnabled` API in #430 that we could use for this rule to figure out if it should be enabled or not.

PR #430, opened 2025-08-24
Linter: Add `isEnabled` hook for `LinterRule`

marcoroth:


marcoroth:
This can also be useful for #313

PR #446, opened 2025-08-26
Add `html-no-underscores-in-attribute-names` rule to linter

### LINKED ISSUE #405: Linter Rule: HTML attribute names must not contain underscores
**Rule:** `html-no-underscores-in-attribute-names`

## Description

Warn when an HTML attribute name contains an underscore (`_`). According to the HTML specification, attribute names should use only lowercase letters, digits, hyphens (`-`), and colons (`:`) in specific namespaces (e.g., `xlink:href` in SVG). Underscores are not valid in standard HTML attribute names and may lead to unpredictable behavior or be ignored by browsers entirely.

## Rationale

Underscores in attribute names violate the HTML specification and are not supported in standard markup. Their use is almost always accidental (e.g., mistyping `data-attr_name` instead of `data-attr-name`) or stems from inconsistent naming conventions across backend or templating layers.

### ✅ Good

```erb
<div data-user-id="123"></div>
<img aria-label="Close">
```

### 🚫 Bad

```erb
<div data_user_id="123"></div>
<img aria_label="Close">
```

## References

\-

mickeytgl:
Fixes #405 

Introduced a rule to flag HTML attribute names containing underscores and recommend using hyphens instead. Included corresponding tests and updated the default rules list.


marcoroth (commented):
Hey @mickeytgl, thank you so much for your contribution! This looks great, thank you for giving it a shot!

I recently added a new [`AttributeVisitorMixin`](https://github.com/marcoroth/herb/blob/80e708b87b183fe7eec3af1e8cb0ea32ff871c63/javascript/packages/linter/src/rules/rule-utils.ts#L622) that could allow to simplify the rule but also help to more strategically visit all kind of attribute name setups, since it also handles static/dynamic attribute names.

Sadly I haven't gotten to update all existing rules to use it, but I think it would be really good for this rule to make use of it. Here's an example: [`html-attribute-double-quotes`](https://github.com/marcoroth/herb/blob/main/javascript/packages/linter/src/rules/html-attribute-double-quotes.ts).

For this rule we could do something like:

```typescript
import { getStaticAttributeName, getStaticContentFromNodes } from "@herb-tools/core"
import { AttributeVisitorMixin } from "./rule-utils.js"

import type { StaticAttributeStaticValueParams, StaticAttributeDynamicValueParams, DynamicAttributeStaticValueParams, DynamicAttributeDynamicValueParams } from "./rule-utils.js"

class HTMLNoUnderscoresInAttributeNamesVisitor extends AttributeVisitorMixin {
  // <div data-class="value"></div>
  protected checkStaticAttributeStaticValue({ attributeName }: StaticAttributeStaticValueParams) {
    // attributeName => "data-class"
  }

  // <div data-class="<%= class %>"></div>
  protected checkStaticAttributeDynamicValue({ attributeName }: StaticAttributeDynamicValueParams) {
    // attributeName => "data-class"
  }

  // <div data-<%= dynamic %>="class"></div>
  protected checkDynamicAttributeStaticValue({ nameNodes }: DynamicAttributeStaticValueParams) {
    const attributeName = getStaticContentFromNodes(nameNodes)
    // attributeName => "data-"
  }

  // <div data-<%= dynamic %>="<%= dynamic %>"></div>
  protected checkDynamicAttributeDynamicValue({ nameNodes }: DynamicAttributeDynamicValueParams) {
    const attributeName = getStaticContentFromNodes(nameNodes)
    // attributeName => "data-"
  }
}
```

That would also allow to account for cases like:

```html
<div data_<%= dynamic %>="value"></div>
<div data_<%= dynamic %>_attribute="value"></div>
<div <%= dynamic %>_attribute="value"></div>
```

Thank you!

javascript/packages/linter/test/rules/html-no-underscores-in-attribute-names.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const html = dedent`
+      <div data-<%= "invalid_name"  %>="value"></div>
+      <div <%= "dynamic_name" %>="value"></div>
+      <div data-<%= "dynamic_name" %>-test="value"></div>
+    `
+
+    const linter = new Linter(Herb, [HTMLNoUnderscoresInAttributeNamesRule])
+    const lintResult = linter.lint(html)
+
+    // expect(lintResult.warnings).toBe(3)
```
mickeytgl: @marcoroth I'm having issues making this expectation to pass. I'm not sure if there is a way to actually parse the ERB code here or if it's a current limitation. 

I see other examples testing `attribute-<%= key %>`, but not sure how to test examples where `key` returns a string containing an underscore.

marcoroth: Hey @mickeytgl, this is a current limitation. We currently don't have a way to evaluate static strings from ERB, but that might come in the future.

For now these should pass, so this can be `.toBe(0)`.


We should catch this though:
```html
      <div data_<%= key %>="value"></div>
      <div data_<%= key %>-test="value"></div>
      <div data-<%= key %>_test="value"></div>
```

Thank you!

javascript/packages/linter/src/rules/html-no-underscores-in-attribute-names.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    this.check(attributeName, attributeNode)
+  }
+
+  private check(attributeName: string | null, attributeNode: HTMLAttributeNode): void {
+    if (!attributeName) return
+
+    if (attributeName.includes("_")) {
+      this.addOffense(
+        `Attribute \`${attributeName}\` should not contain underscores. Use hyphens (-) instead.`,
```
marcoroth: I think if you `import { IdentityPrinter } from "@herb-tools/printer"` and update the message to: 

```suggestion
        `Attribute \`${IdentityPrinter.print(attributeNode.name}\` should not contain underscores. Use hyphens (-) instead.`,
```

It will update the message to:
```
"Attribute `data_<%= key %>-test` should not contain underscores. Use hyphens (-) instead."
```

mickeytgl: Thanks! Yeah that is definitely cleaner. Updated the file 👍🏻

mickeytgl:
@marcoroth I see some failing tests on CI, but I think it's just some flakiness? 🤔 Let me know if I missed something though

marcoroth:
@mickeytgl I think you just need to run `yarn build` to build the CLI and then run `yarn test -u` to update the snapshots in `javascript/packages/linter/`

marcoroth (approved):
Thank you for this contribution, @mickeytgl! 🙏🏼

PR #456, opened 2025-08-27
Linter: Remove `no-title-attribute` and `navigation-has-label` from default rules

marcoroth:
After some discussion and feedback from @brunoprietog in #414 this pull request removes the `html-no-title-attribute` and `html-navigation-has-label` linter rules from the default rules.

As soon as we have a better way and enabling/disabling and configuring linter rules we might be able to bring them back. But for now,  we keep them disabled as they cause more trouble than they help.


brunoprietog:
Really appreciate it! Thank you so much ❤️

Issue #445, opened 2025-08-25
Linter Rule: Attributes like `id` or `class` must not be empty


marcoroth:
**Rule:** `html-no-empty-attributes`

## Description

Warn when certain attributes, such as `id`, `class`, `name`, `for`,  `src`, and similar, are present but have an empty string as their value. These attributes are required to have meaningful values to function properly, and leaving them empty is typically either a mistake or unnecessary.

In most cases, if the value is not available, it's better to omit the attribute entirely.

## Rationale

Many HTML attributes are only useful when they carry a value. Leaving these attributes empty can:

* Produce confusing or misleading markup (e.g., `id=""`, `class=""`)
* Create inaccessible or invalid HTML (e.g., `alt=""` on decorative images is valid, but only if intentional)
* Interfere with CSS or JS selectors expecting meaningful values
* Indicate unused or unfinished logic in ERB

This rule helps ensure that required attributes are only added when they are populated.

### ✅ Good

```erb
<div id="header"></div>
<img src="/logo.png" alt="Company logo">
<input type="text" name="email">

<!-- if no class should be set, omit it completely -->
<div>Plain div</div>
```

### 🚫 Bad

```erb
<div id=""></div>

<img src="">

<input name="">

<div class="">Plain div</div>
```

## References

\-

PR #460, opened 2025-08-28
Linter: Implement `html-no-empty-attributes` linter rule

auslach:
This fixes https://github.com/marcoroth/herb/issues/445.

Leverages [AttributeVisitorMixin](https://github.com/marcoroth/herb/blob/80e708b87b183fe7eec3af1e8cb0ea32ff871c63/javascript/packages/linter/src/rules/rule-utils.ts#L622) to check static and dynamic attribute names.

Dynamic values are not checked since we cannot determine the ERB output during linting.

@marcoroth let me know if there are any additional html attributes we should add.


javascript/packages/linter/test/rules/html-no-empty-attributes.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    const linter = new Linter(Herb, [HTMLNoEmptyAttributesRule])
+    const lintResult = linter.lint(html)
+
+    expect(lintResult.errors).toBe(0)
+    expect(lintResult.warnings).toBe(1)
+
+    expect(lintResult.offenses[0].message).toBe('Attribute `alt` must not be empty. Either provide a meaningful value or remove the attribute entirely.')
+    expect(lintResult.offenses[0].severity).toBe("warning")
+  })
```
marcoroth: I believe this one conflicts with the [`html-img-require-alt`](https://herb-tools.dev/linter/rules/html-img-require-alt) rule. I guess we have to allow empty `alt=""` for that reason.

javascript/packages/linter/test/rules/html-no-empty-attributes.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  test("passes for dynamic attribute name with ERB value", () => {
+    const html = '<div data-<%= key %>="<%= value %>"></div>'
+
+    const linter = new Linter(Herb, [HTMLNoEmptyAttributesRule])
+    const lintResult = linter.lint(html)
+
+    expect(lintResult.errors).toBe(0)
+    expect(lintResult.warnings).toBe(0)
+  })
```
marcoroth, line 277: It would be nice if we could also catch this:
```ts
  test("fails for attribute with ERB that doesn't output anything", () => {
    const html = '<div class="<% value %>"></div>'

    const linter = new Linter(Herb, [HTMLNoEmptyAttributesRule])
    const lintResult = linter.lint(html)

    expect(lintResult.errors).toBe(0)
    expect(lintResult.warnings).toBe(1)
    expect(lintResult.offenses[0].message).toBe('Attribute `class` must not be empty. Either provide a meaningful value or remove the attribute entirely.')
  })

  test("passes for attribute with static value and ERB that doesn't output anything", () => {
    const html = '<div class="something.<% value %>"></div>'

    const linter = new Linter(Herb, [HTMLNoEmptyAttributesRule])
    const lintResult = linter.lint(html)

    expect(lintResult.errors).toBe(0)
    expect(lintResult.warnings).toBe(0)
  })
```

We have the `isEffectivelyStatic()` and `getValidatableStaticContent()` helpers that might help here. But thinking about this now, I guess the `AttributeVisitorMixin` itself could handle this, so that it automatically calls the right `check*StaticValue` callback.

javascript/packages/linter/src/rules/html-no-empty-attributes.ts
```diff
@@ -0,0 +1,78 @@
...
+  'href',
+  'alt',
+  'title',
+  'data',
+  'role'
+])
+
+// Check if attribute name matches any restricted patterns
+function isRestrictedAttribute(attributeName: string): boolean {
+  const lowerName = attributeName.toLowerCase()
```
marcoroth: I think the `attributeName`  that we pass from the `AttributeVisitorMixin` is already lowercased:

https://github.com/marcoroth/herb/blob/b2029f2c9134f3b025639f1412f358c0fbedcd26/javascript/packages/linter/src/rules/rule-utils.ts#L188

javascript/packages/linter/src/rules/html-no-empty-attributes.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// Attributes that must not have empty values
+const RESTRICTED_ATTRIBUTES = new Set([
+  'id',
+  'class',
+  'name',
+  'for',
+  'src',
+  'href',
+  'alt',
```
marcoroth: See comment below
```suggestion
```

marcoroth (commented):
Hey @auslach, thanks so much for this pull request!

I left a few comments, but this looks great! Thanks!

javascript/packages/linter/test/rules/html-no-empty-attributes.test.ts
auslach: Yep, I believe that's correct. I added these tests and they passed without any rule changes. Non output erb is effectively static and checked with `checkStaticAttributeStaticValue`

javascript/packages/linter/test/rules/html-no-empty-attributes.test.ts
auslach: Removed alt from the list of check attribute names

javascript/packages/linter/src/rules/html-no-empty-attributes.ts
auslach: removed toLowerCase()

auslach:
@marcoroth - addressed your comments! Let me know if there's anything else.

javascript/packages/linter/test/rules/html-no-empty-attributes.test.ts
marcoroth: Awesome, even better, thanks!

PR #465, opened 2025-09-01
Ruby: Re-enable Sorbet check

### LINKED ISSUE #114: Run Sorbet in CI
With https://github.com/marcoroth/herb/pull/106 we introduced RBS Inline types annotations.

It would be great if we could also use Sorbet to typecheck the codebase. Right now we are blocked because RBS aliases are not supported yet:

```
$ bundle exec srb tc

lib/herb/errors.rb:21: RBS aliases are not supported https://srb.help/3551
    21 |      #: () -> serialized_error
                       ^^^^^^^^^^^^^^^^
```

Another detail I noticed is that Sorbet doesn't support the inline type signatures for `attr_reader`, i.e this doesn't work:

```ruby
module Herb
  class Position
    attr_reader :line #: Integer
    attr_reader :column #: Integer
  end
end
```

but this does:

```ruby
module Herb
  class Position
    #: Integer
    attr_reader :line 

    #: Integer
    attr_reader :column 
  end
end
```

mickeytgl:
Now that Sorbet supports RBS aliases, this PR updates the Gemfile and modifies the files under `/lib` to fix any type errors raised by Sorbet

Resolves #114

### DIFF
--- a/templates/lib/herb/ast/nodes.rb.erb
+++ b/templates/lib/herb/ast/nodes.rb.erb
@@ -12,8 +12,7 @@ module Herb
 
       <%- node.fields.each do |field| -%>
       <%- is_nilable = !%w[Array[Herb::AST::Node] Array[Herb::AST::ERBWhenNode] Array[Herb::AST::ERBInNode] bool nil].include?(field.ruby_type) -%>
-      #: <%= field.ruby_type %><%= if is_nilable then "?" end %>
-      attr_reader :<%= field.name %>
+      attr_reader :<%= field.name %> #: <%= field.ruby_type %><%= if is_nilable then "?" end %>
       <%- end -%>
 
       #: (<%= ["String", "Location", "Array[Herb::Errors::Error]", *node.fields.map(&:ruby_type)].join(", ") %>) -> void
@@ -34,7 +33,7 @@ module Herb
           <%- node.fields.each do |field| -%>
           <%= field.name %>: <%= field.name %>,
           <%- end -%>
-        })
+        }) #: Herb::serialized_<%= node.human %>
       end
 
       #: (Visitor) -> void
@@ -48,7 +47,7 @@ module Herb
           if field.is_a?(Herb::Template::NodeField) || field.is_a?(Herb::Template::BorrowedNodeField)
             field.name
           elsif field.is_a?(Herb::Template::ArrayField) && (field.specific_kind&.end_with?("Node") || field.union_kind&.all? { |k| k.end_with?("Node") })
-            "*#{field.name}"
+            "*(#{field.name} || [])"
           else
             nil
           end
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,11 +56,11 @@ jobs:
       - name: Steep
         run: bundle exec steep check
 
-      - name: RBS Inline
-        run: bundle exec rake rbs_inline
-
       - name: Sorbet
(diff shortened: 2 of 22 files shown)


marcoroth:
Oh exciting! I missed that Sorbet added support for RBS type aliases!

templates/lib/herb/ast/nodes.rb.erb
```diff
@@ -95,7 +103,7 @@ module Herb
           <%- else -%>
           output += "│   └── "
           <%- end -%>
-          output += <%= field.name %>.tree_inspect(indent).gsub(/^/, "    " * (indent + 1)).lstrip.gsub(/^/, "<%= prefix %>").delete_prefix("<%= prefix %>")
+          output += <%= field.name %>&.tree_inspect(indent)&.gsub(/^/, "    " * (indent + 1))&.lstrip&.gsub(/^/, "<%= prefix %>")&.delete_prefix("<%= prefix %>") || ""
```
mickeytgl: This is kind of annoying, but despite checking for `field_name` above, Sorbet still considers it could be nil, so we need to add safe navigators on every method. It does seem to work on `sig` files though. It might be one of the quirks of using Sorbet with rbs

Image

templates/lib/herb/ast/nodes.rb.erb
```diff
@@ -1,9 +1,17 @@
 module Herb
   module AST
     <%- nodes.each do |node| -%>
+    #: type serialized_<%= node.human %> = {
+      <%- node.fields.each do |field| -%>
+      <%- is_nilable = !%w[Array[Herb::AST::Node] Array[Herb::AST::ERBWhenNode] Array[Herb::AST::ERBInNode] bool nil].include?(field.ruby_type) -%>
```
mickeytgl, line 6: As we talked about in Euruko, everything that isn't an array, or a boolean, or nil itself is nilable. @marcoroth

mickeytgl:
Bumping this @marcoroth ⬆️ let me know if this approach makes sense or if there's something I can change

marcoroth:
Hey @mickeytgl, is this one ready to go from your side?

mickeytgl:
Yes, all good. sorbet should now be green. I explicitely didn't add tying on some files (like `engine.rb`) but I think it can be iterated on once there's a minimum working version

lib/herb/engine/debug_visitor.rb
```diff
@@ -142,8 +144,6 @@ def replace_erb_nodes_recursive(node)
       end
 
       def find_top_level_elements(document_node)
-        @top_level_elements = [] #: Array[Herb::AST::HTMLElementNode]
-
```
marcoroth: I think we might want to keep this line here, so that it will reset the "global" instance variable when you call it again

templates/lib/herb/ast/nodes.rb.erb
marcoroth: I wonder if it might be worth to report this on Sorbet

lib/herb/engine/debug_visitor.rb
marcoroth: Actually, nvm. I think we only really call it once anyway.

marcoroth (approved):
@mickeytgl I was able to rebase it and resolve the last few issues, thank you!

mickeytgl:
Nice! Sorry I ended up forgetting about this one

PR #466, opened 2025-09-02
Formatter: Add args to the CLI

LucasKuhn:
This allows calling `herb-format` with arguments for customizing width ( `--indent-width` ) and line length ( `--max-line-length`) 

The expected behavior is to use the json config, but allowing cli args opens up to some workarounds while we don't have all integrations (eg: Zed)


marcoroth (approved):
Thank you so much @LucasKuhn! 🙏🏼

Issue #471, opened 2025-09-04
Linter: `Unexpected token` with boolean attribute + erb


darthdie:
Hi there!

I'm giving `herb-lint` a whirl, and I'm encountering the error:
```
[error] Unexpected token. Expected: TOKEN_ERB_START, TOKEN_HTML_DOCTYPE, TOKEN_HTML_COMMENT_START, TOKEN_IDENTIFIER, TOKEN_WHITESPACE, TOKEN_NBSP, TOKEN_AT, or TOKE, found: TOKEN_ERB_CONTENT. (UNEXPECTED_ERROR) (parser-no-errors)
```

When a boolean attribute is placed at the end of a tag containing ERB, e.g.
```
<div id="test" hidden>
  <%= "hi" %>
</div>
```

If I re-arrange it to the following it passes just fine:
```
<div hidden id="test">
  <%= "hi" %>
</div>
```

It seems to occur whether it's `hidden`, `disabled`, or `checked`.

marcoroth:
Hey @darthdie, thanks for reporting! This is definitely a bug!

PR #485, opened 2025-09-06
Linter: Add test case for trailing boolean attribute in HTML element

dzunk:
Sharing my investigation so far - I was able to reproduce the issue reported in #471. The core parser appears to be working correctly, but the linter errors with an unexpected token error.

Per brief convo with Marco, issue may be in the WASM build?


marcoroth:
> Per brief convo with Marco, issue may be in the WASM build?

Hey @dzunk, thanks for exploring this issue! I don't think it's in the WASM build, not sure if I said that wrong. This issue seems like it's in the parser itself in [`src/parser.c`](https://github.com/marcoroth/herb/blob/d3c284dda97a12157cc7cbefb5808d7725c9b41c/src/parser.c#L584).

dzunk:
What was throwing me off there is that the [AST snapshot](https://github.com/marcoroth/herb/pull/485/files#diff-649f8a08f197a850f66c7642a2610395da4bb8bcc973b026fc4b3cc9295e5b84) from the parser test I wrote seems correct to my eyes, with no errors. So far I've only been able to reproduce the behavior in the JS side of things. I'll retarget investigation in the C side  starting with the linked file.

marcoroth:
Oh right, I see. Now I remember. You are right, the AST snapshot for the parser test looks right. Given that it also seems fine in the `@herb-tools/browser` ([Playground link](https://herb-tools.dev/playground?tab=diagnostics#DwEwlgbgBGILwCIAuBTAzkhUAWsQoDsA+YAUjigVy1JIHpwIig)) package, means it has to be something in the `@herb-tools/node-wasm` package. This doesn't really make sense to me, very weird.

marcoroth:
@dzunk It turned out to be a bug with the `track_whitespace` parser option. I resolved this via https://github.com/marcoroth/herb/pull/560. ~~I'm going to close this one, thank you!~~

Edit: Actually, I guess we can merge the JavaScript test case you added.

dzunk:
Sounds good, thanks for the heads up (and the fix, of course)

Issue #474, opened 2025-09-05
Move Debug Exception Renderer to ReActionView (or module/class)


marcoroth:

marcoroth:
This is going away sooner rather than later, closing.

Issue #473, opened 2025-09-05
Move Debug Visitor to ReActionView


marcoroth:
The debug Visitor is somewhat Rails-specific and I think it would be better to have that in the `reactionview` gem.

marcoroth:
We'll keep them in Herb for now.

PR #486, opened 2025-09-08
Add Cursor as available editor to link source files

JuanMiguelGM:
When opening a link on ReActionView there's no possibility to open the source code with Cursor, it's opening VSCode by default. 

By adding the editor here I can open cursor now.


marcoroth:
Thanks for the pull request @JuanMiguelGM! 

I'm happy to add Cursor for now, but I guess we should figure out a better solution long-term. Something similar to what @coorasse proposed in https://github.com/marcoroth/reactionview/issues/3

JuanMiguelGM:
You’re right I was thinking about something like https://github.com/rails/rails/commit/b22fd5a519f92e5c51c62e3b74b4407e492b89e2 but my js knowledge is very limited to be able to adapt it and cursor is widely used so I was thinking that it’d be a good short term solution in the meantime I figure it out.

marcoroth:
Sounds good, let's add `cursor` for now, and we'll figure it out in https://github.com/marcoroth/reactionview/issues/3.

I also wanted to move the implementation to the ReActionView gem, since it's mostly Rails-specific and doesn't really belong here in this repo (see #473 and #474).

PR #503, opened 2025-09-17
Parser: Analyze parse errors as `partial_script`

Earlopain:
Removes the need to manually check for `invalid_yield`. Prism docs here: https://github.com/ruby/ruby/blob/4acfdb1faddf4d31d36abdfc74ea12229a6e59e3/prism/options.h#L178-L186


marcoroth (approved):
Nice, very clever! Thanks @Earlopain!


Also love how the Prism docs for the `partial_script` option exactly use this use case as an example 😅

PR #511, opened 2025-09-20
Fix test for errors with source context

timkaechele:
## Description

This PR fixes the error handling test for errors with source context.


marcoroth:
Hey @timkaechele, yeah this is a weird one. I think the problem is that this tests depends on the `herb-highlight` CLI to be available. If it's not available (since the JavaScript packages weren't not built locally) then this test will fail. 

I wonder if we should add a `which` check for this test and either skip the test or fall back to the other output when the CLI isn't available.

timkaechele:
Is there maybe a way to ensure that the functionality provided by `herb-highlight` is always available?

PR #513, opened 2025-09-21
Split init/alloc for lexer and parser

timkaechele:
This PR changes the way lexers and parsers are initialized. 

Instead of making an allocation for the lexer/parser inside the init function of the respective system, it allows the caller of the init function to decide whether the lexer/parser data is going to live on the stack or heap.

The lexer/parser lifetimes are limited to the scope of a single function enabling to use a stack variable.


timkaechele:
PR is ready for review, please take a look when you have time.

@marcoroth

Issue #527, opened 2025-09-23
Linter Rule: CommentSyntax (ERBLint rule port)


domingo2000:
**Rule:** `erb-comment-syntax`

## Description

The one line ruby comments tends to have parsing errors in ActionView parsing implementation. Because of that ERBLint hase the rule [CommentSyntax](https://github.com/Shopify/erb_lint?tab=readme-ov-file#commentsyntax).

## Rationale

Because it avoid parsing ambiguity in the ERB files.

Also the porting of ERBLint rules to Herb project will allow easier porting to Herb.

### ✅ Good

```erb
<% 
 # some good multi line ruby comment
 # some good multi line ruby comment
 # some good multi line ruby comment
 # some good multi line ruby comment
%>
<%# some good erb comment %>
<% 
 # some good ruby comment 
%>
```

### 🚫 Bad

```erb
<% # some bad ruby comment that should be changed to erb one line comment %>
```

## References

[CommentsSyntax](https://github.com/Shopify/erb_lint?tab=readme-ov-file#commentsyntax)

domingo2000:
I am willing to help implementing this rule.

marcoroth:
Hey @domingo2000, thanks for opening this issue!

I think this rule makes sense and that we should implement it either
way! 

But that said, I was literally just working on improving the way
`Herb::Engine` handles comments and I think that `Herb::Engine` could
improve/fix this. So that `<% if true # some comment %> true <% else %>
false <% end %>` would still compile and render.

domingo2000:
Hi  @marcoroth, thanks for the quick response!. I already implemented a PR porting the rule from ErbLint.

My thoughts on this rule is that it indeed a _parser error_ and all the syntax in the good/bad examples of this rules are **valid erb syntax**. But 

1. While Herb::Engine is not widely adopted, the CommentSyntax is still valuable
2. When the user adopts the new engine the rule could be disabled.
3. Current apps finds value in this rule avoiding weird rendering when using action view default implementation.

So the user could enable/disable the rule if it is using a comment resilent erb parser or not.

Edit: I think our vision is the same 👍🏻 . Happy to receive feedback on the PR.

PR #528, opened 2025-09-23
Linter: Implement `erb-comment-syntax`  linter rule

domingo2000:
Closes: #527 
Advances: https://github.com/marcoroth/herb/issues/537

This PR adds the rule `erb-comment-syntax`. It is the same rule implemented in [CommentSyntax](https://github.com/Shopify/erb_lint?tab=readme-ov-file#commentsyntax) of ERB Lint.

The rule itself avoid parsing errors in the action_view erb default parsing implementation.

Also the porting of ERB Lint rules to herb rules facilitates the adoption of Herb as a ERB Lint replacement.

### DIFF
--- a/javascript/packages/linter/docs/rules/README.md
+++ b/javascript/packages/linter/docs/rules/README.md
@@ -4,6 +4,7 @@ This page contains documentation for all Herb Linter rules.
 
 ## Available Rules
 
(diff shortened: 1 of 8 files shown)


marcoroth (approved):
Thanks for implementing this @domingo2000! 🙏🏼

PR #529, opened 2025-09-23
C: Localize token struct members

timkaechele:
## What it does

This MR removes the need for heap allocating the struct members of a token (position, range, location) 

## How it does it

- Change the `token_T` members `range` and `location` to be structs, instead of pointers to structs
- Change `location_T` members `start` and `end` to be structs, instead of pointers to structs
- Removes functions only used to access struct members, as they were not consistently used anyway
- Removes init functions that do not add anything beyond providing a 1-1 mapping of argument to struct members
- Removes copy methods as we are passing the structs by value
- Use 32bit unsigned integers for range/position/location struct members, effectively limiting the parseable filesize to 2^32-1 bytes (4gb) which for all intents and purposes (templates after all) should more than suffice. Saves a bit of memory without any real world drawbacks.

## Outcome

Early testing with real world html documents indicates:

- Lexer command: 1.15x speedup
- Parser command: 1.5x speedup
- Parser command: 25% fewer allocations measured using valgrind

## Open Todos

- [x] More comprehensive performance testing
- [x] Regression testing
- [x] Commit history cleanup (WIP commits)


timkaechele:
@marcoroth I tested the changes with all existing examples and ran the tests are there any more resources used for regression testing?

marcoroth:
Hey @timkaechele, first off, thank you so much for working on this, this looks exciting! 🙏🏼

Luckily the test suite is quite comprehensive already, at least to test parser behavior.

For "stress testing" I usally run the `bundle exec herb analyze ../rubyevents` command on the RubyEvents repo to sanity check.

timkaechele:
I think that doesn't look all that bad. 

Bildschirmfoto 2025-09-25 um 18 48 50

timkaechele:
PR is ready for review, please take a look when you have time.

@marcoroth

marcoroth:
Nice! Looks like this branch consistently lowers the runtime for parsing a decently sized template by around 1 ms in the C-CLI compared to current main (https://github.com/marcoroth/herb/commit/cce09440a1cdda85e7c3f692576761b45757e0ce).

```
herb on local-token-structs via v3.4.1 via v20.19.0
❯ hyperfine "./herb parse file.html.erb" --shell=none --warmup=100 --runs=1000
Benchmark 1: ./herb parse file.html.erb
  Time (mean ± σ):       5.5 ms ±   0.2 ms    [User: 4.6 ms, System: 0.6 ms]
  Range (min … max):     5.2 ms …   6.4 ms    1000 runs

herb on main via v3.4.1 via v20.19.0
❯ hyperfine "./herb parse file.html.erb" --shell=none --warmup=100 --runs=1000
Benchmark 1: ./herb parse file.html.erb
  Time (mean ± σ):       6.6 ms ±   0.3 ms    [User: 5.6 ms, System: 0.6 ms]
  Range (min … max):     6.2 ms …   8.5 ms    1000 runs
```

Also, looks like the physical footprint drops a bit:
```
local-token-structs:
Physical footprint:         3664K
Physical footprint (peak):  5072K

main:
Physical footprint:         3856K
Physical footprint (peak):  5648K
```

Thanks @timkaechele 🙏🏼

timkaechele:
No problem, glad we can see a ~20% performance boost. 

I think there's still a lot of performance left on the table, but that's something for another PR.

---

Sorry for the close. Wrong button :sweat_smile:

PR #531, opened 2025-09-23
Linter: Allow disabling offenses using `<%# herb:disable %>`

### LINKED ISSUE #270: Linter Feature: Disable error for line
## Description

While https://github.com/marcoroth/herb/issues/269 is working on getting resolved, it would be nice to have the ability to disable errors on lines, similar to how [standardrb allows for disabling errors](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-errors). 

## Rationale

Seems like a QOL improvement that would make any future Linting bugs easier to deal with.

```html erb
<script>
  // <div>   <-- Despite being commented out, this will raise an error.
</script>
```
Could be temporarily fixed with...
```html erb
<script>
  // <div> # herb:disable <THE_ERROR_CAUSING_THIS>
</script>
```

domingo2000:
This PR makes possible to disable a rule in a erb line. 

The implementation follows the [Standard way](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-errors) to disable rules, in the same way that [ERBLint allows to disable rules inline](https://github.com/Shopify/erb_lint?tab=readme-ov-file#disable-rule-at-offense-level).

The implementation is very similar this [pull request](https://github.com/Shopify/erb_lint/pull/249). The strategy consist of 

1. Use a regex to see which line skipped rules with the `<%# herb:disable some-rule %>` syntax
2. Partition the offenses collected by the visitor based on the rules that where disabled for each line.
3. Report in the reporters the number of offenses.

Here is a screenshot of how the implementation looks in the Linter CLI.

```erb
<div>
  <h1 class="<%= classes %>">
    <%= title %>
  </h1>

  <DIV>hello</DIV>
  <DIV>hello</DIV> <%# herb:disable html-tag-name-lowercase %>
  <% %> <%# herb:disable erb-no-empty-tags %>

</div>
```

image

Future work:
- Be able to skip the skipped lines (useful when we want to find where are the rules skipped).
- Allow the LSP to make a code action with an autocorrect that adds the line skipping.
- Allow disable multiple rules in one line (this should be pretty straightforward in a follow up PR).
- Add another linter rule to avoid putting disable comments that not disable anything (like erb_lint NoUnusedDisable).

Resolves #270

### DIFF
--- a/javascript/packages/linter/src/cli/summary-reporter.ts
+++ b/javascript/packages/linter/src/cli/summary-reporter.ts
@@ -65,7 +65,7 @@ export class SummaryReporter {
     }
 
     if (totalIgnored > 0) {
-      parts.push(colorize(colorize(`${totalIgnored} ${this.pluralize(totalIgnored, "ignored")}`, "gray"), "bold"))
+      parts.push(colorize(colorize(`${totalIgnored} ignored`, "gray"), "bold"))
     }
 
     if (parts.length === 0) {
--- a/javascript/packages/linter/test/__snapshots__/cli.test.ts.snap
+++ b/javascript/packages/linter/test/__snapshots__/cli.test.ts.snap
@@ -75,6 +75,42 @@ test/fixtures/no-trailing-newline.html.erb:1:29
   Fixable      1 offense | 1 autocorrectable using \`--fix\`"
 `;
 
+exports[`CLI Output Formatting > Ignores disabled rules 1`] = `
+"[error] Opening tag name \`<DIV>\` should be lowercase. Use \`<div>\` instead. [Correctable] (html-tag-name-lowercase)
+
+test/fixtures/ignored.html.erb:6:3
+
+      4 │   </h1>
+      5 │ 
+  →   6 │   <DIV>hello</DIV>
+        │    ~~~
+      7 │   <DIV>hello</DIV> <%# herb:disable html-tag-name-lowercase %>
+      8 │   <% %> <%# herb:disable erb-no-empty-tags %>
+
+
+⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯  [1/2] ⎯⎯⎯⎯
+
+[error] Closing tag name \`</DIV>\` should be lowercase. Use \`</div>\` instead. [Correctable] (html-tag-name-lowercase)
+
+test/fixtures/ignored.html.erb:6:14
+
+      4 │   </h1>
+      5 │ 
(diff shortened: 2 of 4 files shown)


domingo2000:
@marcoroth any feedback on this?. The TODO list is for follow up PRs, this should add value on its own. 

If you think another way to disable the rules is better, we can discuss other implementations.

marcoroth:
Hey @domingo2000, thanks for getting back to this!

I think this is a good feature to have, but I've been wondering if this is the right syntax for this, since comments based syntax for this always felt off to me (also in RuboCop and other linters).

I've been thinking about ways to handle general `herb` config options, including the linter, formatter and engine. I came across the way [svelte handles this](https://svelte.dev/docs/svelte/svelte-options) using a `<svelte:options>` tag. Since we are operating in HTML documents and can already perfectly parse this I thought it might be a really good fit to also have these options for Herb in HTML syntax.

Like 
```html
<div>
  <h1 class="<%= classes %>">
    <%= title %>
  </h1>

  <DIV>hello</DIV>
  <DIV>hello</DIV> <herb:disable html-tag-name-lowercase>
  <% %> <herb:disable erb-no-empty-tags>
</div>
```

Or maybe similarly:
```html
<herb:ignore html-tag-name-lowercase>

<herb:disable rule="html-tag-name-lowercase">

<herb:linter disable="html-tag-name-lowercase">

<herb:linter ignore="html-tag-name-lowercase">
```

or globally at the beginning of the document to also indicate other configuration options for other parts of the Herb ecosystem.
```html
<herb:formatter disabled>
<herb:linter disabled>
<herb:engine disabled>

<herb:options formatter="...">
<herb:options engine="...">
<herb:options linter="...">
```

This would allow us to use the Herb Parser itself to parse the options, and we can avoid regexes. Additionally, it allows for an easy way to ignore all `<herb:*>` tags in the `Herb::Engine` to be skipped when compiling/rendering the template. What are your thoughts on this?

domingo2000:
@marcoroth my first thought on your proposal is that having custom tags like <herb:something> will make other engines to render those tags, and that would couple the Herb::Engine with the Linter when trying to omit rules. 

For example, in a big old project, it would be nice to adopt linter + formatter as dev-tools, but changing the rendering engine could be risky. And this option does not allow to disable linting rules without using the Herb Engine. 

Also, the commenting pattern is widely adopted by all the major linters for different languages, for example ruby Rubocop, [ESLint](https://stackoverflow.com/questions/27732209/turning-off-eslint-rule-for-a-specific-line), [Golang](https://stackoverflow.com/questions/75232420/go-line-length-linter-how-to-ignore-one-line), and many more.

Even in svelte the compiling warning supression [is done via comments](https://svelte.dev/docs/svelte/compiler-warnings). 

I think that the `<herb:something>` syntax is good for engine specific configuration because the engine have the control on the rendering, but not for formatting, niether linting, when another engine could be used.

marcoroth:
@domingo2000 thanks for the feedback, that's a very fair point! I'm happy with moving forward with the comment-based syntax and getting that merged into `main` for now. 

Would you mind resolving the conflicts on this pull request? Thank you!

domingo2000:
@marcoroth done

marcoroth (approved):
Thank you @domingo2000! 🙏🏼

Issue #484, opened 2025-09-05
Segmentation fault when parsing files


luizkowalski:
hey 👋🏻 

For some time now, I'm getting segfault when analyzing views, I downgraded from 0.7 to 0.5 and still geting these erros:

```
================================================================================
/Users/luiz/Projects/switchboard/app/views/account_mailer/token_expired.html.erb
================================================================================

--- CONTENT --------------------------------------------------------------------
``erb
<p>Your Google Calendar token has expired for account <%= @account.email %>.</p>
<p>Please reconnect your Google Calendar account.</p>
``
```

error:

<details><summary>Details</summary>
<p>

--- ERROR OUTPUT ---------------------------------------------------------------
```
/Users/luiz/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/herb-0.5.0-arm64-darwin/lib/herb/project.rb:147: [BUG] Segmentation fault at 0x0000000000000097
ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [arm64-darwin24]

-- Crash Report log information --------------------------------------------
   See Crash Report log file in one of the following locations:
     * ~/Library/Logs/DiagnosticReports
     * /Library/Logs/DiagnosticReports
   for more details.
Don't forget to include the above Crash Report log file in bug reports.

-- Control frame information -----------------------------------------------
c:0035 p:---- s:0226 e:000225 CFUNC  :parse
c:0034 p:0032 s:0221 e:000220 BLOCK  /Users/luiz/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/herb-0.5.0-arm64-darwin/lib/herb/project.rb:147 [FINISH]
c:0033 p:---- s:0216 e:000215 CFUNC  :fork
c:0032 p:0006 s:0212 e:000211 BLOCK  /Users/luiz/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/herb-0.5.0-arm64-darwin/lib/herb/project.rb:142
c:0031 p:0027 s:0205 e:000204 BLOCK  /Users/luiz/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/timeout-0.4.3/lib/timeout.rb:185
[...]

PR #534, opened 2025-09-26
Parser: Namespace Parser C functions

indirect:
As painstakingly worked out with @marcoroth during Kaigi on Rails, macOS 26 Tahoe introduces a function named `parser_init` that can clobber the Herb function with the same name. Precompiled gems no longer crash on macOS 26 after this change to ensure the Herb functions will not be clobbered by the OS functions.

Resolves https://github.com/marcoroth/herb/issues/484
Resolves https://github.com/marcoroth/reactionview/issues/13

### DIFF
--- a/src/herb.c
+++ b/src/herb.c
@@ -29,9 +29,9 @@ array_T* herb_lex(const char* source) {
 
 AST_DOCUMENT_NODE_T* herb_parse(const char* source, parser_options_T* options) {
   lexer_T* lexer = lexer_init(source);
-  parser_T* parser = parser_init(lexer, options);
+  parser_T* parser = herb_parser_init(lexer, options);
 
-  AST_DOCUMENT_NODE_T* document = parser_parse(parser);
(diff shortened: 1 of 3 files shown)


marcoroth:
From what we could tell, macOS has a new(?) framework called "BiomeStreams" which both defines `parser_parse` and `parse_init` functions. 

We believe the reason why this bug only occured in the precompiled version of the gem is because at compile time it wouldn't know about both of these functions. So at runtime  it would "choose" the first matching one, which happened to be the wrong one, thus messing with the stack/structs when calling the function here:

https://github.com/marcoroth/herb/blob/b93eba9f80e7828428e084efd103acdf77bee81a/src/herb.c#L32

which then lead to this `EXC_BAD_ACCESS (code=1, address=0x29)` error later in: 

https://github.com/marcoroth/herb/blob/b93eba9f80e7828428e084efd103acdf77bee81a/src/parser.c#L1210

PR #539, opened 2025-09-28
Implement more efficient buffer resizing

timkaechele:
## Problem

If the buffer is nearly full even one extra character can trigger an expansion of the buffer. Since the buffer only grows by two characters at a time, this can lead to frequent resizing.

### Visualization

buffer_problem

## How the problem is addressed

Rather than just checking the required length, we test whether doubling the current capacity will be enough. If it is, we expand to the doubled capacity. If not, we double the required length itself and resize the buffer to that size.

## Performance impact

Lexing a  [real world html page](https://shop.herthabsc.com)

Before: ~88.255ms

After: ~79.208ms

Parsing showed no significant performance impact though

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -114,7 +114,15 @@ bool buffer_expand_capacity(buffer_T* buffer) {
 bool buffer_expand_if_needed(buffer_T* buffer, const size_t required_length) {
   if (buffer_has_capacity(buffer, required_length)) { return true; }
 
-  return buffer_resize(buffer, buffer->capacity + (required_length * 2));
+  bool fits_in_doubled_buffer = required_length < buffer->capacity;
+  size_t new_capacity = 0;
+  if (fits_in_doubled_buffer) {
+    new_capacity = buffer->capacity * 2;
+  } else {
+    new_capacity = buffer->capacity + required_length * 2;
+  }
+
+  return buffer_resize(buffer, new_capacity);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/buffer.c
```diff
@@ -114,7 +114,15 @@ bool buffer_expand_capacity(buffer_T* buffer) {
 bool buffer_expand_if_needed(buffer_T* buffer, const size_t required_length) {
   if (buffer_has_capacity(buffer, required_length)) { return true; }
 
-  return buffer_resize(buffer, buffer->capacity + (required_length * 2));
+  bool fits_in_doubled_buffer = required_length < buffer->capacity;
```
marcoroth: Thanks for working on this, @timkaechele!

Shouldn't this be `buffer->capacity * 2`? 

```suggestion
  bool fits_in_doubled_buffer = required_length < buffer->capacity * 2;
```

If not, then I guess we should rename the variable from `fits_in_doubled_buffer` to something else, or am I missing something?

timkaechele: Ah, good question, but it should be only the the single buffer capacity.

Reasoning: If you resize the buffer because it is nearly full and you plan to double the capacity you only get `buffer->capacity` more space after the resize. Hence it is `buffer->capacity`.

But I admit the name of the variable can be confusing, I couldn't come up with something significantly better though.

Candidates such as `fits_in_additional_capacity_after_doubling` felt a bit clumsy.

marcoroth: Ah right, I see the trouble, again.

I just renamed this this `should_double_capacity`, thanks!

Issue #551, opened 2025-09-30
erb-lint: RightTrim


marcoroth:
https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#righttrim

marcoroth:
Resolved via #556

PR #556, opened 2025-09-30
Linter: Implement `erb-right-trim`  linter rule

domingo2000:
Solves #551 

Implements the [RightTrim rule from erb-lint](https://github.com/Shopify/erb_lint?tab=readme-ov-file#righttrim). 

This implementation defaults to the "-" character.

If there is a mechanism to configure the linter rules i can change the implementation to match the ErbLinter rule that can configure the "-" or "=" character. 

But i prefer to enforce the "-" character, because the [ERB implementation of the language only knows about "-"](https://docs.ruby-lang.org/en/3.4/ERB.html#method-c-new) for right triming. I don't know in which context the caracter "=" is valid for right trimming as the ErbLint rules says. But at least not in pure erb templates of ruby.


domingo2000:
On the other hand, the syntax ending with `=%>` that the rules enforce, result in a parsing error in Herb. So maybe this rule could also be ommited, because the ParsingErrors shown by the linter will catch this.

image

marcoroth:
Hey @domingo2000, thanks for working on this!

I'm actually confused about `=%>`, as I don't think I ever saw this anywhere and I also don't see which part of Erubi or ERB would implement such a functionality. Unless I'm missing something.

My guess is that this rule mostly serves to guide users not too familiar with ERB, since it seems logical on first sight that a tag opened with `<%=` should be symmetrically closed with `=%>`. Or that you can catch a "typo" when you actually meant `-%>`.

domingo2000:
> Hey @domingo2000, thanks for working on this!
> 
> I'm actually confused about `=%>`, as I don't think I ever saw this anywhere and I also don't see which part of Erubi or ERB would implement such a functionality. Unless I'm missing something.
> 
> My guess is that this rule mostly serves to guide users not too familiar with ERB, since it seems logical on first sight that a tag opened with `<%=` should be symmetrically closed with `=%>`. Or that you can catch a "typo" when you actually meant `-%>`.

In that case, I think the parsing errors should already cover this. If you agree, we could mark the compatibility of this rule as resolved.

marcoroth:
We could add `=%>` to the list in `lexer_parse_erb_close` here:

https://github.com/marcoroth/herb/blob/2f0bb43f69087bd969835e51b29213dc3e428efc/src/lexer.c#L255-L262

So that the parser doesn't stumble and then we could add a more specific and actionable lint/error message in the linter.

domingo2000:
I don’t really like that approach, because as you mentioned, the `=%>` ending isn’t valid syntax, so the parser error is more meaningful than the `RightTrim` rule.

marcoroth:
I'm just confused because [`erb_lint` explicitly talks about it being valid syntax](https://github.com/shopify/erb_lint?tab=readme-ov-file#righttrim). It's definitely also being used in [some places](https://grep.app/search?f.lang=HTML%2BERB&q=%3D%25%3E).

Also, seems like Erubi happliy compiles it:

```shell
❯ cat test_right_trim.html.erb
<%= RUBY_VERSION =%>

❯ bin/erubi-compile test_right_trim.html.erb
__erubi = ::Erubi; _buf = ::String.new; _buf << __erubi.h(( RUBY_VERSION ));
_buf.to_s

domingo2000:
> I'm just confused because [`erb_lint` explicitly talks about it being valid syntax](https://github.com/shopify/erb_lint?tab=readme-ov-file#righttrim). It's definitely also being used in [some places](https://grep.app/search?f.lang=HTML%2BERB&q=%3D%25%3E).

[This issue](https://github.com/jeremyevans/erubi/issues/38) says is indeed a valid syntax for erubi compatibility, but not for ERB compatibility.

Never have seed that syntax before 😶‍🌫️.  But it is valid for erubi

I tried using the `=%>` in different contexts:

##  ActionView: Works because of erubi

image


Rendered HTML in http response
image


Action View uses Erubi by default


## Rails Generatiors: Throw error
image

## ruby ERB bundled gem

Throws error too.

image

image

So given this new information we could indeed add that syntax into the lexer as a valid closing erubi tag. And make the linter default to `-` for trimming because is accepted by all implementations. 

With that, the syntax error on `ERB` would be warned, but the usages in ActionView Erubi could also be supported because it will throw no parsing error.

domingo2000:
I took the aproach to provide compatibility with erubi for the "=%>" closing tag.

1. Added support in the parser vía lexing
2. Added support for the whitespace trimming as another case in Herb::Engine, providing better erubi compatibility
3. Updated the herb:lint erb-right-trim rule with the new ErbClosingTag value read from the parser.

domingo2000:
The only thing left should be to make it configurable vía the future linter config. But for now, having "-" as the default trimming for this rule should be ok. Given that the more supported implementation of trimming is with the "-" character.

marcoroth:
> The only thing left should be to make it configurable vía the future linter config. But for now, having "-" as the default trimming for this rule should be ok. Given that the more supported implementation of trimming is with the "-" character.

My take is that we shouldn't make this linter rule configurable and "force" people to use the `-%>` syntax.

domingo2000:
@marcoroth just rebased and used the updated lexer. 

I agree with enforcing "-" instead of "=" without config 👍🏻

marcoroth (approved):
Sweet, thanks @domingo2000! 🙏🏼

PR #557, opened 2025-10-01
Linter: Implement `erb-no-extra-newline` linter rule

### LINKED ISSUE #541: ExtraNewline

domingo2000:
Implements the `ExtraNewLine` rule from [ERBLint](https://github.com/Shopify/erb_lint).

This rules avoid having multiple extra consecutives new lines in one file.

Resolves #541


domingo2000:
I have the same thoughts of this rule as with SpaceInHtmlTag. This rules are "formatting rules", if the Herb formatter could (or can i don't know) verify that a file is correctly formatter, this rule should not be necessary and can be removed from the roadmap for ErbLint compatibility.

I left an implementation anyways if you belive it provide some value. Feel free to change the severity from error to warning, or to make this rule an optional rule instead of default.

PR #559, opened 2025-10-01
Linter: Implement `html-no-space-in-html-tag`  linter rule

### LINKED ISSUE #549: SpaceInHtmlTag

domingo2000:
Implements the `SpaceInHtmlTag` rule from [ERBLint](https://github.com/Shopify/erb_lint).

This rules avoid having whitespace in the html.

This is a Proof of concept of the implementation. I considered

1. If the parser already throws a parser error, the linter from erb-lint makes no sense (ej: `</   div>`)
2. It follows the result that the formatter would give. For example, for multiline, the whitespace must have 2 spaces.

Resolves #549

### DIFF
--- a/javascript/packages/linter/docs/rules/README.md
+++ b/javascript/packages/linter/docs/rules/README.md
@@ -35,12 +35,12 @@ This page contains documentation for all Herb Linter rules.
 - [`html-no-nested-links`](./html-no-nested-links.md) - Prevents nested anchor tags
 - [`html-no-positive-tab-index`](./html-no-positive-tab-index.md) - Avoid positive `tabindex` values
 - [`html-no-self-closing`](./html-no-self-closing.md) - Disallow self closing tags
+- [`html-no-space-in-tag`](./html-no-space-in-tag.md) - Disallow spaces in HTML tags
 - [`html-no-title-attribute`](./html-no-title-attribute.md) - Avoid using the `title` attribute
-- [`html-tag-name-lowercase`](./html-tag-name-lowercase.md) - Enforces lowercase tag names in HTML
 - [`html-no-underscores-in-attribute-names`](./html-no-underscores-in-attribute-names.md) - Disallow underscores in HTML attribute names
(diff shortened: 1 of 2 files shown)


javascript/packages/linter/src/rules/html-no-space-in-tag.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        if(whitespace.value.value === "\n") {
+          if (previousWhitespace && previousWhitespace.value && previousWhitespace.value.value === "\n") {
+            this.addOffense("Extra space detected where there should be no space.", whitespace.location)
+          }
+        } else {
+          if (index  === lastIndex) {
+            this.addOffense("Extra space detected where there should be no space.", whitespace.location)
+          } else {
+            if (whitespace.value.value.length > 2) {
```
domingo2000: This is not going to work for nested tags, still in development. But waiting for feedback if this rule really adds value if we have the formatter.

marcoroth: Ideally the linter can be used independently of the formatter. At least it's been designed like that, so that if you are auto-fixing linter offenses that it only fixes that, and that it doesn't re-format just to be able to auto-fix.

domingo2000: Ok, I tried to implement the rule based on the tests cases of ErbLint rule. But changed the implementation and some cases to be consistent with the herb formatter.

This way formatting with Herb formatter gives a valid html for this rule. But also the linter throws incorrect formatting in the incorrect whitespaces.

Happy if you tests this rule to find more edge cases that I could have missed.

marcoroth: Just added some more tests and refactored it a bit, but looks good to me! 🙏🏼

domingo2000: Thanks!, i knew the implementation was not perfect, and didn't have time to rebase the changes on the test helpers. thanks for the refactoring! 💯 .

Issue #356, opened 2025-08-08
Rubocop integration


jbigler:
Thanks for the great tool!

Though it is quite slow, one of the reasons I am using erb_lint in my editor and CI is because of its ability to run Rubocop (with the ability to exclude rules that don't apply in ERB files) on the embedded Ruby code.

For example, we have a rule on how to format arrays or hashes in our Ruby code, we want that to be applied to the Ruby in our ERB files as well.

Is there a way to add this functionality to Herb?

marcoroth:
Hey @jbigler, thanks for the suggestion! 

Yes, that's the plan to get to. I'm not 100% how we could achieve that yet architecture-wise (since we don't have any Ruby dependency for the linter or formatter now). But I'm sure we can find a way to make it work. In the worst-case it's a setting to enable and it has to shell-out.

maxjacobson:
It may be useful to reference how ERB Lint implements [their RuboCop integration](https://github.com/Shopify/erb_lint?tab=readme-ov-file#rubocop), including the challenges with providing feedback that makes sense to users:

> **Note**: Each ruby statement (between ERB tags `<% ... %>`) is parsed and analyzed independently of each other. Any rule that requires a broader context can trigger false positives (e.g. `Lint/UselessAssignment` will complaint for an assignment even if used in a subsequent ERB tag).

drjole:
We (that is @LaurinKerkloh and me) have started working on integrating RuboCop into the Herb ecosystem: https://github.com/einfach-nur/herb/tree/rubocop

We added a new package (`@herb-tools/rubocop`) that handles the communication with RuboCop via the RuboCop CLI:
- It extracts the Ruby code from a provided ERB node and invokes the RuboCop CLI to get offenses and their corrections
- While parsing the Ruby code from the ERB node, we keep track of the locations of the Ruby code inside the ERB document
- Finally we translate the offense locations back to ERB locations

With this approach, both the linter and the formatter (and therefore the LSP) can show/correct RuboCop offenses, respectively.

We disabled the following cops for now:

- Style/FrozenStringLiteralComment
- Lint/Void
- Layout/IndentationWidth
- Layout/ArgumentAlignment

There are more cops that require additional logic to work correctly in ERB.

The formatter can already format simple offenses but fails in more complex situations.
The linter works very well already.

We will continue working on this feature and only just started so please don't expect too much for now.

[...]

PR #565, opened 2025-10-01
Linter: Implement `html-input-require-autocomplete` linter rule

### LINKED ISSUE #552: RequireInputAutocomplete

domingo2000:
Implements the [RequireInputAutocomplete ERBLint rule](https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#requireinputautocomplete)

Resolves #552


domingo2000:
Currently I only implemented the rule for the HTML nodes. 

For the ruby CallNodes that are inside of the content of ERBNodes i don't know how to proceed, I checked the other rules, but there is no Ruby Prism AST inside the ERB nodes to analyze with visitors.

I think this could be done in 2 ways.

1. Make the mixes AST nodes with the prsim nodes inside the ERB tags available to the linter: This would be nice and very flexible for other rules, making very similar to lint ruby nodes inside ERB nodes easily.
2. With the [rubocop integration](https://github.com/marcoroth/herb/issues/356), we could write a rubocop rule that lint those helpers in rubocop.

@marcoroth if you have any more ideas in how to cover the remaining tests for the form helpers, I would be happy to help.

PR #568, opened 2025-10-02
Lexer: Support lexing and parsing `=%>` ERB closing tag

marcoroth:
This pull request adds support to be able to lex and parse the `=%>` ERB closing tag. Since it's use is quite unknown and not well-defined we should be able to parse is, so we can guide and advice people in the linter to now use it, i.e using the Right Trim rule introduced in #556.

/cc @domingo2000


marcoroth:
@domingo2000 Oh whoops, just realized you already partly implemented it in #556. My bad. Let me merge this one, so we can keep the linter rule as a separate PR in #556. I'm going to add you as a co-author for this one 🙌🏼

domingo2000:
Nice, i didn't know how to test all the things I implemented 👌🏻. I will rebase this week.

PR #579, opened 2025-10-05
C: Bump default buffer size to `4096`

marcoroth:
This pull request bumps the default buffer size in C from `1024` bytes to  `4096` bytes, since these buffers usually hold file contents.

Bumping the default buffer size requires fewer reallocations/resizings of the buffer, which is especially useful when working with bigger files. This should make it more reliable, since calling `realloc` in quick succession might fail in some scenarios.

Additionally, this pull request introduces two new methods `buffer_init_with_capacity` and `buffer_new_with_capacity` which can be used to create smaller/bigger buffers when needed.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -8,7 +8,11 @@
 #include "include/util.h"
 
 bool buffer_init(buffer_T* buffer) {
-  buffer->capacity = 1024;
+  return buffer_init_with_capacity(buffer, 4096);
+}
+
+bool buffer_init_with_capacity(buffer_T* buffer, const size_t initial_capacity) {
+  buffer->capacity = initial_capacity;
   buffer->length = 0;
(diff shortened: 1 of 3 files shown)


timkaechele:
Is there a realiable way to replicate the realloc failures? 

Would be interesting to figure out the underlying cause.

marcoroth:
Yeah, sadly not easily reproducible. It seems to happen randomly and it might not even be related to the buffer itself/directly.

I also saw it segfault a few times in CI: https://github.com/marcoroth/herb/actions/runs/18236718031/job/51931774274

timkaechele:
Do you know whether there is a way to download the core dump from the build somewhere?

If so, we could take a look what causes the segfault.

marcoroth:
Not for the previous ones, but I guess if we would know the file location we could upload them using the `upload-artifact` GitHub Action.

PR #610, opened 2025-10-10
Ruby: NTFS compliant snapshot filenames

timkaechele:
This PR changes the filenames of the snapshots used for testing to allow checking out this project on NTFS filesystems.


timkaechele:
Done. I just saw that it moved a bunch of files and thought that was it.

Ran the command as instructed and pushed the new state.

PR #613, opened 2025-10-10
C: Remove unused array functions

timkaechele:
This PR removes unused array functions that are not used anymore. 

## Removed function

- array_first
- array_set
- array_remove
- array_remove_item
- array_index_of
- array_push

## Replacement of `array_push`

`array_push` was replaced by defining a marcro calling `array_append`

### DIFF
--- a/src/include/array.h
+++ b/src/include/array.h
@@ -15,9 +15,9 @@ void* array_get(const array_T* array, size_t index);
 void* array_last(array_T* array);
 
 void array_append(array_T* array, void* item);
+#define array_push(array, item) (array_append(array, item))
 void array_free(array_T** array);
 
-void array_push(array_T* array, void* item);
 void* array_pop(array_T* array);
 
 size_t array_capacity(const array_T* array);
--- a/src/array.c
+++ b/src/array.c
@@ -67,11 +67,6 @@ void* array_last(array_T* array) {
   return array->items[array->size - 1];
 }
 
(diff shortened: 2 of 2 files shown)


marcoroth:
I'm not 100% sold on removing these functions (which are technically unused in the codebase). Personally, I see these files more as a "stdlib" of sorts. 

Which is why I wanted to have symmetric methods for all the common operations. Having an `array_pop` without an `array_push`, or an `array_first` without an `array_last` feels weird. 

But I guess these functions are also easily added back when you'd actually need them.

src/include/array.h
```diff
@@ -12,18 +12,12 @@ typedef struct ARRAY_STRUCT {
 array_T* array_init(size_t capacity);
 
 void* array_get(const array_T* array, size_t index);
-void* array_first(array_T* array);
 void* array_last(array_T* array);
 
 void array_append(array_T* array, void* item);
-void array_set(const array_T* array, size_t index, void* item);
+#define array_push(array, item) (array_append(array, item))
```
timkaechele, line 18: Note: array push still exists, it's just changed to a preprocessor macro

timkaechele:
I see. 

My argument for the removal of the `array_first` would have been that, `array_last` can be involved, you have to get the length, remember that you have to subtract 1 etc. On the other hand the `array_first` method is just a shorthand for `->items[0]`.

Not married to it.

PR #618, opened 2025-10-11
C: Implement `hb_string` struct

timkaechele:
Implement a string datastructure representing a string with a known limit, eliminating the need to always use null terminated strings.

This lays the groundwork for reducing the number of required allocations during parsing.

See #580

### DIFF
--- a/src/include/hb_string.h
+++ b/src/include/hb_string.h
@@ -1,0 +1,35 @@
+#ifndef HERB_STRING_H
+#define HERB_STRING_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+typedef struct STRING_STRUCT {
+  char* data;
+  size_t length;
+} hb_string_T;
+
+hb_string_T hb_string_from_c_string(const char* null_terminated_c_string);
+bool hb_string_equals(hb_string_T a, hb_string_T b);
+bool hb_string_equals_case_insensitive(hb_string_T a, hb_string_T b);
+bool hb_string_starts_with(hb_string_T string, hb_string_T expected_prefix);
+bool hb_string_is_empty(hb_string_T string);
+
+/**
+ * @brief Creates a null terminated c string version of the passed string
+ *
+ * @param string The string for which a c string version is to be created
+ * @return A heap allocated null terminated c string
+ *
+ * Example:
+ * @code
+ *
+ * hb_string_T string = hb_string_from_c_string("Hello, world");
+ * char* cstring = hb_string_to_c_string(string);
+ *
+ * free(cstring);
+ * @endcode
+ */
+char* hb_string_to_c_string(hb_string_T string);
+
+#endif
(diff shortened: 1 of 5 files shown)


src/include/str.h
```diff
@@ -0,0 +1,22 @@
+#ifndef HERB_STR_H
+#define HERB_STR_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+typedef struct STR_STRUCT {
+  char* data;
+  size_t length;
+} str_T;
```
marcoroth: Given that we probably need to namespace these utils at some point (see #534) we could start with this one, given we are introducing a new struct here.

Prism uses the `pm_*` prefix, so I guess we could do something like `hb_*` for Herb. I'm personally also not the biggest fan of abbreviations (like `str`) so I'd propose to name this struct `string`. 

With that, I'd propose `hb_string.c` and `hb_string_T`. Maybe we can also move these "stdlib-like" files also into `src/utils/*` like [Prism does too](https://github.com/ruby/prism/tree/main/src/util).

src/include/str.h
```diff
@@ -0,0 +1,22 @@
...
+
+str_T str_from_c_string(const char* null_terminated_c_string);
```
marcoroth: Which means we should probably also update all function names to match the convention, like:
```suggestion
hb_string_T hb_string_from_c_string(const char* null_terminated_c_string);
```

src/include/str.h
timkaechele: No objections. Let's do it.

src/include/hb_string.h
```diff
@@ -0,0 +1,22 @@
...
+} hb_string_T;
+
+hb_string_T hb_string_from_c_string(const char* null_terminated_c_string);
+bool hb_string_equals(hb_string_T a, hb_string_T b);
+bool hb_string_equals_case_insensitive(hb_string_T a, hb_string_T b);
+bool hb_string_starts_with(hb_string_T string, hb_string_T expected_prefix);
+bool hb_string_is_empty(hb_string_T string);
+
+// Allocates a null terminated c string with the contents of the passed string
+// it's your responsibility to free it.
```
marcoroth: We might be able to move this to the `*.c` file and add a proper docstring, like we do in `html_util.c`:

https://github.com/marcoroth/herb/blob/d157d8b8843b475a4cb295576e81a70753b2f026/src/html_util.c#L27-L42

timkaechele: I am happy to add the doc comment, but the placement in the c file seems a bit odd to me. 

The header files represent the public api and for example if you would distribute a closed source binary you would only have access to the headers, not the actual C implementation, hence I am more familiar adding doc comments in the header file.

marcoroth: Good point! I guess working on the library itself it felt more natural in the C file, but I can see why putting it in the header makes more sense!

PR #619, opened 2025-10-11
C: Add null checks back in C-Structs converting helpers

marcoroth:
Follow up on #529 

/cc @timkaechele


marcoroth:
Not sure if we can even properly do this now, now that we don't have pointers anymore.

timkaechele:
I think there's no point in guarding against nil anymore. 

We only have to ensure that all instances of position/location/range are initialized to sensible values.

PR #623, opened 2025-10-11
C: Implement rudimentary arena allocator `hb_arena.c`

timkaechele:
This PR implements an arena allocator.

It provides an initial interface for performing allocations using the function `hb_arena_alloc(...)`.

## Implementation Details

All returned pointers **have an 8 byte alignment.** `malloc` on x64 machines returns 16 byte aligned memory [ref](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf), but from everything I could find only 128bit integers and simd types really require 16 bytes. Since we don't use SIMD right now, I think we can go with 8 bytes.

We use [`mmap`](https://man7.org/linux/man-pages/man2/mmap.2.html) for allocating the memory which only allocates the memory when the page is actually written to on modern UNIX operating systems. This has the advantage that we can reservce e.g. 512mb of memory and the OS only really allocates the memory if we make our first write. 

## `mmap` example

```c
#include <string.h>
#include <stdlib.h>
#include <stddef.h>
#include <unistd.h>
#include <sys/mman.h>

#define KB(kb) (1024 * kb)
#define MB(mb) (1024 * KB(mb))

int main() {
  size_t memory_size = MB(512);
  char *memory = mmap(NULL, memory_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);

  memset(memory, 0, MB(400));

  sleep(100);
}
```

Compile and start program

```bash
$ ./a.out &
$ vmmap $PID_SHOWN_IN_TERMINAL
```

```
...
                                 VIRTUAL   RESIDENT      DIRTY    SWAPPED ALLOCATION      BYTES DIRTY+SWAP          REGION
MALLOC ZONE                         SIZE       SIZE       SIZE       SIZE      COUNT  ALLOCATED  FRAG SIZE  % FRAG   COUNT
===========                      =======  =========  =========  =========  =========  =========  =========  ======  ======
DefaultMallocZone_0x104a5c000     524.5M     400.1M     400.1M         0K        185     512.0M         0K      0%      10
```

### Activity Manager View

memory_usage

### DIFF
--- a/src/hb_arena.c
+++ b/src/hb_arena.c
@@ -1,0 +1,57 @@
+#ifdef __linux__
+#define _GNU_SOURCE
+#endif
+#include "include/hb_arena.h"
+#include <stdbool.h>
+#include <sys/mman.h>
+
+#define KB(kb) (1024 * kb)
+#define MB(mb) (1024 * KB(mb))
+
+bool hb_arena_init(hb_arena_T* allocator, size_t size) {
+  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+  if (allocator->memory == MAP_FAILED) {
+    allocator->memory = NULL;
+    allocator->position = 0;
+    allocator->capacity = 0;
+    return false;
+  }
+
+  allocator->position = 0;
+  allocator->capacity = size;
+  return true;
+}
+
+void* hb_arena_alloc(hb_arena_T* allocator, size_t size) {
+  size_t required_size = ((size + 7) / 8) * 8; // rounds up to the next 8 bytes
+
+  if (allocator->position + required_size > allocator->capacity) { return NULL; }
+
+  size_t offset = allocator->position;
+  void* ptr = &allocator->memory[offset];
+  allocator->position += required_size;
+
+  return ptr;
+}
+
+size_t hb_arena_position(hb_arena_T* allocator) {
+  return allocator->position;
+}
+
+void hb_arena_reset(hb_arena_T* allocator) {
+  allocator->position = 0;
+}
+
+void hb_arena_reset_to(hb_arena_T* allocator, size_t new_position) {
+  if (new_position <= allocator->capacity) { allocator->position = new_position; }
+}
+
+void hb_arena_free(hb_arena_T* allocator) {
+  if (allocator->memory != NULL) {
+    munmap(allocator->memory, allocator->capacity);
+    allocator->memory = NULL;
+    allocator->position = 0;
+    allocator->capacity = 0;
+  }
+}
--- a/src/include/hb_arena.h
+++ b/src/include/hb_arena.h
@@ -1,0 +1,20 @@
+#ifndef HERB_ARENA_H
+#define HERB_ARENA_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+typedef struct HB_ARENA_ALLOCATOR_STRUCT {
+  char* memory;
+  size_t capacity;
+  size_t position;
+} hb_arena_T;
+
+bool hb_arena_init(hb_arena_T* allocator, size_t size);
+void* hb_arena_alloc(hb_arena_T* allocator, size_t size);
+size_t hb_arena_position(hb_arena_T* allocator);
+void hb_arena_reset(hb_arena_T* allocator);
+void hb_arena_reset_to(hb_arena_T* allocator, size_t new_position);
+void hb_arena_free(hb_arena_T* allocator);
+
+#endif
--- a/javascript/packages/node/binding.gyp
+++ b/javascript/packages/node/binding.gyp
@@ -27,6 +27,7 @@
         "./extension/libherb/lexer_peek_helpers.c",
         "./extension/libherb/lexer.c",
         "./extension/libherb/location.c",
+        "./extension/libherb/hb_arena.c",
         "./extension/libherb/parser_helpers.c",
         "./extension/libherb/parser.c",
         "./extension/libherb/pretty_print.c",
--- a/test/c/main.c
+++ b/test/c/main.c
@@ -10,6 +10,7 @@ TCase *lex_tests(void);
 TCase *token_tests(void);
 TCase *util_tests(void);
 TCase *hb_string_tests(void);
+TCase *hb_arena_tests(void);
 
 Suite *herb_suite(void) {
   Suite *suite = suite_create("Herb Suite");
@@ -23,6 +24,7 @@ Suite *herb_suite(void) {
   suite_add_tcase(suite, token_tests());
   suite_add_tcase(suite, util_tests());
   suite_add_tcase(suite, hb_string_tests());
+  suite_add_tcase(suite, hb_arena_tests());
 
   return suite;
 }
--- a/test/c/test_hb_arena.c
+++ b/test/c/test_hb_arena.c
@@ -1,0 +1,57 @@
+#include "include/test.h"
+#include "../../src/include/hb_arena.h"
+
+// Test appending text to buffer
+TEST(test_arena_alloc)
+  hb_arena_T allocator;
+
+  hb_arena_init(&allocator, 1024);
+  ck_assert_int_eq(allocator.position, 0);
+  ck_assert_int_eq(allocator.capacity, 1024);
+
+  {
+    // Ensure memory is writable
+    char *memory = hb_arena_alloc(&allocator, 1);
+    *memory = 'a';
+  }
+
+  ck_assert(allocator.position % 8 == 0);
+  ck_assert_int_eq(allocator.position, 8);
+
+  char *memory = hb_arena_alloc(&allocator, allocator.capacity - allocator.position);
+  ck_assert_ptr_nonnull(memory);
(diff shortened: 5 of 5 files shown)


marcoroth:
Sweet, thanks for exploring this! 

I'm wondering how `mmap` works in WASM and browser environments, is that just supported/handled by emscripten? Judging from the CI build it seems to just work!

timkaechele:
> I'm wondering how `mmap` works in WASM and browser environments, is that just supported/handled by emscripten? Judging from the CI build it seems to just work!

I am not entirely sure about WASM, there were some open issues relating to mmap around 2020, it could very well be that this is implemented by now. 

If not we can think of a platform specific implementation.

E.g. if we would port this to windows, we can't use mmap either, we would have to use windows' [`VirtualAlloc`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc) function.

marcoroth:
I guess the question is if `mmap` buys us that many more benefits over regular `malloc`. But also, we are probably not going to support Windows anyway, and if we do, it'll be through the `@herb-tools/node` and/or `@herb-tools/node-wasm` packages.

timkaechele:
Granted it's a bit more low level than malloc, in exchange you get more control over the memory commit behavior and can control the individual memory pages, as you can for example set the memory page flags (https://linux.die.net/man/2/mprotect). 

Fun fact: malloc itself uses mmap in its implementation [ref](https://github.com/lattera/glibc/blob/master/malloc/malloc.c#L1030), you just cut out the middleman. 

Why would you want to control the memory page flags? You can e.g. set the page before and after an allocation as `PROT_NONE` so that the memory access fails. You then can have different builds, one where all allocations are right at the beginning of the page, and one where the last bit of the allocation aligns with the end of the allocated page, this way every memory over-/underflow leads to a segfault that can easily be debugged.

src/hb_arena.c
```diff
@@ -0,0 +1,47 @@
...
+  if (allocator->position + required_size > allocator->capacity) { return NULL; }
+
+  size_t offset = allocator->position;
+  void* ptr = &allocator->memory[offset];
+  allocator->position += required_size;
+
+  return ptr;
+}
+
+size_t hb_arena_pos(hb_arena_allocator_T* allocator) {
```
marcoroth:
```suggestion
size_t hb_arena_position(hb_arena_allocator_T* allocator) {
```

src/hb_arena.c
```diff
@@ -0,0 +1,47 @@
...
+  return allocator->position;
+}
+
+void hb_arena_reset(hb_arena_allocator_T* allocator) {
+  allocator->position = 0;
+}
+
+void hb_arena_reset_to(hb_arena_allocator_T* allocator, size_t new_position) {
+  allocator->position = new_position;
+}
```
marcoroth: Should we add a guard here to not overflow/underflow? 
```suggestion
void hb_arena_reset_to(hb_arena_allocator_T* allocator, size_t new_position) {
  if (new_position >= 0 && new_position <= allocator->capacity) {
    allocator->position = new_position;
  }
}
```

src/include/hb_arena.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <stddef.h>
+
+typedef struct HB_ARENA_ALLOCATOR_STRUCT hb_arena_allocator_T;
+
+struct HB_ARENA_ALLOCATOR_STRUCT {
+  char* memory;
+  size_t capacity;
+  size_t position;
+};
```
marcoroth: Is there a difference between the explicit `typedef struct` and just naming the `struct`?
```suggestion
struct HB_ARENA_ALLOCATOR_STRUCT {
  char* memory;
  size_t capacity;
  size_t position;
} hb_arena_allocator_T;
```

src/hb_arena.c
```diff
@@ -0,0 +1,47 @@
...
+void hb_arena_init(hb_arena_allocator_T* allocator, size_t size) {
+  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  allocator->position = 0;
+  allocator->capacity = size;
+}
+
+void* hb_arena_alloc(hb_arena_allocator_T* allocator, size_t size) {
+  size_t required_size = ((size + 7) / 8) * 8; // rounds up to the next 8 bytes
+
+  if (allocator->position + required_size > allocator->capacity) { return NULL; }
```
marcoroth, line 29: Does it make sense to `realloc` (or the `mmap` equivalent) in case there is no capacity left? Or should we handle this somewhere else?

src/include/hb_arena.h
```diff
@@ -0,0 +1,21 @@
...
+
+void hb_arena_init(hb_arena_allocator_T* allocator, size_t size);
+void* hb_arena_alloc(hb_arena_allocator_T* allocator, size_t size);
+size_t hb_arena_pos(hb_arena_allocator_T* allocator);
+void hb_arena_clear(hb_arena_allocator_T* allocator);
```
marcoroth: I guess we should either remove `hb_arena_clear` here in the header or implement the function
```suggestion
```

src/hb_arena.c
```diff
@@ -0,0 +1,47 @@
...
+#define _GNU_SOURCE
+#endif
+#include "include/hb_arena.h"
+#include <sys/mman.h>
+
+#define KB(kb) (1024 * kb)
+#define MB(mb) (1024 * KB(mb))
+
+void hb_arena_init(hb_arena_allocator_T* allocator, size_t size) {
+  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```
marcoroth, line 12: Is there a way for us to check if `mmap` was successful? I feel like we'd want to `exit(1)` or similar in the case of unsuccessful allocation.

src/include/hb_arena.h
```diff
@@ -0,0 +1,21 @@
+#ifndef HERB_ARENA_H
+#define HERB_ARENA_H
+
+#include <stddef.h>
+
+typedef struct HB_ARENA_ALLOCATOR_STRUCT hb_arena_allocator_T;
```
marcoroth: While `hb_arena_allocator_T` is the most explicit name we could give it, it's quite long. Given that we are going to have to pass this into a lot of functions I wonder if we should shorten this to `hb_arena_T`:
```suggestion
typedef struct HB_ARENA_ALLOCATOR_STRUCT hb_arena_T;
```

So function declarations could refer to it as:
```diff
-void some_function(hb_arena_allocator_T* allocator, ...);
+void some_function(hb_arena_T* arena, ...);
```

or:
```diff
-void some_function(hb_arena_allocator_T* allocator, ...);
+void some_function(hb_arena_T* allocator, ...);
```

src/include/hb_arena.h
timkaechele: Ah, this is a left over. 

I initially had the idea, that an arena can have a child arena for temporary allocations, but this would have been a recursive definition (arena struct has a member that is also an arena 🤯), hence I declared the struct and gave it a typedef just so that I can define the struct.

src/hb_arena.c
timkaechele: So, we have a couple of ways to do this. 

We can't realloc, because there's no guarantee that we get the same chunk of memory, all pointers would be invalid, instead we can only make new allocations.

For now, I thought we can just chose a large capacity, that is unrealistic to ever be reached in a normal workflow, e.g. make it 256mb. 

An alternative is that you structure the arenas as link list nodes, and every time you run out of memory, in one node, you allocate a new one with double the capacity or if that is too small, the requested allocation size. 

I kept it simple for now, using mmap, we can reserve a large amount of memory and we only get it once we actually write to it.

src/hb_arena.c
timkaechele: We can, there is a constant to check. 

The question is though: what to do if it fails? I know we can't be silent about it, but we can't just exit either, if libherb is used in a server, you can't just exit the server process.

src/include/hb_arena.h
timkaechele: Sure, that works as well.

src/hb_arena.c
timkaechele: The capacity check makes sense, the check for zero or greater is redundant though. `size_t` is an unsigned integer, it can only represent not negative natural numbers.

src/include/hb_arena.h
timkaechele: Oh, this was from the naming phase, I settled on `reset` and forgot to change it in the header.

src/hb_arena.c
timkaechele: I added a check for the `MAP_FAILED` flag and made `arena_init` return a boolean, if it fails it just returns false and an initialized arena that has no capacity.

src/hb_arena.c
marcoroth: Good point 👍🏼 

Thanks!

src/hb_arena.c
marcoroth: Sweet, thanks!

PR #625, opened 2025-10-12
C: Make parser options a normal struct member of the parser

timkaechele:
This PR changes the way we store the parser options in the parser struct. 

Instead of a heap allocation, we just use a local member.

In addition to less pointer chasing when checking the options, we also get an explicit definition of the default parser options. 

Public interfaces are unchanged by this refactoring.

### DIFF
--- a/src/herb.c
+++ b/src/herb.c
@@ -32,7 +32,12 @@ AST_DOCUMENT_NODE_T* herb_parse(const char* source, parser_options_T* options) {
   lexer_T lexer = { 0 };
   lexer_init(&lexer, source);
   parser_T parser = { 0 };
-  herb_parser_init(&parser, &lexer, options);
+
+  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
+
+  if (options != NULL) { parser_options = *options; }
+
+  herb_parser_init(&parser, &lexer, parser_options);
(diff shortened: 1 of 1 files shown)


src/herb.c
```diff
@@ -32,7 +32,12 @@ AST_DOCUMENT_NODE_T* herb_parse(const char* source, parser_options_T* options) {
   lexer_T lexer = { 0 };
   lexer_init(&lexer, source);
   parser_T parser = { 0 };
-  herb_parser_init(&parser, &lexer, options);
+
+  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
+
+  if (options != NULL) { parser_options = *options; }
```
marcoroth, line 38: Is there a reason why we are doing this here in `herb_parse` and not in `herb_parser_init` like we did before?

timkaechele: I didn't want to concern the init function with falling back to defaults. The core functionality shouldn't concern itself anymore with filling in ommitted parameters and working around them, intsead they get an options struct and work based on it. 

The way I saw it, `herb_parse` is an interface that lets you forego the options in favor of herb picked defaults, it's a convenient interface, but that interface choice shouldn't trickle down to the lower levels of the parser.

PR #632, opened 2025-10-13
C: Make lexer source an instance of `hb_string_T`

timkaechele:
This PR makes the lexer source an instance of `hb_string_T` and adapts all usages of the lexer source.

This is purely a refactoring and should not alter any interfaces, or observable behavior for that matter.

### DIFF
--- a/src/include/lexer_struct.h
+++ b/src/include/lexer_struct.h
@@ -2,6 +2,7 @@
 #define HERB_LEXER_STRUCT_H
 
 #include "util/hb_string.h"
+
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


src/include/lexer_struct.h
```diff
@@ -1,6 +1,7 @@
 #ifndef HERB_LEXER_STRUCT_H
 #define HERB_LEXER_STRUCT_H
 
+#include "util/hb_string.h"
```
marcoroth, line 4:
```suggestion
#include "util/hb_string.h"

```

PR #656, opened 2025-10-15
C: Use `hb_string_T` in `lexer_peek_helpers.c`

timkaechele:
This PR migrates the lexer peek helper interfaces to use `hb_string_T` instead of null terminated strings.

### DIFF
--- a/src/lexer_peek_helpers.c
+++ b/src/lexer_peek_helpers.c
@@ -17,17 +17,14 @@ char lexer_peek(const lexer_T* lexer, uint32_t offset) {
 }
 
 bool lexer_peek_for(const lexer_T* lexer, uint32_t offset, hb_string_T pattern, const bool case_insensitive) {
-  for (uint32_t index = 0; index < pattern.length; index++) {
-    const char character = lexer_peek(lexer, offset + index);
+  hb_string_T remaining_source = hb_string_slice(lexer->source, lexer->current_position + offset);
+  remaining_source.length = MIN(pattern.length, remaining_source.length);
 
-    if (case_insensitive) {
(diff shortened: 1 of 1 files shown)


marcoroth (commented):
Thanks for the pull request @timkaechele!

I'm struggling to see the motivation for this migration here. Is this so we more consistently use `hb_string_T` over `char*`, or is there something else I'm missing?

timkaechele:
> Thanks for the pull request @timkaechele!
> 
> I'm struggling to see the motivation for this migration here. Is this so we more consistently use `hb_string_T` over `char*`, or is there something else I'm missing?

Mostly for consistency, and because I think we can later refactor the `lexer_peek_for` function to utilize our string helper functions.

src/lexer_peek_helpers.c
```diff
@@ -3,73 +3,71 @@
...
+char lexer_peek(const lexer_T* lexer, uint32_t offset) {
   return lexer->source.data[MIN(lexer->current_position + offset, lexer->source.length)];
 }
 
-bool lexer_peek_for(const lexer_T* lexer, const int offset, const char* pattern, const bool case_insensitive) {
-  for (int index = 0; pattern[index]; index++) {
-    const char character = lexer_peek(lexer, offset + index);
+bool lexer_peek_for(const lexer_T* lexer, uint32_t offset, hb_string_T pattern, const bool case_insensitive) {
+  hb_string_T remaining_source = hb_string_slice(lexer->source, lexer->current_position + offset);
+  remaining_source.length = MIN(pattern.length, remaining_source.length);
```
marcoroth, line 21: Nice, this is elegant! 😍

PR #658, opened 2025-10-15
C: Use arena allocator in lexer

timkaechele:
This PR refactors the lexer to take an arena allocator, this way it is available once we need it later.

### DIFF
--- a/src/util/hb_system.c
+++ b/src/util/hb_system.c
@@ -4,27 +4,27 @@
 #define _GNU_SOURCE
 #endif
 
-#ifdef HB_USE_MALLOC
-#include <stdlib.h>
-#else
+#ifdef HB_USE_MMAP
 #include <sys/mman.h>
+#else
+#include <stdlib.h>
 #endif
 
 void* hb_system_allocate_memory(size_t size) {
-#ifdef HB_USE_MALLOC
-  return malloc(size);
-#else
+#ifdef HB_USE_MMAP
   void* memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
   if (memory == MAP_FAILED) { return NULL; }
 
   return memory;
+#else
+  return malloc(size);
 #endif
 }
 
 void hb_system_free_memory(void* ptr, size_t size) {
-#ifdef HB_USE_MALLOC
-  free(ptr);
-#else
+#ifdef HB_USE_MMAP
   munmap(ptr, size);
+#else
+  free(ptr);
 #endif
 }


src/token.c
```diff
@@ -13,8 +14,8 @@ size_t token_sizeof(void) {
   return sizeof(struct TOKEN_STRUCT);
 }
 
-token_T* token_init(const char* value, const token_type_T type, lexer_T* lexer) {
-  token_T* token = calloc(1, token_sizeof());
+token_T* token_init(hb_arena_T* allocator, const char* value, token_type_T type, lexer_T* lexer) {
+  token_T* token = hb_arena_alloc(allocator, token_sizeof());
```
marcoroth: Couldn't we get the `allocator`  from the `lexer` argument that we are already passing to `token_init`?

That way we wouldn't need to update all call sites to pass the allocator to it too.
```suggestion
token_T* token_init(const char* value, token_type_T type, lexer_T* lexer) {
  token_T* token = hb_arena_alloc(lexer->allocator, token_sizeof());
```

src/herb.c
```diff
@@ -11,8 +12,11 @@
 #include <stdlib.h>
 
 hb_array_T* herb_lex(const char* source) {
+  hb_arena_T allocator;
+  hb_arena_init(&allocator, MB(512));
```
marcoroth: I think we have to drastically reduce this, as this is allocating 512 MB for each `herb_lex` call, which usually happens quite frequently. Right now it's also failing the `docs` build with a `ENOMEM` exception.

https://github.com/marcoroth/herb/actions/runs/18542026700/job/52851618693#step:13:2045

I guess ideally `herb_lex` also takes an allocator so the arena can be shared between multiple lex calls.

timkaechele: Yep, I just wanted to get it back to compiling for the draft.

Ideally we pass in the arena both to `herb_lex` and `herb_parse`.

src/token.c
timkaechele: TL;DR; Yes, we can use `lexer->allocator`.

---

Hmm, I must say, I am not the biggest fan of passing the lexer to the `token_init` function and the init function then changing the lexer state. 

I understand why we do it, it's more convenient, and a bit less error prone, (you can't just forget to move on to the next line, the `token_init` function does that for you) but it feels very side effect-y. 

I briefly worked on removing the lexer dependency, hence I understand why it is in there, it's more complicated if you do the lexer state changes outside of token init, but all in all I didn't want to introduce even more coupling.

marcoroth:
I think we can close this in favor of #726. But maybe we can repurpose it for the `HB_USE_NMAP` flag.

marcoroth:
Closing in favor of #726 and #1287, thanks! 🙏🏼

PR #676, opened 2025-10-18
C: Inline `size_t_to_string` function

timkaechele:
This PR removes the `size_t_to_string`  function and replaces its only usages the function body.

## Reasoning

- `size_t_to_string` is only used in one place
- it is rather trivial
- if we would migrate it to use the arena allocator we would need to pass an explicit arena allocator to the pretty print function
- Using a stack allocated char array has better cache locality


marcoroth (approved):
Thank you @timkaechele! 🙏🏼

Issue #480, opened 2025-09-05
Linter: Recommends Rails methods in a non-Rails project


jeremyevans:
Example:

`[warning] Prefer image_tag helper over manual <img> with dynamic ERB expressions. Use <%= image_tag "#{Config.base_url}/logo-primary.png", alt: "..." %> instead.`

This only makes sense in a Rails project. If the goal is to make the herb-linter non-Rails specific, this warning is not appropriate. At least, there should be an option to disable any Rails-specific warnings.

marcoroth:
Hey @jeremyevans, thanks for reporting!

Definitely. The Herb Linter should not just be Rails-only. Currently we don't have a good way to selectively enabling/disabling or even configuring Linter Rules.

This is something that we need to tackle in general, also to address issues like https://github.com/marcoroth/herb/issues/313 and https://github.com/marcoroth/herb/issues/400

PR #677, opened 2025-10-18
Linter: Implement `turbo-permanent-no-misleading-value` rule

cb341:
## Summary

Add a new linter rule `html-turbo-permanent` to validate correct usage of the `data-turbo-permanent` boolean attribute in HTML files.

## Context

In HTML, boolean attributes are considered true when present, regardless of their assigned value.  
The following examples are functionally identical:

```html
<a data-turbo-permanent />
<a data-turbo-permanent="" />
<a data-turbo-permanent="true" />
<a data-turbo-permanent="false" />
````

This makes `data-turbo-permanent="false"` misleading since it behaves as if the attribute were true.
The new rule prevents such confusion by flagging `data-turbo-permanent="false"` as an error.

## Rule Behavior

**Allowed**

```html
<div data-turbo-permanent></div>
<div data-turbo-permanent=""></div>
<div data-turbo-permanent="true"></div>
<div data-turbo-permanent="foo"></div>
```

**Disallowed**

```html
<div data-turbo-permanent="false"></div>
```

## Rationale

`data-turbo-permanent` marks elements that persist across Turbo Drive navigations.
Explicitly setting the value to `"false"` does not disable this behavior and introduces ambiguity.
This rule enforces clear, unambiguous use of the attribute.

## References

* [Turbo: Permanent Elements](https://turbo.hotwired.dev/handbook/drive#permanent-elements)
* [MDN: Boolean Attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML)


marcoroth:
Hey @CuddlyBunion341, thanks for working this pull request!

While this is a great rule, I wonder if this should be reserved for a potential `turbo-lint` package, similar to how we put the Stimulus related rules in its own [`stimulus-lint` package](https://github.com/marcoroth/herb/tree/main/javascript/packages/stimulus-lint).

That said, we also have to think about how we deal with Rails and ActionView-specific rules, like https://github.com/marcoroth/herb/issues/480.

So I wonder if it would make more sense if we have some kind of "built-in plugins" or "set of plugins", like `stimulus/*`, `turbo/*`, `rails/*`, `actionview/*` etc. which are disabled by default and that you can manually enable in the upcoming config file. Or that they get enabled when the Linter sees the presence of a dependency in the `Gemfile` or `package.json`.

cb341:
> Hey @CuddlyBunion341, thanks for working this pull request!
> 
> While this is a great rule, I wonder if this should be reserved for a potential `turbo-lint` package, similar to how we put the Stimulus related rules in its own [`stimulus-lint` package](https://github.com/marcoroth/herb/tree/main/javascript/packages/stimulus-lint).
> 
> That said, we also have to think about how we deal with Rails and ActionView-specific rules, like #480.
> 
> So I wonder if it would make more sense if we have some kind of "built-in plugins" or "set of plugins", like `stimulus/*`, `turbo/*`, `rails/*`, `actionview/*` etc. which are disabled by default and that you can enable manually in the upcoming config file. Or that thy get enabled when the Linter sees the presence of a dependency in the `Gemfile` or `package.json`.

Thanks for the suggestion regarding the `turbo-lint` package.
I see how it can be valuable to group related packages. And allow herb to be used on non-rails projects.
Sounds like quite the task introducing a new package and rules for enabling rules. I very much support this idea.

I'd expect a PR that implements just the boilerplate of the turbo package, which I could merge my rules into.
The custom stimulus lint logo is awesome by the way.

cb341:
With the project specific rule activation I can see some problems.
Like for example determining project root or what Gemfile to use in case of nested ruby projects.

We would need to cover Gemfiles, Plugingemfiles, environments..
Also there could be multiple `package.json`s.

I think for now the easiest approach would be to enable all plugins by default (providing the best experience for RoR) and enable users to deactivate certain plugins.

cb341:
@marcoroth

I have created a new package `turbo-lint`, inspired by the already existing `stimulus-lint` package.
Alongside a simple cli which allows you to process a singular file at the moment.

I'd love to hear your thoughts regarding the namespacing and the boilerplate.
I also took the liberty of restricting the rule to only allow `data-turbo-permanent` without a value.

marcoroth:
Thanks for exploring this, @CuddlyBunion341. I checked, and sadly we cannot have the `turbo-lint` package name on NPM, as it's too close to `turbolint`.

But then I was wondering if we should even have its own package/linter CLI. I'm more and more leaning towards migrating `stimulus-lint` to a `herb-linter-stimulus`, `herb-plugin-stimulus`, `@herb-tools/stimulus` or `@herb-tools/stimulus-plugin` package, so that you can configure in the `.herb.yml` like:

```yaml
linter:
  plugins:
    - herb-plugin-stimulus
```

And then just let it run as part of the regular `herb-lint` command. Similar to the `rewriters` config for the formatter (https://github.com/marcoroth/herb/pull/759).

As there are quite a lot of moving pieces right now, I'd like to push this back a bit after Herb v0.8.0, thank you!

marcoroth:
Hey @cb341, sorry for the lack of progress here. 

I just left an update in a similar pull request here: https://github.com/marcoroth/herb/pull/1229#issuecomment-3947340560

I'd suggest that for the meantime that we avoid creating a new `turbo-lint` package and just include the Turbo rules by default in the Herb Linter for now to get things moving.

Would you mind renaming the rule starting with `turbo-*`. Maybe something like: `turbo-permanent-boolean-attribute` or `turbo-permanent-no-misleading-value` so we could later scope the rules by the `turbo-*` prefix.

Sorry for the back and forth here. I think it's better to make some progress here over holding it back for a plugin system that might anyway be overkill to start with, thank you! 🙏🏼

marcoroth:
If it's easier, we can also leave this pull request as-is so we can keep the work on the `turbo-lint` package. In that case feel free to open a new pull request with just the `data-turbo-permanent` linter rule. But either way is fine, whatever is more comfortable to you 🙌🏼

cb341:
@marcoroth 

Thanks for the info regarding the turbo lint package.

I'll rename the rule when I find the time.
Thanks for your patience 🙌

cb341:
@marcoroth thanks for your patience 🙌

I have reverted the linter package changes and renamed the rule.
I have also configured autofixing and tested the rule End to End in one of our projects.

javascript/packages/linter/src/rules/turbo-permanent-no-misleading-value.ts
```diff
@@ -0,0 +1,64 @@
...
+        node: attribute
+      }
+    )
+  }
+}
+
+export class TurboPermanentNoMisleadingValueRule extends ParserRule<TurboPermanentAutofixContext> {
+  static autocorrectable = true
+  static ruleName = "turbo-permanent-no-misleading-value"
+  static introducedIn = this.version("0.9.0")
```
marcoroth:
```suggestion
  static introducedIn = this.version("unreleased")
```

marcoroth (approved):
Thanks for this contribution @cb341! 🙏🏼 

I updated the rule a bit to also allow the detect `data-turbo-permanent` attributes on Action View Helpers.

But this is great, thank you! 🙌🏼

PR #678, opened 2025-10-18
C: Use `hb_string_T` in `herb_analyze` function

timkaechele:
This PR starts using the `hb_string_T` in the interface of `herb_analyze`.  

This is a side effect free refactor, that makes the switch to `hb_string_T` based token values easier later on.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -11,6 +11,7 @@
 #include "include/token_struct.h"
 #include "include/util.h"
 #include "include/util/hb_array.h"
+#include "include/util/hb_string.h"
 #include "include/visitor.h"
 
 #include <prism.h>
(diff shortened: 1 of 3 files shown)


src/analyze.c
```diff
@@ -52,7 +53,7 @@ static bool analyze_erb_content(const AST_NODE_T* node, void* data) {
     const char* opening = erb_content_node->tag_opening->value;
 
     if (strcmp(opening, "<%%") != 0 && strcmp(opening, "<%%=") != 0 && strcmp(opening, "<%#") != 0) {
-      analyzed_ruby_T* analyzed = herb_analyze_ruby(erb_content_node->content->value);
+      analyzed_ruby_T* analyzed = herb_analyze_ruby(hb_string_from_c_string(erb_content_node->content->value));
```
marcoroth, line 56: I wonder, if we eventually [should change the `StringField` in `template.rb` to just be a `hb_string_T`](https://github.com/marcoroth/herb/blob/25fd2f38e69ef315849a66988330667b0c150677/templates/template.rb#L106) on all AST Nodes, so we don't have to always convert them when using them.

marcoroth (approved):
Thanks @timkaechele! 🙏🏼

src/analyze.c
timkaechele: Long term that's definitely the plan. 

I just try to get there in small steps so that we won't have a giant PR that is hard to test, ideally we can always easily check for regressions.

PR #679, opened 2025-10-18
C: Remove unused `pretty_print_analyzed_ruby`

timkaechele:
This PR removes the unused `pretty_print_analyzed_ruby` function

### DIFF
--- a/src/include/pretty_print.h
+++ b/src/include/pretty_print.h
@@ -100,6 +100,4 @@ void pretty_print_errors(
   hb_buffer_T* buffer
 );
 
-void pretty_print_analyzed_ruby(analyzed_ruby_T* analyzed, const char* source);
-
 #endif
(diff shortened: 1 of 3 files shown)


marcoroth:
I wonder if we should rip out that method, given it's not even used right now.

timkaechele:
~~It's actually used in [src/analyze.c:58](https://github.com/marcoroth/herb/blob/main/src/analyze.c#L58)~~

Yes, we can remove it. 😅

src/include/pretty_print.h
```diff
@@ -5,6 +5,7 @@
 #include "ast_nodes.h"
 #include "location.h"
 #include "util/hb_buffer.h"
+#include "util/hb_string.h"
```
marcoroth: Can we remove this one too?
```suggestion
```

PR #680, opened 2025-10-18
Linter: Allow multiple rules to be ignored using `<%# herb:disable %>`

### LINKED ISSUE #665: Linter: Allow disabling multiple rules on a line
Follow up on #531 

Allow disabling multiple rules on a line, maybe with a syntax like:

```erb
<%# herb:disable rule-1 rule-2 %>
```

domingo2000:
Now multiple rules can be ignored with the following syntax

1. `<%# herb:disable rule-1, rule-2 %>`
2. `<%# herb:disable all %>` will disable all rules in that line.

This follows the same convention [proposed by standardrb](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-errors)

Resolves #665

### DIFF
--- a/javascript/packages/linter/README.md
+++ b/javascript/packages/linter/README.md
@@ -284,7 +284,7 @@ JSON output fields:
 - `clean`: Whether there were no offenses (`null` when `completed=false`)
 - `message`: Error or informational message (`null` on success)
 
-### Disabling Rules Inline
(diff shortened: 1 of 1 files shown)


domingo2000:
@marcoroth Documentation for this feature is still pending, but I believe it’s best if you write it wherever you find most appropriate.

Issue #667, opened 2025-10-16
Linter: Allow a TODO list for linter rules


domingo2000:
It should be very usefull if the linter CLI provides a way to ignore known offenses to improve adoption. This is very well described in [StandardRb todo list](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-every-violation-and-converting-them-into-a-todo-list). 

Context: In our Buk Monolith we have thousand of errors when turning on the linter, this feature will help a lot with adoption of the linter, because it will stop feature violations while beign able to mark current violations as TODO. 

This Issue is dependant on a standard Herb Config file. 

My thoughts on this are:

1. Add a TODO section in a herb config file
2. Read that TODO section from the linter, so the cases are ignored when running on the CLI.
3. Don't read that TODO list in the LSP server so the rules are not skipped (This is a nice to have but not _necessary_).

With that something like

```bash
herb-lint --generate-todo
```

Should be enough to adopt herb faster, avoid new offenses and be able to integrate to CI pipelines early.

marcoroth:
Good one, thanks for opening this issue @domingo2000! 🙌🏼

domingo2000:
I researched other implementations, here are some references to take ideas:
1. [Eslint baseline/suppressions support RFC](https://github.com/eslint/rfcs/tree/main/designs/2024-baseline-support)
2. [Ruff Issue on baseline support](https://github.com/astral-sh/ruff/issues/1149)
3. [Standardrb todo list](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-every-violation-and-converting-them-into-a-todo-list)

markokajzer:
I started looking into this a bit, but then I noticed that this is maybe already done?

From what I can see domingo2000 opened #685 on October 28th.
On March 30th, marcoroth commented that the approach was closed _in favor_ of #1485, suggesting that it is an alternate path that solves the same issue (therefore closing this task).

If not, what's missing? I'd gladly look into it 🙏

marcoroth:
@markokajzer https://github.com/marcoroth/herb/pull/1485 is not a 100% replacement for the proposed feature. The TODO list is more of a feature to say "I want to adopt this rule and keep enforcing this rule for newly written code, but that the linter should ignore _this_ list of known offenses that we couldn't get around to fix yet".

RuboCop has a similar feature:
https://github.com/rubocop/rubocop/blob/master/.rubocop_todo.yml

[...]

PR #685, opened 2025-10-18
Linter: Use `--generate-todo` to create a todo list for ignoring rules.

domingo2000:
Solves #667

Use `--generate-todo` to create a todo list that will ignore all the current offenses based on an offense count. Very similar of how [rubocop does](https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration). 

The implementation writes to a `.herb-todo.yml` file, that stores a map of all the files and rules with the counts of offenses and warnings. When the linter runs, reads from that file and ignore all ocurrences declarated as `exclude`.

The file is not dependant on a Herb Config file, because is autogenerated and can be very big, it should be consider as "private". 

The `--generate-todo` name comes from [standardrb recommendation on todo list.](https://github.com/standardrb/standard?tab=readme-ov-file#ignoring-every-violation-and-converting-them-into-a-todo-list)

Usage:

```bash
herb-lint --regenerate-todo <path> # Creates the file and show the offenses.
herb-lint <path> # Pass the linting without errors nor warnings.
```

## TODO

- [ ] Write documentation

### DIFF
--- a/javascript/packages/linter/src/cli/argument-parser.ts
+++ b/javascript/packages/linter/src/cli/argument-parser.ts
@@ -22,7 +22,8 @@ export interface ParsedArguments {
   wrapLines: boolean
   truncateLines: boolean
   useGitHubActions: boolean
-  fix: boolean
+  fix: boolean,
+  generateTodo: boolean,
 }
 
 export class ArgumentParser {
@@ -48,6 +49,7 @@ export class ArgumentParser {
       --no-timing      hide timing information
       --no-wrap-lines  disable line wrapping
       --truncate-lines enable line truncation (mutually exclusive with line wrapping)
+      --generate-todo   generate a .herb-todo.yml file with current diagnostics
   `
 
   parse(argv: string[]): ParsedArguments {
@@ -66,7 +68,8 @@ export class ArgumentParser {
         "no-color": { type: "boolean" },
         "no-timing": { type: "boolean" },
         "no-wrap-lines": { type: "boolean" },
-        "truncate-lines": { type: "boolean" }
+        "truncate-lines": { type: "boolean" },
+        "generate-todo": { type: "boolean" },
       },
       allowPositionals: true
     })
@@ -125,11 +128,23 @@ export class ArgumentParser {
       process.exit(1)
     }
 
+    const generateTodo = values["generate-todo"] || false
+
     const theme = values.theme || DEFAULT_THEME
     const pattern = this.getFilePattern(positionals)
     const fix = values.fix || false
 
-    return { pattern, formatOption, showTiming, theme, wrapLines, truncateLines, useGitHubActions, fix }
+    return { 
+      pattern,
+      formatOption,
+      showTiming,
+      theme,
+      wrapLines,
+      truncateLines,
+      useGitHubActions,
+      fix,
+      generateTodo,
+    }
   }
 
   private getFilePattern(positionals: string[]): string {
--- a/javascript/packages/linter/src/cli/file-processor.ts
+++ b/javascript/packages/linter/src/cli/file-processor.ts
@@ -7,6 +7,8 @@ import { colorize } from "@herb-tools/highlighter"
 
 import type { Diagnostic } from "@herb-tools/core"
 import type { FormatOption } from "./argument-parser.js"
+import { LintOffense } from "../types.js"
+import { LinterTodo } from "../linter-todo.js"
 
 export interface ProcessedFile {
   filename: string
@@ -19,6 +21,7 @@ export interface ProcessingContext {
   projectPath?: string
   pattern?: string
   fix?: boolean
+  generateTodo?: boolean
 }
 
 export interface ProcessingResult {
@@ -35,6 +38,7 @@ export interface ProcessingResult {
 
 export class FileProcessor {
   private linter: Linter | null = null
+  private linterTodo: LinterTodo | null = null
 
   private isRuleAutocorrectable(ruleName: string): boolean {
     if (!this.linter) return false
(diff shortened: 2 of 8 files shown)


javascript/packages/linter/src/cli/file-processor.ts
```diff
@@ -85,10 +100,13 @@ export class FileProcessor {
       }
 
       if (!this.linter) {
-        this.linter = new Linter(Herb)
+        this.linter = new Linter(Herb, undefined, this.linterTodo)
```
domingo2000, line 103: Passing `undefined` to let the linter set the default rules is not very idiomatic. Maybe it is better to pass the default rules on each call 🤷🏻 .

domingo2000:
@marcoroth happy to receive feedback on this. Currently a little WIP and can refactor a little more, but it works well, tried on buk-webapp monolith and is working good (4288 erb files).

Maybe i can follow up this PR with the following:
1. Show how much rules are in  `.herb-todo.yml`
2. Have a command flag that ignore the todo file. This is useful when the you want to get the linter run in all the project, without loosing the current todo.

javascript/packages/linter/src/linter-todo.ts
```diff
@@ -0,0 +1,142 @@
...
+    [ruleName: string]: {
+      [filePath: string]: {
+        warnings: number
+        errors: number
+      }
+    }
+  }
+}
+
+export class LinterTodo {
```
domingo2000, line 18: The name is not the best. Maybe `TodoList`? I belive that just `Todo` could fit too, but Todo could mean many more things.

javascript/packages/linter/src/cli/argument-parser.ts
```diff
@@ -48,6 +49,7 @@ export class ArgumentParser {
       --no-timing      hide timing information
       --no-wrap-lines  disable line wrapping
       --truncate-lines enable line truncation (mutually exclusive with line wrapping)
+      --generate-todo   generate a .herb-todo.yml file with current diagnostics
```
domingo2000, line 52: Should this be `--regenerate-todo` as rubocop, or `--generate-todo` as standardrb.  I think regenerate makes clear that you are destroying the old one. But generate-todo is a little less verbose and simple.

javascript/packages/linter/src/cli/file-processor.ts
marcoroth: Maybe we can refactor the Linter to take in an `options` object as the second argument. So it could be:

```suggestion
        this.linter = new Linter(Herb, { rules: customRules, linterTodo: this.linterTodo })
```

or just:
```suggestion
        this.linter = new Linter(Herb, { linterTodo: this.linterTodo })
```

javascript/packages/linter/src/cli/argument-parser.ts
marcoroth: Maybe we can have both so there is no surprise. 

`--generate-todo` can generate a file if none is present yet. If one is present it could back out and say something like: `.herb-todo.yml already exists, run "--regenerate-todo" to overwrite it`.

`--regenerate-todo` would just always overwrite the current one, no matter if `.herb-todo.yml` exists or not.

marcoroth:
When you run `--generate-todo` it concludes the run with:

```
 Rule offenses:
  html-no-space-in-tag (82 offenses in 24 files)
  html-no-self-closing (46 offenses in 3 files)
  html-input-require-autocomplete (27 offenses in 2 files)
  erb-require-trailing-newline (3 offenses in 3 files)


 Summary:
  Checked      210 files
  Files        28 with offenses | 182 clean (210 total)
  Offenses     158 errors | 0 warnings (158 offenses across 28 files)
  Fixable      158 offenses | 131 autocorrectable using `--fix`
  Start at     12:14:46
  Duration     1323ms (40 rules)
```

Maybe we can mention that we generated a `.herb-todo.yml` file here too.

On subsequent runs, I think it would also be great to mention the amount of "suppressed offenses", so people are aware that the linter is only passing because of the todos in the `.herb-todo.yml` file.


```
❯ herb-lint ../../../../rubyevents

 Summary:
  Checked      210 files
  Files        210 clean (210 total)
  Offenses     0 offenses
  Start at     12:15:49
  Duration     1066ms (40 rules)

 ✓ All files are clean!
```

marcoroth:
Another thing (which might not be as common in day-to-day use-cases), but I run the `herb-lint` out of my current Herb working directory, so I often use relative paths (but this could also happen when you run this within your app).

Running `herb-lint ../../../../rubyevents --generate-todo` generates the `.herb-todo.yml` file in the current working directory, and not at `../../../../rubyevents/.herb-todo.yml`. I think we might want to resolve the path argument and 

a) Put the `.herb-todo.yml` at the location you provide when running the linter (or if none is provided, use the current working directory like we do now)
b) Look at that path passed when running the linter for a config file first, then in the CWD
c) Inside the `.herb-todo.yml`: I think the paths should all be relative to the project root, currently they were generated as:

```yaml
excludes:
  html-no-space-in-tag:
    ../../../../rubyevents/app/views/topics/show.html.erb:
      warning: 0
      error: 1
    ../../../../rubyevents/app/views/templates/new.html.erb:
      warning: 0
      error: 3

# ...
```


Maybe we can also mention the absolute path of which `.herb-todo.yml` file the linter is using when running the the linter, something like:

```
npx herb-lint 

Using TODO Config file at: /Users/[user]/[project]/.herb-todo.yml
```

Thank you so much for working on this @domingo2000! 🙏🏼

marcoroth:
I'm closing this in favor of https://github.com/marcoroth/herb/pull/1485, thank you @domingo2000! 🙏🏼

PR #686, opened 2025-10-19
C: Migrate `is_void_element` to `hb_string_T`

timkaechele:
This PR changes the interface to the `is_void_element` function to use `hb_string_T` and adapts all call sites. 

This makes the switch to `hb_string_T` based token values easier later on.

### DIFF
--- a/src/html_util.c
+++ b/src/html_util.c
@@ -8,7 +8,6 @@
 #include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
-#include <strings.h>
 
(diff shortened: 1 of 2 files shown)


src/html_util.c
```diff
@@ -10,15 +11,19 @@
 #include <strings.h>
```
marcoroth: I think we might be able to remove the `strings.h` include now, given that we use `hb_string_equals_case_insensitive`
```suggestion
```

marcoroth (approved):
Thank you @timkaechele! 🙏🏼

PR #689, opened 2025-10-19
C: Use `hb_string_T` in `parser_check_matching_tag` function

timkaechele:
This PR changes the interface and implementation of `parser_check_matching_tag` to make use of the new `hb_string_T` struct and accompanying functions.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -936,7 +936,7 @@ static AST_HTML_ELEMENT_NODE_T* parser_parse_html_regular_element(
     close_tag = parser_parse_html_close_tag(parser);
   }
 
-  bool matches_stack = parser_check_matching_tag(parser, close_tag->tag_name->value);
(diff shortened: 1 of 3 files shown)


src/parser.c
```diff
@@ -936,7 +936,7 @@ static AST_HTML_ELEMENT_NODE_T* parser_parse_html_regular_element(
     close_tag = parser_parse_html_close_tag(parser);
   }
 
-  bool matches_stack = parser_check_matching_tag(parser, close_tag->tag_name->value);
+  bool matches_stack = parser_check_matching_tag(parser, hb_string_from_c_string(close_tag->tag_name->value));
```
marcoroth, line 939: For the future: I feel like it would be great if we could move the `hb_string_from_c_string` to the token-level, so that `value` always returns a `hb_string_T`.

marcoroth (approved):
Thanks, looks good to me @timkaechele! 🙏🏼

PR #693, opened 2025-10-19
C: Run `valgrind` in CI to catch memory leaks

timkaechele:
This PR starts performing a memory leak check as part of the CI.

It adds valgrind as an apt dependency, adds a script to run valgrind on all examples and adapts the github actions.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -30,7 +30,7 @@ errors:
 
     - name: UnexpectedTokenError
       message:
-        template: "Found `%s` when expecting `%s` at (%zu:%zu)."
+        template: "Found `%s` when expecting `%s` at (%u:%u)."
         arguments:
           - token_type_to_string(found->type)
           - token_type_to_string(expected_type)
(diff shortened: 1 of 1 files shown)


bin/leak_check_examples
```diff
@@ -0,0 +1,13 @@
...
+
+for path in examples/*
+do
+  if ! valgrind --error-exitcode=1 --leak-check=full ./herb parse $path > /dev/null
+  then
+    failure=1
+  fi
+done
+
+exit $failure
```
marcoroth, line 13: You might be able to simplify this using `set -e`

```suggestion
set -e  # Exit on error

for path in examples/*
do
  valgrind --error-exitcode=1 --leak-check=full ./herb parse $path > /dev/null
done
```

timkaechele: Does this exit early?

I wanted it to run all examples, so that in case someone who doesn't have access to valgrind (e.g. only having  mac) doesn't have to fix one leak, check CI and repeat. I rather have them see all failures at once.

marcoroth: ah I see. Yeah with this, it would exit immediately.

marcoroth:
This action is currently expected to fail, since we are resolving the remaining memory leaks in pull request https://github.com/marcoroth/herb/pull/698.

PR #696, opened 2025-10-20
C: Use `hb_string_T` in `parser_helpers.c` functions

timkaechele:
This PR refactors the `parser_get_foreign_content_closing_tag` and `parser_is_expected_closing_tag_name` to use `hb_string_T` instead of c strings.

### DIFF
--- a/src/parser_helpers.c
+++ b/src/parser_helpers.c
@@ -12,7 +12,6 @@
 #include "include/util/hb_string.h"
 
 #include <stdio.h>
-#include <strings.h>
 
 void parser_push_open_tag(const parser_T* parser, token_T* tag_name) {
   token_T* copy = token_copy(tag_name);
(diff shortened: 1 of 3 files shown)


src/parser_helpers.c
```diff
@@ -49,7 +48,8 @@ bool parser_in_svg_context(const parser_T* parser) {
     token_T* tag = (token_T*) hb_array_get(parser->open_tags_stack, i);
 
     if (tag && tag->value) {
-      if (strcasecmp(tag->value, "svg") == 0) { return true; }
+      hb_string_T tag_value_string = hb_string_from_c_string(tag->value);
+      if (hb_string_equals(tag_value_string, hb_string_from_c_string("svg"))) { return true; }
```
timkaechele, line 52: It will get better once we refactor token values to use `hb_string_T` as well. :sweat_smile:

PR #708, opened 2025-10-21
C: Implement page-based architecture for Arena Allocator

marcoroth:
This pull request refactors the `hb_arena_T` from a fixed-size design to a dynamically growing page-based implementation. The previous implementation had a hard capacity limit set at initialization. 

Instead of allocating one huge arena at initialization we now use a linked list of memory pages that grow automatically on demand. This allows us to start with a much smaller footprint and only allocate additional memory as actually needed, even though we are using `mmap`.  When the current page runs out of space, the allocator transparently allocates a new page and continues.

/cc @timkaechele

### DIFF
--- a/src/include/util/hb_arena.h
+++ b/src/include/util/hb_arena.h
@@ -4,15 +4,25 @@
 #include <stdbool.h>
 #include <stddef.h>
 
-typedef struct HB_ARENA_ALLOCATOR_STRUCT {
-  char* memory;
+typedef struct HB_ARENA_PAGE_STRUCT hb_arena_page_T;
+
+struct HB_ARENA_PAGE_STRUCT {
+  hb_arena_page_T* next;
   size_t capacity;
   size_t position;
+  char memory[];
+};
+
+typedef struct HB_ARENA_STRUCT {
+  hb_arena_page_T* head;
+  hb_arena_page_T* tail;
+  size_t default_page_size;
 } hb_arena_T;
 
-bool hb_arena_init(hb_arena_T* allocator, size_t size);
+bool hb_arena_init(hb_arena_T* allocator, size_t initial_size);
 void* hb_arena_alloc(hb_arena_T* allocator, size_t size);
 size_t hb_arena_position(hb_arena_T* allocator);
+size_t hb_arena_capacity(hb_arena_T* allocator);
 void hb_arena_reset(hb_arena_T* allocator);
 void hb_arena_reset_to(hb_arena_T* allocator, size_t new_position);
 void hb_arena_free(hb_arena_T* allocator);
--- a/src/util/hb_arena.c
+++ b/src/util/hb_arena.c
@@ -3,60 +3,179 @@
 #endif
 
 #include "../include/util/hb_arena.h"
+#include "../include/macros.h"
 
+#include <assert.h>
 #include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
 #include <sys/mman.h>
 
-#define KB(kb) (1024 * kb)
-#define MB(mb) (1024 * KB(mb))
+#define hb_arena_for_each_page(allocator, page)                                                                        \
+  for (hb_arena_page_T* page = (allocator)->head; page != NULL; page = page->next)
 
-bool hb_arena_init(hb_arena_T* allocator, size_t size) {
-  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+static inline size_t hb_arena_align_size(size_t size, size_t alignment) {
+  assert(size <= SIZE_MAX - (alignment - 1));
 
-  if (allocator->memory == MAP_FAILED) {
-    allocator->memory = NULL;
-    allocator->position = 0;
-    allocator->capacity = 0;
+  return ((size + (alignment - 1)) / alignment) * alignment;
+}
 
-    return false;
+static inline bool hb_arena_page_has_capacity(hb_arena_page_T* page, size_t required_size) {
+  assert(page->position <= page->capacity);
+
+  return page->position + required_size <= page->capacity;
+}
+
+static inline void* hb_arena_page_alloc_from(hb_arena_page_T* page, size_t size) {
+  assert(size > 0);
+  assert(page->position + size <= page->capacity);
+
+  void* result = &page->memory[page->position];
+  page->position += size;
+
+  return result;
+}
+
+static inline void hb_arena_page_reset(hb_arena_page_T* page) {
+  page->position = 0;
+}
+
+static inline void hb_arena_reset_pages_after(hb_arena_page_T* start_page) {
+  for (hb_arena_page_T* page = start_page; page != NULL; page = page->next) {
+    hb_arena_page_reset(page);
   }
+}
 
-  allocator->position = 0;
-  allocator->capacity = size;
+static bool hb_arena_append_page(hb_arena_T* allocator, size_t minimum_size) {
+  assert(minimum_size > 0);
+
+  size_t page_size = MAX(allocator->default_page_size, minimum_size);
+
+  assert(page_size <= SIZE_MAX - sizeof(hb_arena_page_T));
+  size_t total_size = page_size + sizeof(hb_arena_page_T);
+
+  hb_arena_page_T* page = mmap(NULL, total_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+  if (page == MAP_FAILED) { return false; }
+
+  *page = (hb_arena_page_T) { .next = NULL, .capacity = page_size, .position = 0 };
+
+  if (allocator->head == NULL) {
+    allocator->head = page;
+    allocator->tail = page;
+  } else {
+    hb_arena_page_T* last = allocator->head;
+
+    while (last->next != NULL) {
+      last = last->next;
+    }
+
+    last->next = page;
+    allocator->tail = page;
+  }
 
   return true;
 }
 
+bool hb_arena_init(hb_arena_T* allocator, size_t initial_size) {
+  assert(initial_size > 0);
+
+  allocator->head = NULL;
+  allocator->tail = NULL;
+  allocator->default_page_size = initial_size;
+
+  return hb_arena_append_page(allocator, initial_size);
+}
+
 void* hb_arena_alloc(hb_arena_T* allocator, size_t size) {
-  size_t required_size = ((size + 7) / 8) * 8; // rounds up to the next 8 bytes
+  assert(allocator->tail != NULL);
+  assert(size > 0);
 
-  if (allocator->position + required_size > allocator->capacity) { return NULL; }
+  size_t required_size = hb_arena_align_size(size, 8);
 
-  size_t offset = allocator->position;
-  void* ptr = &allocator->memory[offset];
-  allocator->position += required_size;
+  if (hb_arena_page_has_capacity(allocator->tail, required_size)) {
+    return hb_arena_page_alloc_from(allocator->tail, required_size);
+  }
 
-  return ptr;
+  for (hb_arena_page_T* page = allocator->tail->next; page != NULL; page = page->next) {
+    if (hb_arena_page_has_capacity(page, required_size)) {
+      allocator->tail = page;
+      return hb_arena_page_alloc_from(allocator->tail, required_size);
+    }
+  }
+
+  bool allocated = hb_arena_append_page(allocator, required_size);
+
+  if (!allocated) { return NULL; }
+
+  return hb_arena_page_alloc_from(allocator->tail, required_size);
 }
 
 size_t hb_arena_position(hb_arena_T* allocator) {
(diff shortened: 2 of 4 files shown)


src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
 #include "../include/util/hb_arena.h"
 
 #include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
 #include <sys/mman.h>
 
 #define KB(kb) (1024 * kb)
 #define MB(mb) (1024 * KB(mb))
+#define HB_ARENA_PAGE_SIZE_DEFAULT KB(4)
```
timkaechele: While 4kb is usually the OS page size, I think it's not advantageuous to make an arena page 4kb big. 

Be liberal here, a one mb page size wouldn't be outrageous and would reduce allocations tremendously.

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
+  if (size == 0) { return NULL; }
+
+  size_t required_size = ((size + 7) / 8) * 8;
+
+  char* result = allocator->tail->current;
+
+  uintptr_t current = (uintptr_t) allocator->tail->current;
+  uintptr_t delta = 0;
+
+  if (current % 8 != 0) { delta = 8 - (current % 8); }
```
timkaechele: Question: Why is this necessary?

I am not entirely sure why this is necessary. The calculation for `required_size` already pads the size to be a multiple of 8 [[ref]](https://github.com/marcoroth/herb/pull/708/files#diff-38530fcdce39968fc659335346b4b6c0d6be9af0f16e2ed1354ca3bb693b2461R35)



## Example

```c
#include <stdio.h>

int main(int argc, char const *argv[])
{
  for (size_t i = 0; i < 24; ++i)
  {
    size_t required_size = ((i + 7) / 8) * 8;
    printf("AllocSize: %zu, RequiredSize: %zu Mod: %zu\n",
      i,
      required_size,
      required_size % 8);
  }
}
```

Output: 

```
AllocSize: 0, RequiredSize: 0 Mod: 0
AllocSize: 1, RequiredSize: 8 Mod: 0
AllocSize: 2, RequiredSize: 8 Mod: 0
AllocSize: 3, RequiredSize: 8 Mod: 0
AllocSize: 4, RequiredSize: 8 Mod: 0
AllocSize: 5, RequiredSize: 8 Mod: 0
AllocSize: 6, RequiredSize: 8 Mod: 0
AllocSize: 7, RequiredSize: 8 Mod: 0
AllocSize: 8, RequiredSize: 8 Mod: 0
AllocSize: 9, RequiredSize: 16 Mod: 0
AllocSize: 10, RequiredSize: 16 Mod: 0
AllocSize: 11, RequiredSize: 16 Mod: 0
AllocSize: 12, RequiredSize: 16 Mod: 0
AllocSize: 13, RequiredSize: 16 Mod: 0
AllocSize: 14, RequiredSize: 16 Mod: 0
AllocSize: 15, RequiredSize: 16 Mod: 0
AllocSize: 16, RequiredSize: 16 Mod: 0
AllocSize: 17, RequiredSize: 24 Mod: 0
AllocSize: 18, RequiredSize: 24 Mod: 0
AllocSize: 19, RequiredSize: 24 Mod: 0
AllocSize: 20, RequiredSize: 24 Mod: 0
AllocSize: 21, RequiredSize: 24 Mod: 0
AllocSize: 22, RequiredSize: 24 Mod: 0
AllocSize: 23, RequiredSize: 24 Mod: 0
```

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
-  allocator->capacity = size;
+  allocator->tail = &allocator->head;
+  allocator->default_page_size = initial_size;
 
   return true;
 }
 
 void* hb_arena_alloc(hb_arena_T* allocator, size_t size) {
-  size_t required_size = ((size + 7) / 8) * 8; // rounds up to the next 8 bytes
+  if (size == 0) { return NULL; }
```
timkaechele: Suggestion: Nobody should ever allocate nothing. 

Consider replacing this check with an `assert`.

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
 #include <sys/mman.h>
 
 #define KB(kb) (1024 * kb)
 #define MB(mb) (1024 * KB(mb))
+#define HB_ARENA_PAGE_SIZE_DEFAULT KB(4)
 
-bool hb_arena_init(hb_arena_T* allocator, size_t size) {
-  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+bool hb_arena_init(hb_arena_T* allocator, size_t initial_size) {
+  if (initial_size == 0) { initial_size = HB_ARENA_PAGE_SIZE_DEFAULT; }
```
timkaechele: Suggestion: Remove default.

Making people think about what their initial size should be may be advantageuous here. 

Additionally it just looks weird to init an arena with an initial size 0.

src/include/util/hb_arena.h
```diff
@@ -4,15 +4,23 @@
...
 
-typedef struct HB_ARENA_ALLOCATOR_STRUCT {
-  char* memory;
-  size_t capacity;
-  size_t position;
+typedef struct HB_ARENA_PAGE_STRUCT {
+  struct HB_ARENA_PAGE_STRUCT* next;
+  char* start;
+  char* current;
+  char* end;
```
timkaechele: Question: Why use pointers here? 

IMO: Using capacity and position is usually a bit clearer and reduces the need for pointer shenanigans.

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
+    char* start = memory + sizeof(hb_arena_page_T);
+
+    *new_page = (hb_arena_page_T) { .next = NULL, .start = start, .current = NULL, .end = start + next_page_size };
+
+    allocator->tail->next = new_page;
+    allocator->tail = new_page;
+
+    result = allocator->tail->start;
+    next = result + required_size;
+  }
```
timkaechele: Suggestion: Make internal function

Consider creating a function for appending a new memory page to the allocator.

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
+    char* memory = mmap(NULL, total_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
 
-  size_t offset = allocator->position;
-  void* ptr = &allocator->memory[offset];
-  allocator->position += required_size;
+    if (memory == MAP_FAILED) { return NULL; }
 
-  return ptr;
+    hb_arena_page_T* new_page = (hb_arena_page_T*) memory;
+    char* start = memory + sizeof(hb_arena_page_T);
```
timkaechele: Suggestion: Reorder arena page members to use C99.

When you reorder the struct members to put the memory at the end you can use [flexible array members](https://stackoverflow.com/questions/3047530/flexible-array-member-in-c-structure)

```c
struct page {
  size_t capacity;
  size_t position;
  struct HB_ARENA_PAGE_STRUCT* next;
  uint8_t memory[];
} 
```

src/include/util/hb_arena.h
```diff
@@ -4,15 +4,23 @@
 #include <stdbool.h>
 #include <stddef.h>
 
-typedef struct HB_ARENA_ALLOCATOR_STRUCT {
-  char* memory;
-  size_t capacity;
-  size_t position;
+typedef struct HB_ARENA_PAGE_STRUCT {
+  struct HB_ARENA_PAGE_STRUCT* next;
```
timkaechele: Suggestion: Use forward declaration.

```c
typedef struct HB_ARENA_PAGE_STRUCT hb_arena_page_T;

struct HB_ARENA_PAGE_STRUCT { 
  hb_arena_page_T *next; 
  ...
}; 
```

src/util/hb_arena.c
```diff
@@ -5,58 +5,149 @@
...
+  if (current % 8 != 0) { delta = 8 - (current % 8); }
+
+  result += delta;
+
+  char* next = result + required_size;
+
+  if (next > allocator->tail->end) {
+    size_t next_page_size = allocator->default_page_size;
+
+    if (required_size > next_page_size) { next_page_size = required_size; }
```
timkaechele: Suggestion: Go bigger than the required size.

If you end up having to make the page bigger because of the required size, it may make sense to already allocate some more, so that the next alloc doesn't have to allocate yet another page.

src/util/hb_arena.c
```diff
@@ -3,60 +3,152 @@
...
+#define hb_arena_for_each_page(allocator, page)                                                                        \
+  for (hb_arena_page_T* page = (allocator)->head; page != NULL; page = page->next)
 
-bool hb_arena_init(hb_arena_T* allocator, size_t size) {
-  allocator->memory = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+static inline size_t hb_arena_align(size_t size, size_t alignment) {
+  return ((size + (alignment - 1)) / alignment) * alignment;
+}
+
+static inline bool hb_arena_page_is_full(hb_arena_page_T* page, size_t required_size) {
```
timkaechele: Naming: `has_capacity` may be more appropriate

Is full indicates that it is a general state, the page is full, period, there's no more space left. But this is not exactly what this function does, it asks whether there is still room for another allocation of size x

src/util/hb_arena.c
```diff
@@ -3,60 +3,152 @@
...
-    return false;
+  return result;
+}
+
+static inline void hb_arena_page_reset(hb_arena_page_T* page) {
+  page->position = 0;
+}
+
+static inline void hb_arena_reset_pages_after(hb_arena_page_T* start_page) {
+  for (hb_arena_page_T* page = start_page; page != NULL; page = page->next) {
```
timkaechele, line 44: Question: Can the `hb_arena_for_each_page` function be used here?

src/util/hb_arena.c
```diff
@@ -3,60 +3,152 @@
...
+    hb_arena_page_reset(page);
   }
+}
+
+static bool hb_arena_append_page(hb_arena_T* allocator, size_t minimum_size) {
+  size_t page_size = allocator->default_page_size;
+  if (minimum_size > page_size) { page_size = minimum_size; }
```
timkaechele: Suggestion: Use `MIN` macro

src/util/hb_arena.c
```diff
@@ -3,60 +3,152 @@
...
+    allocator->tail->next = page;
+  }
+
+  allocator->tail = page;
 
   return true;
 }
 
+bool hb_arena_init(hb_arena_T* allocator, size_t initial_size) {
+  if (initial_size == 0) { return false; }
```
timkaechele: Suggestion: Use `assert` here.

IMO setting `initial_size` to zero is more of a programming mistake than an actual situation, hence I would make it an assert.

This way the return value is only used for the allocation failure, not layer 8 issues.

src/util/hb_arena.c
```diff
@@ -3,60 +3,152 @@
...
+
+  return hb_arena_append_page(allocator, initial_size);
+}
+
 void* hb_arena_alloc(hb_arena_T* allocator, size_t size) {
-  size_t required_size = ((size + 7) / 8) * 8; // rounds up to the next 8 bytes
+  assert(size > 0);
 
-  if (allocator->position + required_size > allocator->capacity) { return NULL; }
+  size_t required_size = hb_arena_align(size, 8);
```
timkaechele: Suggestion: Rename to fit actual behavior 

Very important, `hb_arena_align` does not align the returned pointer but the **next pointer**.

You add more space than you need at the end, so that the next pointer is aligned to i.e. 8 bytes.

```
+-----------------------------------+----------------------------------- 
| alloc (4bytes)  | 4 bytes padding | NEXT ....
+-----------------------------------+-----------------------------------
^
|
This is the memory pointer we are going to return
```

See [`hb_arena_page_alloc_from`](https://github.com/marcoroth/herb/pull/708/files#diff-38530fcdce39968fc659335346b4b6c0d6be9af0f16e2ed1354ca3bb693b2461R24)

src/util/hb_arena.c
marcoroth: Sadly I don't think so, since the start page is different. Unless we tweak `hb_arena_for_each_page` to accept another argument for `start_page`.

src/util/hb_arena.c
marcoroth: I renamed this to `hb_arena_align_size` 👍🏼

src/util/hb_arena.c
marcoroth: What would the algorithm be in such a case? Just double the required size?

timkaechele: For the bufferes we used twice the size, but this could fire back if you have large buffers, maybe use 1.5

marcoroth: Opened https://github.com/marcoroth/herb/issues/724 to track this if we end up seeing it becoming a problem/bottleneck.

Issue #716, opened 2025-10-22
"Edit this page on GitHub" link is not working on linter rules pages


maxjacobson:
Hello!

- When I visit https://herb-tools.dev/linter/rules/erb-right-trim
- And I click on "Edit this page on GitHub"
- Then I arrive at https://github.com/marcoroth/herb/edit/main/docs/docs/linter/rules/erb-right-trim.md, which GitHub responds to with a 404

It seems that the correct URL would be https://github.com/marcoroth/herb/edit/main/javascript/packages/linter/docs/rules/erb-right-trim.md

I took a look at this: https://github.com/marcoroth/herb/blob/3906b1cda452330151693a5081f80a9af39fa342/docs/.vitepress/config/theme.mts#L133-L136

I'm not sure the correct change to make though, because I'm not familiar with vitepress. But I'm sure you can figure it out 😅

marcoroth:
Hey @maxjacobson, thanks for reporting!

Sadly, the problem is that these files only get created at build-time using this script here so we don't have to manually add each rule documentation page a second time in `docs/`:

https://github.com/marcoroth/herb/blob/3906b1cda452330151693a5081f80a9af39fa342/docs/.vitepress/generate-rules.mts#L1-L10

I wonder if Vitepress let's us fix the path for these special files, or allows us to remove the button in this case 🤔

maxjacobson:
I see! Thanks for the context

marcoroth:
Opened #718 to address this

maxjacobson:
🔥 sweet, thanks!

PR #717, opened 2025-10-22
Document an additional behavior of erb-right-trim

maxjacobson:
erb-right-trim flagged a number of instances where we're using `-%>` when we should be using `%>`. In fixing them, I'd like to be able to link my team to the herb docs where they can read more about the change, but the page doesn't currently mention this behavior of this linter. Seems worth mentioning.

Incidentally, this is the tweak I wanted to make when I noticed https://github.com/marcoroth/herb/issues/716 🙂

### DIFF
--- a/javascript/packages/linter/docs/rules/erb-right-trim.md
+++ b/javascript/packages/linter/docs/rules/erb-right-trim.md
@@ -6,6 +6,8 @@
 
 This rule enforces the use of `-%>` for right-trimming ERB output tags (like `<%= %>`) instead of `=%>`.
 
+Additionally, this rule enforces the use of `%>` instead of `-%>` or `=%>` when right-trimming has no effect.
+
 ## Rationale
 
 While `=%>` can be used for right-trimming whitespace in some ERB engines (like Erubi), it is an obscure and not well-defined syntax that lacks consistent support across most ERB implementations.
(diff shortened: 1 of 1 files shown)


maxjacobson:
I tried to put together a playground link showing the issues that I'm seeing locally, but surprisingly this is not flagged on the web 🤔 

https://herb-tools.dev/playground?tab=diagnostics&diagnosticsFilter=linter#DwUgxABANg9gxgQygZwFwQBQEoIFoQB8AUEcAA4EASAplLBAO4wBOUAJsAPQVA

But locally we are seeing things like this:

```
[error] Right-trimming with -%> has no effect on non-output ERB tags. Use %> instead (erb-right-trim)

app/views/foo.html.erb:1:15

  →   1 │ <%# locals: () -%>
        │                ~~~
      2 │ <div class="usa-accordion">
      3 │   <h3 class="usa-accordion__heading">
```

Hm 🤔

marcoroth:
Hey @maxjacobson, sorry for the confusion.

The version you are running is v0.7.5, the playground is running on `main`.

The rule in v0.7.5 incorrectly flagged cases where `-%>` does have an effect on non-output ERB tags, which is why we removed that part of the rule in https://github.com/marcoroth/herb/pull/645. But it hasn't been released yet, it's going to be part of v0.8.0.

maxjacobson:
ah! whoops. sounds like this one can safely be closed then right?

marcoroth:
Yes, I think so. Sorry for the confusion. I hope I get to wrap up this big v0.8.0 release soon-ish here.

maxjacobson:
No worries, and no rush 🙂

PR #719, opened 2025-10-22
C: Implement merging arena pages

timkaechele:
Based on the arena pages proposed in #708 by @marcoroth, this PR builds on the concept and adds a more refined reset behavior that merges obsolete pages into one big page.

This has the advantage that we won't have a bunch of small pages, and instead have more contiguous memory available.

## How it works

```
# Initial state
| Page 1 (4kb) | - next > | Page 2 (4kb) | - next > | Page 1 (16kb) | 

# reset to  4kb
| Page 1 (4kb) | - next > | Page 2 (24kb) |

# reset to 0kb
| Page 1 (4kb) | - next > | Page 2 (24kb) |
```

For now the first page is never deallocated and merged into a bigger page for two reasons:

1. It's an edge case that needs to be handled.
2. If the allocation of a new arena page fails at least you still have the first page. 🤷‍♂️ 

Related: #1155

### DIFF
--- a/src/util/hb_arena.c
+++ b/src/util/hb_arena.c
@@ -116,6 +116,11 @@ void* hb_arena_alloc(hb_arena_T* allocator, size_t size) {
     }
   }
 
+  size_t page_size = MAX(allocator->default_page_size, required_size);
+  bool allocated = hb_arena_append_page(allocator, page_size);
+
+  if (!allocated) { return NULL; }
+
   return hb_arena_page_alloc(allocator->tail, required_size);
 }
 
@@ -141,6 +146,7 @@ size_t hb_arena_capacity(hb_arena_T* allocator) {
 
 void hb_arena_reset(hb_arena_T* allocator) {
   hb_arena_reset_to(allocator, 0);
+  allocator->allocation_count = 0;
 }
 
 void hb_arena_reset_to(hb_arena_T* allocator, size_t target_position) {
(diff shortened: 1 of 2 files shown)


src/util/hb_arena.c
```diff
@@ -4,59 +4,185 @@
...
+  assert(size > 0);
+
+  size_t required_size = hb_arena_align_size(size, 8);
 
-  if (allocator->position + required_size > allocator->capacity) { return NULL; }
+  if (!hb_arena_page_has_capacity(allocator->tail, required_size)) {
+    if (allocator->tail->next != NULL && hb_arena_page_has_capacity(allocator->tail->next, required_size)) {
+      allocator->tail = allocator->tail->next;
+    } else {
+      bool allocated = hb_arena_append_page(allocator, required_size);
```
timkaechele: TODO: Potential memory leak.

In a situation where the tail has a next page, but the page does not have a sufficient capacity , we would just override the page next pointer and leak the existing page that didn't have enough capacity to hold the value 

Reference: `append_page` implementation
https://github.com/marcoroth/herb/pull/719/files#diff-38530fcdce39968fc659335346b4b6c0d6be9af0f16e2ed1354ca3bb693b2461R78

marcoroth: I think that should be addressed with this: https://github.com/marcoroth/herb/compare/f1f5b6a9c95c05728ffae8dde5911aab5704010f..e1cbb82b82f97ddea4c2d61b6352a54769a0d13f

PR #729, opened 2025-10-24
C: Allow to control memory allocation strategy

timkaechele:
This PR introduces a compile time flag that allows to use `malloc` instead of `mmap` to acquire memory for the arena.


marcoroth (commented):
Thanks for extracting this to its own pull request, @timkaechele!

Should we extract `hb_system_allocate_memory` and `hb_system_free_memory` to it's own `hb_system` (or similar) file, just so we have less branching logic in `hb_arena.c`?

PR #744, opened 2025-10-27
Introduce project-specific linter configuration

noelblaschke:
Allows to define project-specific linter options in .herb-lsp/config.json.

ATTOW all maintainers of a project must use e.g. a vscode (fork) editor to share project specific linter settings via workspace settings. This PR adds the ability to configure linter settings in the config.json like for the formatter and prioritize these settings over editor specific configurations.

### DIFF
--- a/javascript/packages/language-server/README.md
+++ b/javascript/packages/language-server/README.md
@@ -111,13 +111,31 @@ npx @herb-tools/language-server --stdio
 
 The language server can be configured using a `.herb-lsp/config.json` file in your project root. This file is automatically created when the language server starts if it doesn't exist.
 
+### Linter Configuration
+
+You can configure formatting behavior by adding a `linter` section to your config:
+
+```json
+{
+  "version": "0.7.5",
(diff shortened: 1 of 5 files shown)


marcoroth:
Hey @noelblaschke, thank you so much for working on this!

Sadly we got a bit of unfortunate timing going on here, as I've been working on a new `.herb.yml` config file that should be solving the issue you are talking about. It is one of the last missing pieces for the Herb v0.8.0 release.

I'm going to check if I can fold your changes into my branch.

noelblaschke:
@marcoroth Don't worry. This was just a quick draft to make it work. It could also be a solution for in between 0.7.x and 0.8, so you don't need to rush 0.8.0. A migration step/script may also help users to transfer to the new .herb.yml.

marcoroth:
Thanks @noelblaschke, I just wanted to say thanks for trying to take this on and sorry for the overlap here.

We don't really need to migrate the `.herb-lsp/config.json` as there wasn't anything that made use of it.

PR #747, opened 2025-10-30
Introduce `@herb-tools/config` and `.herb.yml` configuration file

### LINKED ISSUE #349: Config: Herb Formatter cannot be enabled in Zed
It seems like the configuration in `.herb-lsp/config.json` has no effect when running the language server. Since the Herb Formatter is disabled by default it cannot be enabled by the config when running the LSP in Zed (and possibly any other editor which is not VS Code).

### LINKED ISSUE #400: Linter: Support for selectively enabling or disabling linter rules
Unless I missed something, it's currently not possible to choose specific rules when running the linter.

Some rules can be problematic in some codebases (e.g. `html-no-duplicate-ids` currently has some known issues, cf. #273 and #388), so it would be nice to be able to selectively enable or disable specific rules.

This could be done via command-line flags (Rubocop has `--only` and `--except` flags) and/or a configuration file (would make it easier to use the same configuration during local development and in CI).

### LINKED ISSUE #589: Config: How to configure herb rules
I have herb_ls in neovim installed. I would like to exclude some rules. I have .herb-lint.yml with following:

```yaml
excludedRules:
  - html-tag-name-lowercase
```

This dosn't work My question how to configure herb formatter and linter ?

Thanks

### LINKED ISSUE #709: Config: Ignore Node Modules and Vendor Bundle
The `HERB_FILES_GLOB` glob introduced in https://github.com/marcoroth/herb/pull/672 now also makes it so that files in `node_modules/` and `.vendor/bundle/` are included. I think we should update the glob to make sure it doesn't include files within these two folders by default.

marcoroth:
This pull request introduces a new `@herb-tools/config` package that helps manage a new `.herb.yml` configuration file that can be used by all the tools.

The Herb Linter and Herb Formatter are the only user-facing tools that are going to be supported initially. This pull request also updates the Language Server to fully respect this `.herb.yml` configuration file.

Additionally, it adds extensive integration for the Herb Visual Studio Code Extension, so that using and updating the `.herb.yml` configuration file is a really pleasant experience.

https://github.com/user-attachments/assets/dcffb7c2-bff6-4000-b92e-1c97a164acf7

Resolves #349 
Resolves #400
Resolves #589 
Resolves #709 
Resolves #744 

Enables #313
Enables #480

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -1,4 +1,4 @@
-# Configuration
+# Configuration <Badge type="tip" text="^0.8.0" />
 
 Herb uses a `.herb.yml` configuration file to customize how the tools behave in your project. This configuration is shared across all Herb tools including the linter, formatter, and language server.
 
--- a/javascript/packages/config/package.json
+++ b/javascript/packages/config/package.json
@@ -33,14 +33,14 @@
     "@herb-tools/core": "0.7.5",
     "yaml": "^2.3.4"
   },
+  "devDependencies": {
+    "zod": "^4.1.12",
+    "zod-validation-error": "^4.0.2"
+  },
   "files": [
     "package.json",
     "README.md",
     "src/",
     "dist/"
-  ],
-  "devDependencies": {
-    "zod": "^4.1.12",
-    "zod-validation-error": "^4.0.2"
-  }
+  ]
 }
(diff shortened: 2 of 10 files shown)


olivier-thatch:
Very exciting! 🎉

avucic:
@marcoroth Thanks for the great work! Eager to try. Any clue when this should be released? I think I can't install this version until is released through nvim mason.

marcoroth:
Hey @avucic, no promises, but I'm aiming to release Herb v0.8.0 before the end of next week! 🤞🏼

If you really can't wait, you could pull down the repo, build the packages locally and then point the mason config to your local package, but it's not super convenient, sadly. 

For the future, maybe we can also set up nightly builds that people could opt in to use, so that these releases aren't blocking.

PR #749, opened 2025-10-30
C: Add Node.js and WASM extension helpers for `hb_string_T`

timkaechele:
This PR adds helpers that allow to work with `hb_string_T` types in WASM and node apis.

The functions are later used in #750

### DIFF
--- a/javascript/packages/node/extension/extension_helpers.cpp
+++ b/javascript/packages/node/extension/extension_helpers.cpp
@@ -46,6 +46,13 @@ napi_value CreateString(napi_env env, const char* str) {
   return result;
 }
 
+napi_value CreateStringFromHbString(napi_env env, hb_string_T string) {
+  napi_value result;
+  napi_create_string_utf8(env, string.data, string.length, &result);
+  return result;
+}
+
+
 napi_value CreatePosition(napi_env env, position_T position) {
   napi_value result;
   napi_create_object(env, &result);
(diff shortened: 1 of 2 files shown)


javascript/packages/node/extension/extension_helpers.cpp
```diff
@@ -46,6 +46,13 @@ napi_value CreateString(napi_env env, const char* str) {
   return result;
 }
 
+napi_value CreateStringFromHbString(napi_env env, hb_string_T string) {
+  napi_value result;
+  napi_create_string_utf8(env, string.data, string.length, &result);
+  return result;
```
marcoroth, line 52: Should we use `CreateString` here?
```suggestion
  return CreateString(string.data);
```

timkaechele: We sadly can't use `CreateString` as it [uses `NAPI_AUTO_LENGTH`](https://github.com/marcoroth/herb/pull/749/files#diff-6f7e35780ed4dea5630d55c673c35560decf20f5c125cf895552f6da5fec0ce3R45) which expects a null terminated string [[ref]](https://nodejs.org/api/n-api.html), but our `hb_string_T`is not necessarily null terminated.

marcoroth: Ah, I see. That's a good point. Thanks for the context!

marcoroth (approved):
Thanks @timkaechele! 🙏🏼

Issue #648, opened 2025-10-15
Herb: Java Bindings


marcoroth:
[Following Prism's lead](https://github.com/ruby/prism/tree/main/java), it would be awesome to also have Herb bindings for [Java](https://www.oracle.com/java/).

- [Prism Java Docs](https://ruby.github.io/prism/java/)

marcoroth:
In case it's going to be useful, here's the early FFI implementation we ripped out in #751:

https://github.com/marcoroth/herb/tree/libherb-ffi

PR #751, opened 2025-10-30
Ruby: Remove `libherb` FFI Ruby Interface

timkaechele:
This PR removes the ffi wrapper called libherb in the herb ruby gem. 

It wasn't used or for that matter covered by tests.

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -12,8 +12,6 @@ target :lib do
   library "tempfile"
 
   ignore "lib/herb/cli.rb"
-  ignore "lib/herb/libherb.rb"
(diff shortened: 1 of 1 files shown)


marcoroth:
Currently the `libherb` Ruby FFI interface is not used. This code is from the days before I figured out how to properly write a Ruby extension and before I followed Prism's lead.

We'll merge this pull request and then put up a new branch contenting the revert commit of this squashed merge commit, in case we need it at a later point for the Java Bindings in #648.

marcoroth:
Pushed in https://github.com/marcoroth/herb/tree/libherb-ffi

PR #752, opened 2025-10-30
C: Use `hb_string_T` in `element_source_to_string` function

timkaechele:
This PR starts using the `hb_string_T` for strings when returning the element source string.

### DIFF
--- a/templates/ext/herb/nodes.c.erb
+++ b/templates/ext/herb/nodes.c.erb
@@ -39,7 +39,11 @@ static VALUE rb_<%= node.human %>_from_c_struct(<%= node.struct_type %>* <%= nod
   <%- when Herb::Template::ArrayField -%>
   VALUE <%= node.human %>_<%= field.name %> = rb_nodes_array_from_c_array(<%= node.human %>-><%= field.name %>);
   <%- when Herb::Template::ElementSourceField -%>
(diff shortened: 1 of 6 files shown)


templates/ext/herb/nodes.c.erb
```diff
@@ -39,7 +39,11 @@ static VALUE rb_<%= node.human %>_from_c_struct(<%= node.struct_type %>* <%= nod
   <%- when Herb::Template::ArrayField -%>
   VALUE <%= node.human %>_<%= field.name %> = rb_nodes_array_from_c_array(<%= node.human %>-><%= field.name %>);
   <%- when Herb::Template::ElementSourceField -%>
-  VALUE <%= node.human %>_<%= field.name %> = rb_utf8_str_new_cstr(element_source_to_string(<%= node.human %>-><%= field.name %>));
+  VALUE <%= node.human %>_<%= field.name %>;
+  {
+    hb_string_T element_source_string = element_source_to_string(<%= node.human %>-><%= field.name %>);
+    <%= node.human %>_<%= field.name %> = rb_utf8_str_new(element_source_string.data, element_source_string.length);
+  }
```
marcoroth, line 46: For the future, maybe we should also introduce an equivalent `create_string` function for the Ruby extension (like we did in #749).

PR #772, opened 2025-11-03
C: Add `compile_flags.txt`

timkaechele:
This PR adds a compile flags configuration [[ref](https://clangd.llvm.org/installation#compile_flagstxt)] for clangd. 

This way C files using prism are actually correctly resolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,6 @@ docs/docs/public/c-reference
 
 # NPM
 **/node_modules/**/*
+
+# Clangd config
+compile_flags.txt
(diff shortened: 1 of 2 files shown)


marcoroth:
Does it make sense to look at [`bear`](https://github.com/rizsotto/Bear) so we can automatically generate the files/flags needed?

timkaechele:
I am not sure what that would solve. The compile flags file is super basic, we could generate it using the makefile. Only dynamic part seems to be the check include path.

marcoroth:
Mostly so that we don't have to maintain these flags in multiple places. I'm just worried that we add this file now and then it gets out of date.

PR #777, opened 2025-11-04
C: Use `uint32_t` in `utf8.c`

timkaechele:
This PR makes all utf8 apis use `uint32_t` instead of the more generic `int`


timkaechele:
> I guess if we tighten up the return types, does it make sense to treat them as `uint8_t`?

I think we won’t gain much in terms of performance. 

Using `uint32_t` for all string lengths may be just a bit more straightforward

PR #782, opened 2025-11-05
C: New `hb_array` implementation

timkaechele:
This PR adds a new version of the `hb_array` called `hb_narray` that implements an array that can actually hold the elements it stores.

The existing `hb_array` only allows to store pointers to heap allocated elements, the new `hb_narray` acts as a container that also provides the backing memory for the elements it holds. 

It's important to note that you can still make the elements pointers if you want to.

## Why? 

- Improves locality of the elements
- Reduces the number of individual heap allocations for elements

## Next steps

1. Get this merged
2. Refactor existing `hb_array` usages
3. Remove `hb_array` and rename `hb_narray` to `hb_array` 

## Memory layout

herb_new_array

### DIFF
--- a/javascript/packages/node/binding.gyp
+++ b/javascript/packages/node/binding.gyp
@@ -40,6 +40,7 @@
         "./extension/libherb/util.c",
         "./extension/libherb/util/hb_arena.c",
         "./extension/libherb/util/hb_array.c",
+        "./extension/libherb/util/hb_narray.c",
         "./extension/libherb/util/hb_buffer.c",
         "./extension/libherb/util/hb_string.c",
         "./extension/libherb/util/hb_system.c",
--- a/src/include/util/hb_narray.h
+++ b/src/include/util/hb_narray.h
@@ -1,0 +1,29 @@
+#ifndef HERB_NARRAY_H
+#define HERB_NARRAY_H
+
+#include <stdint.h>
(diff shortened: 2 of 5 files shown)


src/include/util/hb_narray.h
```diff
@@ -0,0 +1,28 @@
+#ifndef HERB_ARRAY_H
+#define HERB_ARRAY_H
```
marcoroth:
```suggestion
#ifndef HERB_NARRAY_H
#define HERB_NARRAY_H
```

javascript/packages/node/binding.gyp
```diff
@@ -36,6 +36,7 @@
         "./extension/libherb/util.c",
         "./extension/libherb/util/hb_arena.c",
         "./extension/libherb/util/hb_array.c",
+        "./extension/libherb/util/hbn_array.c",
```
marcoroth:
```suggestion
        "./extension/libherb/util/hb_narray.c",
```

marcoroth:
I'm curious where you think this would come in handy. I think the problem is, at least for the AST Nodes, that they don't share the same item size, unless you have an array of a specific AST Node type.

marcoroth (approved):
Thank you @timkaechele! 🙏🏼

PR #799, opened 2025-11-07
C: Consistently use `hb_string_T` in lexer helper functions

timkaechele:
Based on the changes made in https://github.com/marcoroth/herb/pull/798 this PR changes the remaining functions in the lexer to use `hb_string_T`.

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -102,9 +102,9 @@ static void lexer_advance_by(lexer_T* lexer, const size_t count) {
   }
 }
 
-static token_T* lexer_advance_with(lexer_T* lexer, const char* value, const token_type_T type) {
-  lexer_advance_by(lexer, strlen(value));
-  return token_init(hb_string(value), type, lexer);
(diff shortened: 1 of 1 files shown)


src/lexer.c
```diff
@@ -120,7 +120,11 @@ static token_T* lexer_advance_with_next(lexer_T* lexer, size_t count, token_type
 }
 
 static token_T* lexer_advance_current(lexer_T* lexer, const token_type_T type) {
-  return lexer_advance_with(lexer, (char[]) { lexer->current_character, '\0' }, type);
+  char buffer[2];
+  buffer[0] = lexer->current_character;
+  buffer[1] = '\0';
```
marcoroth, line 125: Does it make sense to extract this as `hb_string_from_char()`?

timkaechele: I wouldn't do it, it's not a common case, and opens a can of worms when it comes to allocations and the lifetimes of those allocations.

marcoroth: 👍🏼

marcoroth (approved):

Issue #789, opened 2025-11-06
Formatter: herb-format breaks HTML having <span> and if-else block


tk0miya:
**Input:**
```erb
<span>
  <% if true %>
  <% else %>
  <% end %>
</span>

```

**Output:**
```erb
<span><% if true %><% end %><% end %></span>

```

"else" was vanished and "end" appeared twice.

**Expected:**
```erb
<span>
  <% if true %>
  <% else %>
  <% end %>
</span>
```
(not changed from input)

**Playground:**

https://herb-tools.dev/playground?tab=format#DwZwDghgdgfAUAAgcApAglgMwQFwE4CuApgivEqgkQDYglmLJpFQAmp8wA9ONPEA

Is this okay? playground says "ReferenceError: process is not defined" ...

**Additional context:**

I got the correctly formatted result when I replaced `<span>` with `<div>` or other tags (eg. `<p>`). It seems this bug is related with `<span>` tag.

timkaechele:
The output from `herb parse $filename`

```
timkaechele@fedora:~/dev/herb$ ./herb parse examples/bug.html.erb 
@ DocumentNode (location: (1,0)-(6,0))
└── children: (2)
    ├── @ HTMLElementNode (location: (1,0)-(5,7))
        ├── open_tag: 
            └── @ HTMLOpenTagNode (location: (1,0)-(1,6))
                ├── tag_opening: "<" (location: (1,0)-(1,1))
                ├── tag_name: "span" (location: (1,1)-(1,5))
                ├── tag_closing: ">" (location: (1,5)-(1,6))
                ├── children: []
                └── is_void: false

        ├── tag_name: "span" (location: (1,1)-(1,5))
        ├── body: (3)
            ├── @ HTMLTextNode (location: (1,6)-(2,2))
                └── content: "\n  "
            
            ├── @ ERBIfNode (location: (2,2)-(4,11))
                ├── tag_opening: "<%" (location: (2,2)-(2,4))
                ├── content: " if true " (location: (2,4)-(2,13))
                ├── tag_closing: "%>" (location: (2,13)-(2,15))
                ├── statements: (1)
                    └── @ HTMLTextNode (location: (2,15)-(3,2))
                        └── content: "\n  "

                ├── subsequent: 
                    └── @ ERBElseNode (location: (3,2)-(4,2))
                        ├── tag_opening: "<%" (location: (3,2)-(3,4))
                        ├── content: " else " (location: (3,4)-(3,10))
[...]

PR #809, opened 2025-11-08
Formatter: Fix formatting `<% else %>` in inline mode

marcoroth:
This pull request fixes a bug in the formatter, where we accidentally visited the `end_node` twice, instead of the subsequent node.

This following document
```erb
<span>
  <% if valid? %>
    Valid
  <% else %>
    Invalid
  <% end %>
</span>
```

was previously formatter to:

```erb
<span><% if valid? %>Valid<% end %><% end %></span>
```

Now it gets correctly formatted as:

```erb
<span><% if valid? %>Valid<% else %>Invalid<% end %></span>
```

Resolves https://github.com/marcoroth/herb/issues/789


tk0miya:
Thank you for the quick fix!

PR #830, opened 2025-11-10
Go: Add Herb Go FFI Bindings

omarluq:
This PR adds Go bindings for Herb using `c-for-go` (FFI), allowing Go
  applications to parse and analyze ERB templates with the same implementation
  as the native C, Ruby C-Extension, C++ Emscripten WASM, Java JNI, Rust FFI,
  and the C++ Node.js NAPI bindings.

  Building the Go bindings requires Go 1.20+, a C compiler, and c-for-go:

  ```bash
    cd go
    make build
  ```

  Usage:

  ```go
    package main

    import (
        "fmt"
        "github.com/marcoroth/herb"
    )

    func main() {
        template := "<h1><%= title %></h1>"

        // Lex
        tokens := herb.Lex(template)

        // Parse
        ast := herb.Parse(template, nil)

        // Extract Ruby
        ruby := herb.Extract(template, 0)
    }
  ```
  The Go bindings use a minimal wrapper header approach (similar to Java's JNI
  bindings) to avoid exposing prism dependencies while linking against both
  `libherb.a` and `libprism.a`


omarluq:
Will reopen later after multi backend PR

Issue #872, opened 2025-11-13
Incorrect MissingERBEndTagError


joelhawksley:
```erb
<span
  <% if @replaceable %>data-menu-button<% end %>
  class="css-truncate css-truncate-target"
></span>
```

Returns:

```
  1. MissingERBEndTagError at 2:2
     `<% if %>` started here but never closed with an end tag. The end tag may be in a different scope.
```

(Using v0.8.0)

I believe this is valid ERB/HTML.

marcoroth:
Thanks for reporting @joelhawksley! 

I just opened #873 to address this 🙌🏼

PR #873, opened 2025-11-13
Parser: Fix ERB control flow tags in attribute positions

### LINKED ISSUE #1063: Parser: Conditional boolean attribute attribute
The following
```
<details <% if @doc.registration_id.present? %>open<% end %>>
```
generates the following error:
```
<div class="herb-error-message">
   `&lt;% if %&gt;` started here but never closed with an end tag. The end tag may be in a different
</div>
```
referencing the opening `<%` tag.

The only way around the generation of this error is much longer & repetitive:
```
<% if @doc.registration_id.present? %>
  <details open>
   [...]
<% else %>
  <details open>
  [...]
<% end %>
```

marcoroth:
This pull request fixes the parser to correctly handle ERB control flow tags when parsing HTML attributes.

Previously, control flow tags like `<% if %>`, `<% end %>`, and `<%# comment %>` were incorrectly included as part of attribute names and values. Now, only ERB output tags (`<%=`, `<%==`) can be part of attribute names, while control flow tags properly terminate attribute name and attribute value parsing.

This enables patterns like this, where there is no space between the closing `%>` and the attribute name:

```erb
<dialog data-controller="dialog" <% if condition %>data-turbo-permanent<% end %>></dialog>
<div <% if admin? %>data-admin data-role="admin"<% end %>></div>
```

Resolves https://github.com/marcoroth/herb/issues/872
Resolves #1063

### DIFF
--- a/test/parser/attributes_test.rb
+++ b/test/parser/attributes_test.rb
@@ -196,6 +196,7 @@ module Parser
       assert_parsed_snapshot(%(<div attribute-name=>div-content</div>))
     end
 
+    # https://github.com/marcoroth/herb/issues/856#issuecomment-3525179643
     test "Conditional attribute with ERB control flow and no surrounding spaces" do
       assert_parsed_snapshot(%(<dialog data-controller="dialog" <% if local_assigns[:permanent] %>data-turbo-permanent<% end %>></dialog>))
     end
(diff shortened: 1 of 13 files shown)


src/parser.c
```diff
@@ -295,6 +295,11 @@ static AST_HTML_ATTRIBUTE_NAME_NODE_T* parser_parse_html_attribute_name(parser_T
     TOKEN_EOF
   )) {
     if (token_is(parser, TOKEN_ERB_START)) {
+      const char* tag = parser->current_token->value;
+      bool is_output_tag = (tag[2] == '=');
```
timkaechele: Question: What if the tag length is less than 3?

timkaechele: Nitpick: Remove the braces

marcoroth: Updated in [92f4d9a](https://github.com/marcoroth/herb/pull/873/commits/92f4d9ab86382b3b7f857ceb7055f4f1d75a6fb4), thanks!

Issue #874, opened 2025-11-14
Herb 0.8.0 Failed to build gem native extension


morgoth:
On our GH CI, dependabot tried to update gems and installing herb failed:
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory:
/home/runner/work/trmz/trmz/vendor/bundle/ruby/3.4.0/gems/herb-0.8.0/ext/herb
/opt/hostedtoolcache/Ruby/3.4.6/x64/bin/ruby extconf.rb
/opt/hostedtoolcache/Ruby/3.4.6/x64/lib/ruby/3.4.0/rubygems.rb:269:in
'Gem.find_spec_for_exe': can't find gem rake (>= 0.a) with executable rake
(Gem::GemNotFoundException)
from /opt/hostedtoolcache/Ruby/3.4.6/x64/lib/ruby/3.4.0/rubygems.rb:288:in
'Gem.activate_bin_path'
	from /opt/hostedtoolcache/Ruby/3.4.6/x64/bin/rake:25:in '<main>'
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--without-opt-dir
	--with-opt-include=${opt-dir}/include
	--without-opt-include
	--with-opt-lib=${opt-dir}/lib
	--without-opt-lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
	--curdir
	--ruby=/opt/hostedtoolcache/Ruby/3.4.6/x64/bin/$(RUBY_BASE_NAME)
extconf.rb:6:in 'Kernel#system': Command failed with exit 1: rake (RuntimeError)
	from extconf.rb:6:in 'block in <main>'
	from extconf.rb:5:in 'Dir.chdir'
	from extconf.rb:5:in '<main>'

extconf failed, exit code 1

Gem files will remain installed in
/home/runner/work/trmz/trmz/vendor/bundle/ruby/3.4.0/gems/herb-0.8.0 for
inspection.
Results logged to
/home/runner/work/trmz/trmz/vendor/bundle/ruby/3.4.0/extensions/x86_64-linux/3.4.0/herb-0.8.0/gem_make.out

/opt/hostedtoolcache/Ruby/3.4.6/x64/lib/ruby/3.4.0/rubygems/ext/builder.rb:125:in
[...]

PR #880, opened 2025-11-14
Allow Herb to be installed from source

andreaslillebo:
Thanks again for all the awesome work on the Herb. It is amazing 🤩

Previously Herb could only be installed if there was a precompiled gem available.

This changes this by vendoring Prism on compile time.

Allows workaround for issues like #874

### DIFF
--- a/ext/herb/extconf.rb
+++ b/ext/herb/extconf.rb
@@ -1,9 +1,53 @@
 # frozen_string_literal: true
 
 require "mkmf"
+require "bundler"
+require "fileutils"
+
+def vendor_prism(prism_vendor_path)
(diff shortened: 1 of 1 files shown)


marcoroth:
Hey @andreaslillebo, Prism already gets vendored with the gem that comes from RubyGems.org (see #515). We vendor Prism before we publish the gem using the `vendor:prism` rake task, defined in [Rakefile](https://github.com/marcoroth/herb/blob/d0806e7f5dcf3a39860e6e266a11df561ea6b357/Rakefile#L133-L174).

Thank you for the pull request! 🙏🏼

PR #885, opened 2025-11-15
C: Deprecate `hb_array_size`

timkaechele:
This PR deprecates the `hb_array_size` function in favor of directly accessing `hb_array_T.size`

As part of the deprecation, ast nodes now always have fully initialized array members.


marcoroth (commented):
Is there a reason why we don't remove `hb_array_size` with this right away?

timkaechele:
It’s going to be dropped as part of the migration to the new array implementation.

Issue #900, opened 2025-11-18
Compilation error on valid html/erb


morgoth:
I got a compilation error on herb 0.8.2, that I cannot reproduce on the playground.
Here are the details:

```
--- CONTENT --------------------------------------------------------------------
```erb
<p>
  <%= t(".entering") %>
  <p>
    <%= t(".manual") %>
  </p>
  <p>
    <%= t(".using_copy") %>
    <ul>
      <li>
        <%= t(".copy_step_1") %>
      </li>
      <li>
        <%= t(".copy_step_2") %>
      </li>
      <li>
        <%= t(".copy_step_3") %>
      </li>
    </ul>
  </p>
</p>
<p>
  <%= t(".finish") %>
</p>
```

--- COMPILATION ERROR ----------------------------------------------------------
```
NoMethodError: undefined method 'location' for an instance of Hash
```

--- BACKTRACE ------------------------------------------------------------------
```
/home/wojtek/.local/share/mise/installs/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/herb-0.8.2-x86_64-linux-gnu/lib/herb/engine/error_formatter.rb:104:in 'Herb::Engine::ErrorFormatter#format_error'
/home/wojtek/.local/share/mise/installs/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/herb-0.8.2-x86_64-linux-gnu/lib/herb/engine/error_formatter.rb:87:in 'block in Herb::Engine::ErrorFormatter#format_all_without_highlighter'
/home/wojtek/.local/share/mise/installs/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/herb-0.8.2-x86_64-linux-gnu/lib/herb/engine/error_formatter.rb:86:in 'Array#each'
/home/wojtek/.local/share/mise/installs/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/herb-0.8.2-x86_64-linux-gnu/lib/herb/engine/error_formatter.rb:86:in 'Enumerable#each_with_index'
/home/wojtek/.local/share/mise/installs/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/herb-0.8.2-x86_64-linux-gnu/lib/herb/engine/error_formatter.rb:86:in 'Herb::Engine::ErrorFormatter#format_all_without_highlighter'
[...]

PR #915, opened 2025-11-21
Engine: Fix incomplete `Hash` handling in `error_formatter`

christos:
Fixes `NoMethodError: undefined method 'location' for an instance of Hash`, as reported in #900, by adding proper Hash handling in `format_error()` method.

The bug occurs when:

1. Validation errors (Hash objects) are formatted
2. The highlighter is unavailable/disabled
3. `format_error()` is called directly (via `format_all_without_highlighter`)

The fix follows the same pattern as the rest of the formatting code, checking if the `error.is_a?(Hash)` 

Ideally both the Ruby Validator and the C Parser errors should be wrapped in a class with `.location`, `.message`, `.type` accessors, but that change is too big to tackle in this fix.

### DIFF
--- a/test/engine/error_handling_test.rb
+++ b/test/engine/error_handling_test.rb
@@ -200,9 +200,17 @@ module Engine
         Herb::Engine.new(template)
       end
 
-      assert_includes error.message, "Working"
-      assert_includes error.message, "wrong_tag"
-      assert_includes error.message, "←"
+      assert_includes error.message, "HTML+ERB Compilation Errors:"
+
+      assert_includes error.message, "MissingClosingTag"
+      assert_includes error.message, "Opening tag `<div>` at (1:1) doesn't have a matching closing tag `</div>` in the same scope."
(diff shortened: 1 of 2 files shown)


christos:
> Does that failure in CI make sense?

I think I forgot to update some tests. Let me re-do the PR and get `bundle exec rake test` to be all green.

christos:
@marcoroth I think I've fixed it. 

Made the assertions ignore the `format_all_without_highlighter` vs `format_all_with_highlighter` differences by matching just the errors, not the decoration.

All tests with `bundle exec rake test` are green now (with some warnings that were already there).

CleanShot 2025-12-01 at 11 18 24@2x

christos:
Weird. I see the failure on CI but `bundle exec mtest` won't fail locally. I am at a loss...

CleanShot 2025-12-01 at 11 39 30@2x

marcoroth:
Hey @christos, sorry for the trouble and thanks for exploring. The problem is that in the "JavaScript" build the highlighter package is available and that's why it's using the JS-based highlighter in those tests. 

This is brittle from my part, let me see if I can untangle this, thank you!

PR #944, opened 2025-12-03
Formatter: Disable formatter in default config

### LINKED ISSUE #910: Turn off fix-on-save/format-on-save with NeoVim?
Hi,

I cannot figure out how to turn off "fix-on-save" or "format-on-save" when using the language server with NeoVim. There are options to turn it off for VSCode, but none of them seem to work in the `.herb.yml` file, and the `.herb-lsp/config.json` file mentioned [here](https://herb-tools.dev/projects/language-server#configuration) does not seem to have any effect.

Not being able to turn this off makes Herb unusable for me because I'm working in a big codebase and I need to be able to be selective about what I fix/format.

Thanks,
Dan

### LINKED ISSUE #911: Formatter: VSCode: stop ignoring `formatOnSave`
Herb unfortunately ignores VSCode's setting to disable formatting on save:

```json
"[erb]": {
  "editor.formatOnSave": false
}
``` 

**Expected:**
+ file doesn't get formatted on save with the setting

**Actual:**
+ Herb ignores the setting and formats the file on save.

marcoroth:
This pull request changes the default configuration to not have the formatter enabled by default. Since the formatter is still experimental people should be opting in to use it. 

Since the default configuration had the formatter enabled by default this meant that the language server started formatting on save by default as well, which is not ideal.

This pull request addresses this, so that people have to opt-in and enable the formatter in teh `.herb.yml` configuration file. Additionally, it splits up the fix and format concerns in the language server, so that the `editor.formatOnSave` setting in VS Code is also properly respected.

Resolves #910
Resolves #911

### DIFF
--- a/javascript/packages/language-server/src/document_save_service.ts
+++ b/javascript/packages/language-server/src/document_save_service.ts
@@ -18,12 +18,31 @@ export class DocumentSaveService {
     this.formattingService = formattingService
   }
 
+  /**
+   * Apply only autofix edits on save.
+   * Called by willSaveWaitUntil - formatting is handled separately by editor.formatOnSave
+   */
+  async applyFixes(document: TextDocument): Promise<TextEdit[]> {
+    const settings = await this.settings.getDocumentSettings(document.uri)
+    const fixOnSave = settings?.linter?.fixOnSave !== false
+
+    this.connection.console.log(`[DocumentSave] applyFixes fixOnSave=${fixOnSave}`)
+
+    if (!fixOnSave) return []
+
+    return this.autofixService.autofix(document)
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


danbee:
@marcoroth Does this fix allow me to enable the formatter but *disable* format on save? I want to be able to use the LSP formatter, but I need to be explicit about it and specifically don't want to format on save.

marcoroth:
@danbee yes! This is how it worked before v0.8.0 and this pull request restores that behavior 👍🏼 

Enabling the formatter is a project-level config (configured in `.herb.yml`). Configuring format-on-save is a personal setting and thus not part of the `.herb.yml` config and depends on the editor you use.

PR #966, opened 2025-12-10
docs: fix file name in Linter Custom Rules example code

kozy4324:
The sample code showed duplicated file names, which was a bit confusing, so I fixed them. Thank you!

image


marcoroth (approved):
Great catch, thanks @kozy4324! 🙏🏼

Issue #956, opened 2025-12-08
Parser: HTML tag names expected to be treated as case-insensitive


kudoas:
## Bug Report

HTML tag names are defined as case‐insensitive by the HTML Living Standard, but the herb parser currently treats opening and closing tags as case‐sensitive. Because of this, tags like \<div> and \</DIV> are not matched correctly.

> In the HTML syntax, tag names, even those for foreign elements, may be written with any mix of lower- and uppercase letters that, when converted to all-lowercase, matches the element’s tag name; tag names are case-insensitive.
> refs. https://html.spec.whatwg.org/multipage/syntax.html#elements-2

Looking at this, tag name comparison seems to be performed case-sensitively here:

https://github.com/marcoroth/herb/blob/9722d342be53fa3816f8a81e9f6117734552c646/src/parser_helpers.c#L21-L28

## Reproduction

You can reproduce this issue using the following playground link:

https://herb-tools.dev/playground#DwEwlgbgfMD0AiBJAalIA

## Expected Behavior

HTML tag names are case-insensitive, so `<div></DIV>` are expected to be treated as the same tag.

## Actual Behavior

```
❯ herb parse sample.erb.html
@ DocumentNode (location: (1:0)-(2:0))
└── children: (3 items)
    ├── @ HTMLOpenTagNode (location: (1:0)-(1:5))
    │   ├── errors: (1 error)
    │   │   └── @ MissingClosingTagError (location: (1:0)-(1:5))
    │   │       ├── message: "Opening tag `<div>` at (1:1) doesn't have a matching closing tag `</div>` in the same scope."
    │   │       └── opening_tag: "div" (location: (1:1)-(1:4))
    │   │
    │   ├── tag_opening: "<" (location: (1:0)-(1:1))
    │   ├── tag_name: "div" (location: (1:1)-(1:4))
    │   ├── tag_closing: ">" (location: (1:4)-(1:5))
    │   ├── children: []
    │   └── is_void: false
    │
    ├── @ HTMLCloseTagNode (location: (1:5)-(1:11))
    │   ├── errors: (1 error)
    │   │   └── @ MissingOpeningTagError (location: (1:5)-(1:11))
    │   │       ├── message: "Found closing tag `</DIV>` at (1:7) without a matching opening tag in the same scope."
    │   │       └── closing_tag: "DIV" (location: (1:7)-(1:10))
    │   │
    │   ├── tag_opening: "</" (location: (1:5)-(1:7))
[...]

PR #968, opened 2025-12-10
Parser: HTML tags treat as case-insensitive

kudoas:
This pull request fixes https://github.com/marcoroth/herb/issues/956

HTML tags are handled case-insensitively across open/close matching, foreign content detection for script/style and SVG contexts, and the tests updated to cover mixed-case tags to ensure parsing.

Let me know if you see any improvements or have any suggestions!


marcoroth (approved):
Awesome, thank you so much for this fix, @kudoas! 🙏🏼 

Also, thank you for moving the new test cases to the end of the test files, that reduces the pull request diff a lot! 🙌🏼

Issue #975, opened 2025-12-13
Linter: Improve Make html-tag-name-lowercase Quick Fix to update both tag names


kudoas:
## Why

When applying the html-tag-name-lowercase Quick Fix in VSCode, it would enhance the user experience if the entire element (both opening and closing tags) could be corrected in one go. 

## Expected Behaviour

When executing a Quick Fix for one side (either Opening or Closing), the corresponding tag name on the opposite side should also be converted to lowercase simultaneously.

Example: `<DIV></DIV>` → `<div></div>`

## Actual Behaviour

Quick Fix only corrects the selected suggestion (Opening or Closing), leaving the other side uncorrected.

Example: `<DIV></DIV>` → `<DIV></div>`

https://github.com/user-attachments/assets/b0c41675-9fe9-438e-b0e4-2446c3d3e1d3

kudoas:
I opened a pull request that attempts to fix https://github.com/marcoroth/herb/pull/976
Let me know if you see any improvements or have any suggestions!

marcoroth:
Great idea, let's move to conversation to the pull request 🙌🏼

PR #976, opened 2025-12-13
Linter: Lowercase both opening/closing tag when fixing `html-tag-name-lowercase` offense

kudoas:
This pull request fixes https://github.com/marcoroth/herb/issues/975

Updates autofix() so that after fixing the offending tag name, it walks up to the parent element and also fixes the related counterpart tag name (opening/closing) in the same element.

Let me know if you see any improvements or have any suggestions! Thank you.

### DIFF
--- a/javascript/packages/linter/test/autofix/html-tag-name-lowercase.autofix.test.ts
+++ b/javascript/packages/linter/test/autofix/html-tag-name-lowercase.autofix.test.ts
@@ -107,6 +107,17 @@ describe("html-tag-name-lowercase autofix", () => {
     expect(result.fixed).toHaveLength(0)
   })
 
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Brilliant, this is a great feature, thank you @kudoas!

Is there any chace we can add some SVG-related tests for the autofix. Especially for the few special cases in SVG tags where they have mixed-case tag names.

Otherwise, this is great! Thanks!

kudoas:
Thanks for the review! @marcoroth

I’ve added SVG-related tests that cover converting uppercase SVG tags to lowercase, while preserving mixed-case SVG elements. 1cdba355a95576fb41f065d06db1b9cc5dcea730

Please re-review when you have time.

PR #1000, opened 2025-12-21
Dev Tools: Ensure Herb overlay label is visible when attached to html element

kozy4324:
When attaching the Herb overlay to the `<html>` element, the label may be positioned outside of the viewport because the default CSS uses a negative top value.

This adds a small conditional so the label remains visible when the overlay is attached to `<html>`.

I wasn’t sure whether adding this kind of conditional styling in JavaScript is acceptable here, or if you’d prefer a CSS-based solution. Please let me know if you have a preferred direction.

### DIFF
--- a/javascript/packages/dev-tools/src/herb-overlay.ts
+++ b/javascript/packages/dev-tools/src/herb-overlay.ts
@@ -537,7 +537,9 @@ export class HerbOverlay {
 
       parent.setAttribute('data-herb-debug-attached-outline-type', type);
 
-      parent.style.position = 'relative';
+      if (window.getComputedStyle(parent).position === 'static') {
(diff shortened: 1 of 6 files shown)


javascript/packages/dev-tools/src/herb-overlay.ts
```diff
@@ -547,6 +547,9 @@ export class HerbOverlay {
     }
 
     element.style.position = 'relative';
+    if (element.tagName.toLowerCase() === 'html') {
```
marcoroth:
```suggestion
    if (element.localName === 'html') {
```

kozy4324:
Makes sense. Thanks for the suggestion!

Issue #987, opened 2025-12-17
Linter: Exit with an error code for warnings


mbaird:
I've just added the herb linter to our CI job, relying on a non-zero exit code to fail the build.

The command I'm using is `yarn herb-lint --simple`.

The project has zero offences, and it'd be really handy if we could keep any new warnings from making it into production, having them fail the build (in addition to errors).

Examples of this include Rubocop, which does this via [`--fail-level`](https://docs.rubocop.org/rubocop/usage/basic_usage.html#command-line-flags), and biome, which offers [`--error-on-warnings` ](https://biomejs.dev/blog/biome-v1/#exit-on-warnings).

marcoroth:
Thanks for reporting this @mbaird!

This a great point and I think that makes sense. Maybe we can also expose it as config option in the `.herb.yml` file, so you don't always have to specify level when running the linter CLI.

marcoroth:
@mbaird this is now released with https://github.com/marcoroth/herb/releases/tag/v0.8.7 🙌🏼

mbaird:
Amazing, works great. Thanks @marcoroth and @ydah!

marcoroth:
Awesome, thanks for the quick confirmation! 🙏🏼

PR #1007, opened 2025-12-25
Linter: Add `--fail-level` option to control exit code

ydah:
Fixes: https://github.com/marcoroth/herb/issues/987

### DIFF
--- a/javascript/packages/linter/test/cli.test.ts
+++ b/javascript/packages/linter/test/cli.test.ts
@@ -369,7 +369,7 @@ describe("CLI Output Formatting", () => {
                 severity: warning
         `)
 
-        const { output, exitCode } = runLinter("clean-file.html.erb", "--fail-level", "warning")
+        const { exitCode } = runLinter("clean-file.html.erb", "--fail-level", "warning")
 
         expect(exitCode).toBe(0)
       } finally {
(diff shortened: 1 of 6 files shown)


javascript/packages/linter/test/cli.test.ts
```diff
@@ -336,6 +336,110 @@ describe("CLI Output Formatting", () => {
...
+      } finally {
+        try { unlinkSync(configPath) } catch {}
+      }
+    })
+
+    test("exits with error code when failOnWarning is set in config", () => {
+      try {
+        writeFileSync(configPath, dedent`
+          linter:
+            failOnWarning: true
```
marcoroth: I'm wondering if we should use kabab-case here for the key: `fail-on-warning`? 

Or maybe it would make more sense to use a more flexible key, like: `fail-level: warning`. That way we could even say `fail-level: info`

ydah: I tried using flexible key. What do you think?

marcoroth: Prefect, thank you! I simplified it a bit, added docs and added some more test cases!

ydah: Thank you!

marcoroth:
This is now released with https://github.com/marcoroth/herb/releases/tag/v0.8.7, thank you!

Issue #978, opened 2025-12-15
Formatter: BUG: Duplicates LOC


gobijan:
**Input:**
```erb

<p class="text-reversed">
  <strong><%= t("admin.tickets.form.savings") %></strong><%= ticket.ticketable.savings_percentage %>%
  <br>
  <strong><%= t("admin.tickets.form.price_per_ticket") %></strong><%= format_price(ticket.ticketable.price_per_ticket) %>
</p>

```

**Output:**
```erb

<p class="text-reversed">
  <strong><%= t("admin.tickets.form.savings") %></strong><%= ticket.ticketable.savings_percentage %>%
  <br>
  <strong><%= t("admin.tickets.form.savings") %></strong>
  <%= ticket.ticketable.savings_percentage %>
  %
  <br>
  <strong><%= t("admin.tickets.form.price_per_ticket") %></strong>
  <%= format_price(ticket.ticketable.price_per_ticket) %>
</p>

```

**Expected:**
```erb
No duplication per save.
```

**Playground:**

**Additional context:**

PR #1008, opened 2025-12-25
Formatter: Fix infinite loop when rendering multiple groups of adjacent inline elements

ydah:
Fixes: https://github.com/marcoroth/herb/issues/978


marcoroth:
Thank you @ydah! 🙏🏼

Issue #995, opened 2025-12-19
Parser: ERB Tag with `yield` in block is being analyzed as ERBBlockNode


marcoroth:
```erb
<% content = capture { yield } if block_given? %>
```

```js
@ DocumentNode (location: (1:0)-(1:49))
├── errors: []
└── children: (1 item)
    └── @ ERBBlockNode (location: (1:0)-(1:49))
        ├── errors: (1 item)
        │   └── @ MissingERBEndTagError (location: (1:0)-(1:49))
        │       ├── message: "ERB block started here but never closed with an end tag. The end tag may be in a different scope."
        │       └── keyword: "ERB block"
        │       
        │   
        ├── tag_opening: "<%" (location: (1:0)-(1:2))
        ├── content: " content = capture { yield } if block_given? " (location: (1:2)-(1:47))
        ├── tag_closing: "%>" (location: (1:47)-(1:49))
        ├── body: []
        └── end_node: ∅
```

PR #1009, opened 2025-12-25
Parser: Fix block detection to exclude closed brace blocks without yield

ydah:
Fixes: https://github.com/marcoroth/herb/issues/995


marcoroth (approved):
Awesome, thank you so much @ydah! 🙏🏼

PR #1013, opened 2025-12-26
Dev Tools: Ensure Herb overlay label is visible in overflow-clipped elements

kozy4324:
Previously handled only the html element, but labels can be clipped whenever overflow is not visible.

Relates to #1000


marcoroth:
This is now released with https://github.com/marcoroth/herb/releases/tag/v0.8.7, thank you!

Issue #186, opened 2025-06-27
Linter Rule: Disallow headings inside buttons


NotGrm:
##### Description

Checks <button> elements to see if they contain heading (<h1> - <h6>) elements, and gives an error message if they are found.

##### Rationale

Assistive technology allows users to browse a page by heading elements (<h1> - <h6>). However, if those heading elements are nested inside of button elements, they will automatically be marked as presentational by browsers. Any HTML element where ["children presentational" is true](https://w3c.github.io/aria/#button) should be coerced by the browser to be presentational, and therefore not included in the accessibility tree.

##### ✅ Good

```html+erb
<button><span>Button Text</span></button>
```

##### 🚫 Bad

```html+erb
<button><h1>Some Text</h1></button>
```

#### References

Inspired by [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-heading-inside-button.md)

Issue #267, opened 2025-07-15
Enable and rework `html-no-block-inside-inline` rule


marcoroth:

ksweetie:
I could definitely see this being a valuable rule if it works well! Just commenting to add another bug. The documentation states:

> Placing block-level elements (like `<div>`, `<p>`, `<section>`) inside inline elements (like `<span>`, `<a>`, `<strong>`) violates HTML content model rules

However, `<a>` tags _can_ have block-level elements in them. That wasn't the case in HTML 4.01, but is allowed since HTML 5. Here's [the W3c spec](https://www.w3.org/TR/2011/WD-html5-20110405/text-level-semantics.html#the-a-element), which says:

> The [a](https://www.w3.org/TR/2011/WD-html5-20110405/text-level-semantics.html#the-a-element) element may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g. buttons or other links). This example shows how this can be used to make an entire advertising block into a link:
> 
>     <aside class="advertising">
>      <h1>Advertising</h1>
>      <a href="http://ad.example.com/?adid=1929&amp;pubid=1422">
>       <section>
>        <h1>Mellblomatic 9000!</h1>
>        <p>Turn all your widgets into mellbloms!</p>
>        <p>Only $9.99 plus shipping and handling.</p>
>       </section>
>      </a>
>      <a href="http://ad.example.com/?adid=375&amp;pubid=1422">
>       <section>
>        <h1>The Mellblom Browser</h1>
>        <p>Web browsing at the speed of light.</p>
>        <p>No other browser goes faster!</p>
>       </section>
>      </a>
>     </aside>

Thanks for your hard work, I'm really excited about the Herb tools!

marcoroth:
That's a very good point, thanks for the context @ksweetie! 🙏🏼 

I guess what you are describing is similar to the rule proposed in https://github.com/marcoroth/herb/issues/281

ksweetie:
I haven't thought too deeply on it, but it might make sense to have two rules.

1. This one, which disallows blocks inside of inline elements, to prevent things like `<span><div></div></span>`.
2. #281, which would be specific to not nesting interactive elements.

I could see combining them into a single "valid nesting" rule, but they feel different enough to separate.

kozy4324:
Hi, thanks for the discussion on this rule!

[...]

Issue #291, opened 2025-07-22
Possible diagnostic issue: `html-no-block-inside-inline`


andycroll:
**Line:** 40
  **Diagnostic:** Unknown element `<svg>` cannot be placed inside inline element `<button>`.
  **Severity:** Error

  **File Content:**
  Please paste your HTML ERB file content below to help us reproduce the issue:

  ```html erb
  <!-- Paste your HTML ERB file content here -->
  ```

  **Problematic Line:**
  ```erb
            <svg class="h-6 w-6 text-white" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" aria-hidden="true">
  ```

  **Environment:**
- VS Code Extension Version: 0.4.0
- VS Code Version: 1.102.1
- Platform: darwin
- Node Version: v22.15.1

  **Herb Settings:**
```json
{
"linter.enabled": true,
"trace.server": "verbose"
}
```

  **Additional Context:**

Running it on a fairly simple Rails project.

Fairly common to place an inline SVG inside a `button` or `a` tag. Is this intended behaviour for this rule?

andycroll:
Noticed you've disabled in 0.4.1. so that solved "the problem" for me, at least until it gets turned back on.

Side note, this is lovely software.

marcoroth:
Hey @andycroll, thanks for reporting!

Yeah, I just disabled this rule in https://github.com/marcoroth/herb/pull/266 since it was causing more trouble than it was worth it for right now. 

I also opened https://github.com/marcoroth/herb/issues/267 to rework and enable this rule again.

Issue #272, opened 2025-07-17
Linter: Diagnostic issue: `html-no-block-inside-inline` option cannot be placed inside element select


geoffgom:
**Line:** 106
  **Diagnostic:** Unknown element `<option>` cannot be placed inside inline element `<select>`.
  **Severity:** Error

  **File Content:**
  Please paste your HTML ERB file content below to help us reproduce the issue:

  ```html erb
  <!-- Paste your HTML ERB file content here -->
  ```

  **Problematic Line:**
  ```erb
             <select
                id="my_property_combobox_<%=pc.id%>"
                name="property_combobox-<%=pc.id%>[]"
                class="block w-full px-3 py-1 mt-2 border border-gray-400 rounded-md shadow outline-none dark:bg-gray-700 dark:border-gray-600 disabled:bg-gray-100"
              >
                <% pc.property_combobox_options.each do |pco| %>
                  <% selected = pc.property_list_values.where(user_property_group:).first&.values&.include?(pco.id) %>
                  <option value="<%= pco.id %>">
                    <%= pco.label %>
                  </option>
                <% end %>
              </select>
  ```

  **Environment:**
- VS Code Extension Version: 0.4.0
- VS Code Version: 1.102.1
- Platform: darwin
- Node Version: v22.15.1

  **Herb Settings:**
```json
{
"linter.enabled": true,
"trace.server": "verbose"
}
```

  **Additional Context:**

DanishIbrahimKhan:
Hi @geoffgom 
Try placing a comment to "break up" the parser confusion
<select ...>
  
  <% pc.property_combobox_options.each do |pco| %>
    <option value="<%= pco.id %>">
      <%= pco.label %>
    </option>
  <% end %>
  
</select>

marcoroth:
Hey @geoffgom, thanks for reporting!

We disabled the `html-no-block-inside-inline` rule for now in https://github.com/marcoroth/herb/pull/266 and I'm hoping to release v0.4.1 with it today.

We have https://github.com/marcoroth/herb/issues/267 to bring the rule back later.

PR #1030, opened 2025-12-30
Linter: Add `html-content-model-conformance` rule

kozy4324:
This PR implements a content model–based validation approach as discussed in #267, focusing on a small and practical subset of checks derived from the HTML specification.

### What this PR does

This implementation checks whether child elements are allowed under a given parent based on its content model, without aiming for full structural validation.

It currently supports the following patterns:

1. Flow disallowed cases (simplest case)
    - Detects cases where flow content appears in places where flow content is not permitted.
1. Transparent content model (e.g. `<a>`)
    - Elements with a transparent content model delegate validation to their parent’s content model.
1. Specific tag patterns (e.g. `<ul>` → `<li>`)
    - Handles elements that only allow specific child tags.
1. Context-dependent patterns
    - Handles elements whose allowed content depends on context (e.g. `<div>`, `<span>`).
1. Custom elements and unknown elements
    - Since custom and unknown elements do not have an explicit content model, they are treated as allowing any content.

### What this PR intentionally does not do

The following are intentionally out of scope for this PR:

1. Child element ordering
    - The order of children defined by a content model is not validated.
1. Descendant-level constraints
    - Rules that forbid specific elements anywhere in the descendant tree are not checked.
1. Validation under `<head>`
    - These cases are already covered by the existing `html-head-only-elements` and `html-body-only-elements` rules.

I’d be happy to adjust this if you have any feedback or suggestions. Thank you.


kozy4324:
This PR also seems to fix #186, #248, #255, #260, #272, and #291.

kozy4324:
The commit ca3758b2afadd06b78bc82c29c1d9b3d382d7cf9 adds support for ERBBlockNode and ERBContentNode with the `tag` helper.

Extracting helper methods via regex is getting tricky.  
Since a Prism-based approach isn’t available yet, should we wait for that, or keep extending the regex support for now?

marcoroth:
Great, thank you so much for exploring this @kozy4324! 

> Since a Prism-based approach isn’t available yet, should we wait for that, or keep extending the regex support for now?

We can leave out the `tag.div` and `content_tag` and similar helpers in ERB for now. A lot of rules need to be updated once we have the tag detection working.  

If you want, we can keep the tests and use `test.fails("...", () => {})` so we know which ones to update later on.

kozy4324:
Got it. I’ll keep the scope limited to the `tag` helper for now. Thank you!

PR #1031, opened 2026-01-01
Fix test failure caused by year change

kozy4324:
Happy New Year! This PR fixes a test that didn’t survive the year change.


marcoroth (approved):
Happy New Year, and thank you for catching this! 😅

PR #1035, opened 2026-01-03
Add `irb` to `Gemfile`

### LINKED ISSUE #1034: `bundle console` fails with “irb is not available” on Ruby 4.0
When running `bundle console` as described in the documentation, I get the following error on Ruby 4.0.0:

https://github.com/marcoroth/herb/blob/0514903287a699a2748ca6433e4ea6e83f295e45/CONTRIBUTING.md#building-the-ruby-extension

```
$ bundle console
/Users/kozy4324/.rbenv/versions/4.0.0/lib/ruby/gems/4.0.0/gems/bundler-2.7.2/lib/bundler/cli/console.rb:20: warning: irb used to be loaded from the standard library, but is not part of the default gems since Ruby 4.0.0.
You can add irb to your Gemfile or gemspec to fix this error.
irb is not available
```

It looks like adding `irb` to the Gemfile resolves the issue on my side.

However, when I modify the Gemfile and run `bundle install`, additional gems are added to `Gemfile.lock`, so I wanted to report this as an issue rather than opening a PR.

kozy4324:
fixes #1034 .

Ruby 4.0 no longer includes `irb` by default, which causes `bundle console` to fail.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,3 +27,5 @@ gem "steep", "~> 1.10"
 # You can add tsort to your Gemfile or gemspec to silence this warning.
 # 🎉 Generated 0 RBS files under sig/
 gem "tsort", "~> 0.2.0"
+
+gem "irb", "~> 1.16"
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Thank you, @kozy4324! 🙏🏼

Issue #1048, opened 2026-01-11
Linter Rule: Enforce correct `locals:` strict locals comment syntax


marcoroth:
**Rule:** `erb-strict-locals-comment-syntax`

## Description

Warn when a comment intended to declare strict locals is misspelled or uses an invalid syntax, such as `<%# locals() %>`.  

## Rationale

Strict locals comments are used by Rails itself and for tooling/linters to statically understand which locals are expected in a template. If the comment is misspelled or malformed, it silently does nothing or leads to unexpected errors when rendering, since certain locals might not be defined.

By enforcing a strict syntax, this rule prevents broken annotations and ensures that declared locals are actually respected.

### ✅ Good

```erb
<%# locals: (user:, admin: false) %>

<p><%= user.name %></p>
```

```erb
<%# locals: (title:) %>

<h1><%= title %></h1>
```

### 🚫 Bad

```erb
<%# locals() %>
```

```erb
<%# local: (user:) %>
```

```erb
<%# locals (user:) %>
```

```erb
<%# locals: user %>
```

```erb
<%# locals: %>
```

```erb
<%# locals: (user: %>
```

```erb
<% # locals: (user:) %>
```

## References

- [Action View - Strict Locals](https://guides.rubyonrails.org/action_view_overview.html#strict-locals)

PR #1049, opened 2026-01-12
Linter Rule: Add `erb-strict-locals-comment-syntax` rule to enforce Rails strict locals comment syntax

ydah:
Fixes: https://github.com/marcoroth/herb/issues/1048


ydah:
Thank you!

PR #1066, opened 2026-01-15
Linter: Accessibility rules should be warnings, not errors

joelhawksley:
I would like to propose that we downgrade the default severity of a11y-related rules to `warning` from `error`.

While I admire the courage it takes to establish such a strict set of default rules for the linter, I believe we should reserve errors for cases where the HTML is truly invalid and use lower severities for other cases.

The changes here are just an example. If y'all agree with this proposal, I think we should discuss what additional rules need downgraded.

### DIFF
--- a/javascript/packages/linter/test/cli.test.ts
+++ b/javascript/packages/linter/test/cli.test.ts
@@ -235,8 +235,8 @@ describe("CLI Output Formatting", () => {
   test("--no-github disables GitHub Actions annotations", () => {
     const { output, exitCode } = runLinter("test-file-with-errors.html.erb", "--no-github", { GITHUB_ACTIONS: "true" })
 
-    expect(output).not.toMatch(/^::error/)
-    expect(output).toMatch(/error.*Missing required.*alt.*attribute/)
+    expect(output).not.toMatch(/^::warning/)
+    expect(output).toMatch(/warning.*Missing required.*alt.*attribute/)
     expect(exitCode) .toBe(1)
   })
 


marcoroth:
In general, I agree with changing the severity of rules 👍🏼 

For context: initially, I wanted to have all rules enabled by default, so that these rules get seen, used, and "validated" by as many people as possible. That meant that we also just added all rules to the `error` severity. Consequently, all rules are probably still "too aggressive" as we haven't balanced the levels yet.

RE: the a11y rules: I wanted to have them enabled by default so people get exposed to these concerns and maybe can even pick up and learn something from the linter. Having the linter give you a little nudge as you write the HTML is good thing. Especially, since a11y is not the first thing most people think about when writing code.

If we would make these rules opt-in way less people will actually use and see them. As accessibility gots more important I think it's fair to say that people should willingly disable the rules if they feel like they get "too annoying". But then it's a conscious decision on their part.

Therefore, I’d rather have these a11y rules report at the `info` level by default (so they don’t fail the build), than being disabled by default.

joelhawksley:
> For context: initially, I wanted to have all rules enabled by default, so that these rules get seen, used, and "validated" by as many people as possible.

I still think this is correct, in that we should run and report all rules by default.

> I’d rather have these a11y rules report at the info level by default (so they don’t fail the build), than being disabled by default.

Interesting, I'd have expected `warning` to behave this way.

marcoroth:
Alright, I updated all a11y-related rules to `warning`. Looking at our config, we actually have the `linter.failLevel` set to `error` by default, meaning that warnings won't let CI fail, unless you set the `failLevel` to `warning`.

I wonder we should lower the `failLevel` to `warning` so that warnings also let the build fail and that people have to opt-out by setting it to `error` explicitly. Either way, this shouldn't be part of this PR, but I'd love to get your opinion on that.

Also, happy to review and downgrade any of the rules to warning as well if you have a strong opinion 🙌🏼

joelhawksley:
> I wonder we should lower the `failLevel` to `warning` so that warnings also let the build fail and that people have to opt-out by setting it to `error` explicitly. Either way, this shouldn't be part of this PR, but I'd love to get your opinion on that.
> 
> Also, happy to review and downgrade any of the rules to warning as well if you have a strong opinion 🙌🏼

My gut says that we should only raise errors for issues that will block Herb from working at all, at least when it comes to folks adopting ReActionView in an existing codebase, given the project's intention of being ActionView-compatible and the (continued?) intention to eventually replace ActionView. 

That also tracks with how I generally see warnings used elsewhere in the Ruby ecosystem: they are for situations where the code in question will still run, but is in some way incorrect.

PR #1067, opened 2026-01-16
Upgrade Prism to v1.8.0

marcoroth:
Prism 1.8.0 (https://github.com/ruby/prism/pull/3827) changed how the `closing_loc` is populated for unclosed blocks. Previously it was `NULL` for unclosed blocks, but now it's always set (pointing to body content). This broke our detection logic that relied on checking if `closing_loc` was empty.

https://github.com/ruby/prism/releases/tag/v1.8.0


marcoroth:
Quite a few interesting failures 🤔 

Gonna let this sit for a second and come back later

PR #1112, opened 2026-01-31
Parser: Add error for case/when in same ERB tag

tvararu:
This adds test cases and a custom error to handle code like:

```erb
<% case variable when "a" %>
  A
<% when "b" %>
  B
<% else %>
  C
<% end %>

<% case value in 1 %>
  One
<% in 2 %>
  Two
<% end %>
```

While this is valid ERB, it poses issues for the parser, as it would mean we'd need to turn a single statement into multiple nodes.

Those same blocks can be rewritten as:

```erb
<% case variable %>
<% when "a" %>
  A
<% when "b" %>
  B
<% else %>
  C
<% end %>

<% case value %>
<% in 1 %>
  One
<% in 2 %>
  Two
<% end %>
```

Which is unambiguous to parse, and arguably more legible.

This adds a new ERBCaseWithConditionsInTagError to detect this pattern and direct users to split into separate tags.

### DIFF
--- a/sig/herb/errors.rbs
+++ b/sig/herb/errors.rbs
@@ -253,5 +253,15 @@ module Herb
       # : (?indent: Integer, ?depth: Integer, ?depth_limit: Integer) -> String
       def tree_inspect: (?indent: Integer, ?depth: Integer, ?depth_limit: Integer) -> String
     end
+
+    class ERBCaseWithConditionsError < Error
+      include Colors
+
+      # : () -> String
+      def inspect: () -> String
+
+      # : (?indent: Integer, ?depth: Integer, ?depth_limit: Integer) -> String
+      def tree_inspect: (?indent: Integer, ?depth: Integer, ?depth_limit: Integer) -> String
+    end
   end
 end


tvararu:
@marcoroth I tried approaching this in a different way, like:

```c
if ((analyzed->case_node_count > 0 && analyzed->when_node_count > 0) ||
    (analyzed->case_match_node_count > 0 && analyzed->in_node_count > 0)) {
  append_erb_case_with_conditions_error(
    erb_content_node->base.location.start,
    erb_content_node->base.location.end,
    erb_content_node->base.errors
  );
}
```

But it doesn't look like `case x when "a"` increments the `when_node_count`, it stays `0`...

We have to add something to `src/analyzed_ruby.c` as the `analyzed` struct doesn't store enough information for us to infer whether a case node has inline conditionals; `case_node->conditions.size` is what we need but we need to extract it somewhere...

We could have `analyzed->inline_conditionals_count`, that would be an int instead of a boolean so it's more consistent with the rest of the struct. I pushed up a new commit with this but it's very similar to the boolean one, I've also changed the error name and the message

If you like this approach better I can rebase and squash the commits 👍

tvararu:
Thanks @marcoroth! Yeah this is much cleaner than what I could come up with 😄

marcoroth:
Released as part of [v0.8.10](https://github.com/marcoroth/herb/releases/tag/v0.8.10)! 🚀

PR #1118, opened 2026-02-03
Mark certain targets as `.PHONY` in Makefile

citizen428:
To quote the [fine manual](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html):

> A phony target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request. There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.

Note: While you can also declare all `.PHONY` targets in a list, I've found that individual declarations work better for larger Makefiles like this one, since they require less vertical scanning.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,7 @@ ifeq ($(os),Darwin)
   clang_tidy = $(llvm_path)/bin/clang-tidy
 endif
 
+.PHONY: all
 all: templates prism $(exec) $(lib_name) $(static_lib_name) test wasm clangd_config
 
 $(exec): $(objects)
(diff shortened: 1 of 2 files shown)


citizen428:
I should have added that this is relevant for at least one existing target:

Without `.PHONY: wasm`

```
❯ make wasm
make: 'wasm' is up to date.
```

With `.PHONY: wasm`:

```
❯ make wasm
cd wasm && make
make[1]: Entering directory '/Users/me/src/herb/wasm'
...
```

citizen428:
@marcoroth Finally 😅

**The issue (this was a fun adventure)**: 

1. The `libherb` project runs `make` with no target from the root. This will use the first target in the file, which is `all`.
2. `all` includes `wasm`:
    ```make
    all: templates prism $(exec) $(lib_name) $(static_lib_name) test wasm clangd_config
    ```
3. Prior to this PR, `make wasm` was essentially a no-op, because the target wasn't declared as `.PHONY` and a directory named `wasm` existed, so `make` considered the target up to date and skipped it.
4. Now that the target is actually declared as `.PHONY`, it is executed.

**The fix:** 
`libherb:build` shouldn't run `make all` since it only needs the library. I updated `project.json` to only build the necessary targets:
```make
make templates prism build/libherb.a
```

marcoroth (approved):
Awesome, thank you so much for working through this, @citizen428! 🙏🏼 

Also, TIL I finally learned what the `PHONY` targets are for. So far I was never curious enough to look it up, but this totally makes sense! I'm surprised this didn't cause any issues earlier 😅

PR #1124, opened 2026-02-05
Add `flake.nix`

citizen428:
_Note: This needs #1123 merged first._

**Rationale:**
Since I generally work on multiple projects at the same time, I manage my development dependencies via Nix flakes in combination with [nix-direnv](https://github.com/nix-community/nix-direnv). The Herb project specifically has quite a few dependencies that I'd rather not install globally. 

I was a bit hesitant about adding that at first, but since this project already contains an `Aptfile` and a `Brewfile` I don't see a problem with adding one more file, given that it significantly lowers the barrier of entry for Nix users.


CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ If you encounter any issues when following along with this file please don't hes
 - [**Doxygen**](https://www.doxygen.nl): For building the C-Reference documentation pages.
 
 
-##### For Linux
+#### For Linux
```
citizen428, line 22: While not directly related I noticed that this and the following headline included one `#` too many.

Makefile
```diff
@@ -67,13 +67,16 @@ ifeq ($(os),Linux)
...
 
 ifeq ($(os),Darwin)
-  brew_prefix := $(shell brew --prefix check)
-  test_cflags = $(test_flags) -I$(brew_prefix)/include
-  test_ldflags = -L$(brew_prefix)/lib -lcheck -lm $(prism_ldflags)
-  llvm_path = $(shell brew --prefix llvm@21)
-  cc = $(llvm_path)/bin/clang
-  clang_format = $(llvm_path)/bin/clang-format
-  clang_tidy = $(llvm_path)/bin/clang-tidy
+  llvm_version := 21
```
citizen428, line 70: This is from #1123, so just a temporary diff.

citizen428:
@marcoroth I understand if this is maybe a tad to specific to merge (though maybe not more specific than `Aptfile` or `Brewfile`). It could also work to enable the Wiki for this project and have me document some things there. My main goal is to make contributing to Herb easier for new people, I don't really care how we achieve that.

marcoroth (commented):
Yeah, I'm not 100% sure that the adding Nix here is improving the onboarding for people, as it is super specific and something I cannot confidently maintain.

I personally use the `Brewfile` for my work on macOS and the `Aptfile` is being used in CI to install the packages, so at least those are somewhat ensured to work. 

I think I'd rather keep the Wiki disabled and update the `CONTRIBUTING.md`, and if really needed, we can have a dedicated page on the [`herb-tools.dev`](https://herb-tools.dev) docs page.

PR #1127, opened 2026-02-06
C: Add test for `herb_read_file`

citizen428:
Document that the function exits when the filename is invalid.

**Rationale:**
I was originally a bit concerned when reading `main.c` that there was no `NULL` check around this line:

```c
char* source = herb_read_file(argv[2]);
```

However, `argv[2]` should never be empty due to this check:

```c
if (argc < 3) {
  printf("Please specify input file.\n");
  return 1;
}
```

This leaves the potential problem of non-existent files, which is handled directly inside `herb_read_file`. This behavior surprised me a little bit, so I thought it might be a good idea to document it with a test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ $(static_lib_name): $(objects)
 src/%.o: src/%.c templates
 	$(cc) -c $(flags) -fPIC $< -o $@
 
-test/%.o: test/%.c templates
+test/%.o: test/%.c templates prism
 	$(cc) -c $(test_cflags) $(test_flags) $(prism_flags) $< -o $@
 
 .PHONY: test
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -95,7 +95,7 @@ $(static_lib_name): $(objects)
 src/%.o: src/%.c templates
 	$(cc) -c $(flags) -fPIC $< -o $@
 
-test/%.o: test/%.c templates
+test/%.o: test/%.c templates prism
```
citizen428, line 98: Note: I needed to add `prism` as a dependency of the `test` target, otherwise `make test` could not complete after a `make clean` due to a missing Prism header.

PR #1128, opened 2026-02-06
C: Add `free` calls to `prism` command

citizen428:
**Rationale:**
While this is not really a memory leak (the OS will reclaim the memory since we `exit` right after), the missing `free`s are not that big of a deal. However, all other subcommands in `main.c` explicitly free the allocated memory, so I thought it'd be good to add some consistency here.


marcoroth (approved):
I'm sure there are more like those somewhere in the codebase, thank for catching this one! 🙏🏼

PR #1129, opened 2026-02-06
C: Remove old `visit` subcommand

citizen428:
**Rationale:**
While reading `main.c` I noticed a subcommand called `visit`, which was not documented in `herb`'s help:

```c
    printf("./herb lex [file]      -  Lex a file\n");
    printf("./herb parse [file]    -  Parse a file\n");
    printf("./herb ruby [file]     -  Extract Ruby from a file\n");
    printf("./herb html [file]     -  Extract HTML from a file\n");
    printf("./herb prism [file]    -  Extract Ruby from a file and parse the Ruby source with Prism\n");
```

@marcoroth clarified that this is a leftover from the early days of Herb, so this PR removes the subcommand.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -61,24 +61,6 @@ int main(const int argc, char* argv[]) {
   struct timespec start, end;
   clock_gettime(CLOCK_MONOTONIC, &start);
 
-  if (strcmp(argv[1], "visit") == 0) {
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thanks! 🙏🏼

PR #1130, opened 2026-02-06
C: Introduce `string_equals` utility function

citizen428:
_Note: this was built on top of #1129, which should be merged first. I didn't want to update the already removed `visit` subcommand._

This PR introduces a string utility function called `string_equals` and updates usages of `strcmp` to use that instead.

**Rationale:**
While perusing `main.c`, I found the repetitive `strcmp` calls a bit of a nuisance. I originally considered adding the following macro:

```c
#define STRING_EQ(a, b) (strcmp((a), (b)) == 0)
```

However, since we're targeting C99, we can also use a `static inline` function instead.

**File size considerations:**
I built the binary on my machine with both the `strcmp` and `string_equals` versions. As expected, with our production flags, this resulted in no increased binary size as `clang` successfully inlines all the calls. For debug builds, there's an increase of 448 bytes, which I think is a reasonable tradeoff for the increased readability.

| Build | `strcmp` | `string_eq` | Diff |
|-------|---------------------|-------------|------|
| Debug | 841,208 bytes | 841,656 bytes | +448 bytes (+0.05%) |
| Production | 649,960 bytes | 649,960 bytes | 0 bytes |

**Open question:**
Are we actually using the production flags to build releases? The assigned variable is never used in the `Makefile`.

### DIFF
--- a/src/include/util/string.h
+++ b/src/include/util/string.h
@@ -1,5 +1,5 @@
-#ifndef HERB_STR_UTILS_H
-#define HERB_STR_UTILS_H
+#ifndef HERB_UTILS_STRING_H
+#define HERB_UTILS_STRING_H
 
 #include <stdbool.h>
 #include <string.h>


src/include/util/str_utils.h
```diff
@@ -0,0 +1,11 @@
+#ifndef HERB_STR_UTILS_H
+#define HERB_STR_UTILS_H
+
+#include <stdbool.h>
+#include <string.h>
+
+static inline bool string_eq(const char* a, const char* b) {
```
citizen428: Note that this is a header-only definition, which is common for this type of `static inline` function to ensure every compilation unit has its copy of the function.

marcoroth: Would you mind renaming this to `string_equals`? 
```suggestion
static inline bool string_equals(const char* a, const char* b) {
```

src/include/util/string.h
marcoroth, line 1: Can we also rename this file to either:

`src/include/util/string_utils.h` or `src/include/util/string.h`

src/include/util/str_utils.h
citizen428: Sure, I'll do that and the other rename later. Old C habits die hard. 😅

src/include/util/string.h
citizen428: Renamed to `src/include/util/string.h`.

citizen428:
@marcoroth Renamed everything as requested.

src/include/util/string.h
```diff
@@ -0,0 +1,11 @@
+#ifndef HERB_STR_UTILS_H
+#define HERB_STR_UTILS_H
```
marcoroth:
```suggestion
#ifndef HERB_UTILS_STRING_H
#define HERB_UTILS_STRING_H
```

PR #1131, opened 2026-02-07
C: Add default language case in `herb_extract`

citizen428:
This is admittedly a rather defensive code change, but my previous work on parsers has taught me that exhaustive `switch` cases are generally a GoodIdea™. 

**Rationale:** 
So far, the `switch` in `herb_extract` only considers the two obvious cases, `HERB_EXTRACT_LANGUAGE_RUBY` and `HERB_EXTRACT_LANGUAGE_HTML`. In case the language type is invalid, we'd just silently return an empty string. I'd posit that failing an assertion and being made aware of this in development is a better approach.

### DIFF
--- a/src/extract.c
+++ b/src/extract.c
@@ -4,6 +4,7 @@
 #include "include/util/hb_array.h"
 #include "include/util/hb_buffer.h"
 
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -129,6 +130,7 @@ char* herb_extract(const char* source, const herb_extract_language_T language) {
   switch (language) {
     case HERB_EXTRACT_LANGUAGE_RUBY: herb_extract_ruby_to_buffer(source, &output); break;
     case HERB_EXTRACT_LANGUAGE_HTML: herb_extract_html_to_buffer(source, &output); break;
+    default: assert(0 && "invalid extract language");
   }
 
   return output.value;


src/extract.c
```diff
@@ -129,6 +130,7 @@ char* herb_extract(const char* source, const herb_extract_language_T language) {
   switch (language) {
     case HERB_EXTRACT_LANGUAGE_RUBY: herb_extract_ruby_to_buffer(source, &output); break;
     case HERB_EXTRACT_LANGUAGE_HTML: herb_extract_html_to_buffer(source, &output); break;
+    default: assert(0 && "invalid extract language");
```
citizen428, line 133: @marcoroth Since you mentioned the other day that you're rather new to C, this idiom is a workaround to get a message printed when an assert fails. Since && short-circuits, the value of the assert will always be 0. But the message will be printed to the terminal.

Compare the various approaches:

```
assert(0); // Assertion failed: (0) 
assert(0 /* invalid extract usage*/); // nicer for readers, but same message since preprocessor strips comments
assert(0 && "invalid extract language"); // Assertion failed: (0 && "invalid extract language")
```

It's ugly but works and is a relatively common workaround.

marcoroth (approved):
Thanks @citizen428, I think this makes sense.

I learned way too late about `assert` and haven't really been using it in much of the codebase. There are probably quite a few other places where an `assert` could make sense to convey more meaning about an error-state or unexpected data.

PR #1133, opened 2026-02-07
Dev Tools: Show colored outlines in Herb Debug Menu

marcoroth:
CleanShot 2026-02-07 at 17 11 50@2x

Thanks for the idea @ChaelCodes! 🙏🏼


ChaelCodes:
This looks great!!! ✨

marcoroth:
Might want to change the blue we use for the views, as it's quite similar to the purple we use for the ERB outlines 🙈

ChaelCodes:
> Might want to change the blue we use for the views, as it's quite similar to the purple we use for the ERB outlines 🙈

1. I'm just excited to have inspired a new idea in Herb intentionally instead of inspiring one by forgetting basic syntax. :joy:

2. Since I did inspire it... and you are looking for colors... [my editor theme](https://horizontheme.netlify.app) /jk

Issue #476, opened 2025-09-05
Formatter: Does not handle heredocs correctly


jeremyevans:
**Input:**
```erb
<%= <<EXAMPLE
example
EXAMPLE
%>
```

**Output:**
```erb
<%= <<EXAMPLE
example
EXAMPLE %>
```

**Expected:**
```erb
<%= <<EXAMPLE
example
EXAMPLE
%>
```

**Playground:**

https://herb-tools.dev/playground?tab=format#DwUgvABMwKIBoEECyAFAMjAUAUwB4EMBbABwBttN5l0sQA+IA

**Additional context:**

Not a theoretical issue, affects the first application I tried the formatter on (https://github.com/ubicloud/ubicloud)

tcannonfodder:
Likewise, it would be great if Herb was able to detect the `HERB` heredoc to address inline ERB in ViewComponents: https://viewcomponent.org/guide/getting-started.html#implementation

pinzonjulian:
Hey!

Just wanted to report that we're trying to adopt ReActionView at Buildkite and are running into this same issue.

citizen428:
This issue comes from the following function in `javascript/packages/formatter/src/format-printer.ts`:

```ts
  /**
   * Format ERB content with proper spacing around the inner content.
   * Returns empty string if content is empty, otherwise wraps content with single spaces.
   */
  private formatERBContent(content: string): string {
    return content.trim() ? ` ${content.trim()} ` : ""
  }
```

Trimming the tag content on both sides removes the newline and moves the closing `%>` up to the previous line. I have a fix for this, which is admittedly a bit hacky, but it fixes the issue and doesn't break any of the existing tests. I'll open a PR and see if @marcoroth thinks it's good enough.

marcoroth:
I think #1142 is good enough for now, as it improves the current situation. But ideally, once we have #1121, we can rework this and make it a bit more accurate and correct 🙌🏼

marcoroth:
The fix in #1142 was released as part of [v0.8.10](https://github.com/marcoroth/herb/releases/tag/v0.8.10)! 🚀

PR #1142, opened 2026-02-10
Formatter: Fix heredoc formatting

citizen428:
Fixes https://github.com/marcoroth/herb/issues/476

This PR addresses the heredoc formatting problem described in the issue.

### DIFF
--- a/javascript/packages/formatter/src/format-printer.ts
+++ b/javascript/packages/formatter/src/format-printer.ts
@@ -285,12 +285,16 @@ export class FormatPrinter extends Printer {
 
   /**
    * Format ERB content with proper spacing around the inner content.
-   * Returns empty string if content is empty, otherwise wraps content with single spaces.
+   * Returns empty string if content is empty, otherwise adds a leading space
+   * and a trailing space (or newline for heredoc content starting with "<<").
    */
   private formatERBContent(content: string): string {
     let trimmedContent = content.trim();
+
     // See: https://github.com/marcoroth/herb/issues/476
+    // TODO: revisit once we have access to Prism nodes
     let suffix = trimmedContent.startsWith("<<") ? "\n" : " "
+
     return trimmedContent ? ` ${trimmedContent}${suffix}` : ""
   }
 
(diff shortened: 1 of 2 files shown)


javascript/packages/formatter/src/format-printer.ts
```diff
@@ -288,7 +288,10 @@ export class FormatPrinter extends Printer {
    * Returns empty string if content is empty, otherwise wraps content with single spaces.
    */
   private formatERBContent(content: string): string {
-    return content.trim() ? ` ${content.trim()} ` : ""
+    let trimmedContent = content.trim();
+    // See: https://github.com/marcoroth/herb/issues/476
+    let suffix = trimmedContent.startsWith("<<") ? "\n" : " "
```
citizen428, line 296: This admittedly feels a bit hacky. But `content` is just a string, and we don't have another effective way to determine if we're dealing with a heredoc here, at least not one I can see. The change is also pretty minimal, fixes the described issue and doesn't break any of the related existing tests, so maybe it's good enough?

javascript/packages/formatter/test/erb/erb.test.ts
```diff
@@ -1578,6 +1578,19 @@ describe("@herb-tools/formatter", () => {
     expect(result).toBe(expected)
   })
 
+  test("ERB output with heredoc (issue 476)", () => {
+    const input = dedent`
```
citizen428, line 1582: This example has been directly taken from the issue.

javascript/packages/formatter/src/format-printer.ts
marcoroth: Yeah, I think this is good enough for now. The formatter is still considered experimental so I'm fine this for now, as it improves the current behavior and doesn't break any of the existing known behavior.

javascript/packages/formatter/test/erb/erb.test.ts
marcoroth: Great thanks!

I added some more test cases to also document what doesn't work yet so we can "lock in" some of the behavior. This should also help for when we want to change/refactor this in the future.

PR #1144, opened 2026-02-11
Parser: Add `analyze` option to `parser_options_T`

### LINKED ISSUE #631: Parser: Add new parser option for `analyze: false`
Add a new `parser_option_T` called `analyze` (or similar) and expose it to the `Herb.parse` C, JavaScript and Ruby APIs.

---

**When `true` (default and current behavior):**

We should call `herb_analyze_parse_tree()`

**When `false`:**

We should avoid calling `herb_analyze_parse_tree()`

---

**C-Cli**:
https://github.com/marcoroth/herb/blob/3e0cc8849e10de9d6d96c414683adfe62348c0d8/src/main.c#L98

**Node**:
https://github.com/marcoroth/herb/blob/3e0cc8849e10de9d6d96c414683adfe62348c0d8/javascript/packages/node/extension/herb.cpp#L105

**WASM**:
https://github.com/marcoroth/herb/blob/3e0cc8849e10de9d6d96c414683adfe62348c0d8/wasm/herb-wasm.cpp#L53

**Ruby**:
https://github.com/marcoroth/herb/blob/3e0cc8849e10de9d6d96c414683adfe62348c0d8/ext/herb/extension.c#L64

---

There are some use cases where we don't want or don't to analyze the Ruby code within the ERB tags, and for those cases we should have this `analyze: true/false` option.

Maybe we can also unify the `herb_analyze_parse_tree` call to a single location, so that each client-library doesn't have to implement and remember to call it.

citizen428:
Fixes #631 

This PR

1. adds a new option called `analyze` to `parser_options_T` and defaults it to `true` in `HERB_DEFAULT_PARSER_OPTIONS`,
2. moves the call to `herb_analyze_parse_tree()` into `herb_parse()` and guards it with an `if`,
3. exposes the options to all bindings and
4. updates the Rust bindings to be on par with other bindings when it comes to option handling.

### DIFF
--- a/ext/herb/extension.c
+++ b/ext/herb/extension.c
@@ -44,21 +44,19 @@ static VALUE Herb_parse(int argc, VALUE* argv, VALUE self) {
 
   char* string = (char*) check_string(source);
 
-  parser_options_T opts = HERB_DEFAULT_PARSER_OPTIONS;
+  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
 
   if (!NIL_P(options)) {
     VALUE track_whitespace = rb_hash_lookup(options, rb_utf8_str_new_cstr("track_whitespace"));
     if (NIL_P(track_whitespace)) { track_whitespace = rb_hash_lookup(options, ID2SYM(rb_intern("track_whitespace"))); }
-
-    if (!NIL_P(track_whitespace) && RTEST(track_whitespace)) { opts.track_whitespace = true; }
+    if (!NIL_P(track_whitespace) && RTEST(track_whitespace)) { parser_options.track_whitespace = true; }
 
     VALUE analyze = rb_hash_lookup(options, rb_utf8_str_new_cstr("analyze"));
     if (NIL_P(analyze)) { analyze = rb_hash_lookup(options, ID2SYM(rb_intern("analyze"))); }
-
-    if (!NIL_P(analyze) && !RTEST(analyze)) { opts.analyze = false; }
+    if (!NIL_P(analyze) && !RTEST(analyze)) { parser_options.analyze = false; }
   }
 
-  AST_DOCUMENT_NODE_T* root = herb_parse(string, &opts);
+  AST_DOCUMENT_NODE_T* root = herb_parse(string, &parser_options);
 
   VALUE result = create_parse_result(root, source);
 
--- a/javascript/packages/node/extension/herb.cpp
+++ b/javascript/packages/node/extension/herb.cpp
@@ -75,7 +75,7 @@ napi_value Herb_parse(napi_env env, napi_callback_info info) {
   char* string = CheckString(env, args[0]);
   if (!string) { return nullptr; }
 
-  parser_options_T opts = HERB_DEFAULT_PARSER_OPTIONS;
+  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
 
   if (argc >= 2) {
     napi_valuetype valuetype;
@@ -92,7 +92,7 @@ napi_value Herb_parse(napi_env env, napi_callback_info info) {
         napi_get_value_bool(env, track_whitespace_prop, &track_whitespace_value);
 
         if (track_whitespace_value) {
-          opts.track_whitespace = true;
+          parser_options.track_whitespace = true;
         }
       }
 
@@ -106,13 +106,13 @@ napi_value Herb_parse(napi_env env, napi_callback_info info) {
         napi_get_value_bool(env, analyze_prop, &analyze_value);
 
         if (!analyze_value) {
-          opts.analyze = false;
+          parser_options.analyze = false;
         }
       }
     }
   }
 
-  AST_DOCUMENT_NODE_T* root = herb_parse(string, &opts);
+  AST_DOCUMENT_NODE_T* root = herb_parse(string, &parser_options);
   napi_value result = CreateParseResult(env, root, args[0]);
 
   ast_node_free((AST_NODE_T *) root);
--- a/rust/src/herb.rs
+++ b/rust/src/herb.rs
@@ -53,12 +53,12 @@ pub fn parse_with_options(source: &str, options: &ParserOptions) -> Result<Parse
   unsafe {
     let c_source = CString::new(source).map_err(|e| e.to_string())?;
 
-    let mut c_opts = crate::bindings::parser_options_T {
+    let mut c_parser_options = crate::bindings::parser_options_T {
       track_whitespace: options.track_whitespace,
       analyze: options.analyze,
     };
 
-    let ast = crate::ffi::herb_parse(c_source.as_ptr(), &mut c_opts);
+    let ast = crate::ffi::herb_parse(c_source.as_ptr(), &mut c_parser_options);
 
     if ast.is_null() {
       return Err("Failed to parse source".to_string());
--- a/java/herb_jni.c
+++ b/java/herb_jni.c
@@ -25,7 +25,7 @@ JNIEXPORT jobject JNICALL
 Java_org_herb_Herb_parse(JNIEnv* env, jclass clazz, jstring source, jobject options) {
   const char* src = (*env)->GetStringUTFChars(env, source, 0);
 
-  parser_options_T opts = HERB_DEFAULT_PARSER_OPTIONS;
+  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
 
   if (options != NULL) {
     jclass optionsClass = (*env)->GetObjectClass(env, options);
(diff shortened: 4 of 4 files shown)


citizen428:
@marcoroth So far, this is just a PoC and should not change any behavior. If I see any build errors, I'll fix them. If you're ok with the overall approach, I'll update the various callers to deal with `analyze` in the same way as `track_whitespace`.

marcoroth:
@citizen428 thanks! Yes, this looks exactly what I had in mind! Please feel free to continue this 🙏🏼

ext/herb/extension.c
```diff
@@ -46,22 +45,18 @@ static VALUE Herb_parse(int argc, VALUE* argv, VALUE self) {
 
   char* string = (char*) check_string(source);
 
-  parser_options_T* parser_options = NULL;
```
citizen428, line 49: You'll see this pattern — or similar — in several places. We can no longer initialize with `{0}`, because unlike `track_whitespace`, `analyze` defaults to `true`. It also seems more sensible to actually use `HERB_DEFAULT_PARSER_OPTIONS` more pervasively.

javascript/packages/node/extension/herb.cpp
```diff
@@ -76,33 +75,44 @@ napi_value Herb_parse(napi_env env, napi_callback_info info) {
...
 
   if (argc >= 2) {
     napi_valuetype valuetype;
     napi_typeof(env, args[1], &valuetype);
 
     if (valuetype == napi_object) {
       napi_value track_whitespace_prop;
-      bool has_prop;
-      napi_has_named_property(env, args[1], "track_whitespace", &has_prop);
+      bool has_track_whitespace_prop;
```
citizen428, line 86: Renaming this seemed sensible.

rust/src/herb.rs
```diff
@@ -3,6 +3,21 @@ use crate::convert::token_from_c;
 use crate::{LexResult, ParseResult};
 use std::ffi::CString;
 
+#[derive(Debug, Clone)]
```
citizen428, line 6: @marcoroth: Please let me know if these Rust changes are out of scope for this PR, but it seemed to me like Rust was a bit behind the other implementations and it was an easy enough fix.

rust/src/herb.rs
```diff
@@ -31,16 +46,24 @@ pub fn lex(source: &str) -> Result<LexResult, String> {
 }
 
 pub fn parse(source: &str) -> Result<ParseResult, String> {
+  parse_with_options(source, &ParserOptions::default())
```
citizen428, line 49: We're just delegating here so `parse` works exactly like it did before.

rust/src/herb.rs
```diff
@@ -31,16 +46,24 @@ pub fn lex(source: &str) -> Result<LexResult, String> {
...
+}
+
+pub fn parse_with_options(source: &str, options: &ParserOptions) -> Result<ParseResult, String> {
   unsafe {
     let c_source = CString::new(source).map_err(|e| e.to_string())?;
-    let ast = crate::ffi::herb_parse(c_source.as_ptr(), std::ptr::null_mut());
+
+    let mut c_opts = crate::bindings::parser_options_T {
```
citizen428: @marcoroth: `herb_parse` takes a `parser_options_T *` as an argument, but it isn't `const`, which forces the Rust binding to use `&mut`. Since `herb_parse` doesn't mutate the options, the C signature could be changed to `const parser_options_T *` in a follow-up to make the intent clearer and allow us to use `&` instead of `&mut` on the Rust side. Let me know if you want me to do this as part of this PR.

citizen428:
@marcoroth This should be ready for review now 😅 Please point out anything you can think of, this involved quite a few parts of the codebase I hadn't touched before. I left some inline comments to help a bit with reviewing.

rust/src/herb.rs
marcoroth: Good point, this wasn't intentional and is something I'd be happy to change in a follow-up, thanks!

rust/src/herb.rs
marcoroth: I think this is totally fine to keep here, as it only really become relevant now with the new option 👍🏼

rust/src/herb.rs
marcoroth: Does it make sense to change the method signature here and make `parse` be a `parse(source: &str, options: &ParserOptions)`, maybe even using a `impl Into<Option<T>>` so you can also pass `None`?

I haven't done Rust enough to know what the more natural and/or idiomatic way here would be.

wasm/herb-wasm.cpp
```diff
@@ -35,22 +34,25 @@ val Herb_lex(const std::string& source) {
 }
 
 val Herb_parse(const std::string& source, val options) {
-  parser_options_T* parser_options = nullptr;
-  parser_options_T opts = {0};
+  parser_options_T opts = HERB_DEFAULT_PARSER_OPTIONS;
```
marcoroth: Since we removed `parser_options`, does it make sense to rename `opts` to `parser_options` now?


```suggestion
  parser_options_T parser_options = HERB_DEFAULT_PARSER_OPTIONS;
```

javascript/packages/node/test/node.test.ts
```diff
@@ -58,6 +58,20 @@ describe("@herb-tools/node", () => {
     )
   })
 
+  test("parse() with analyze: true (default) transforms ERB nodes", async () => {
```
marcoroth, line 61: Would you mind porting these tests to `node-wasm`,`rust/` and `java/` too? I think testing the top-level options of the public API makes sense to test in all bindings. Thank you!

ext/herb/extension.c
marcoroth: I think this is reasonable and probably even clearer/more intentional anyway 👍🏼

citizen428:
I should be able to get to those changes tomorrow. I'm actually a bit surprised there wasn't more 😂

rust/src/herb.rs
citizen428: TBH, I did the quickest thing that could work here without changing the API surface, because I wasn't sure if we'll keep this change around as part of this PR. Let me think about the best approach.

rust/src/herb.rs
citizen428: @marcoroth I think I'll do this as a true follow-up, i.e. not as part of this PR. It's getting pretty big already, and I don't want to make any additional changes at this point.

rust/src/herb.rs
citizen428: @marcoroth Looking around some Rust libraries, the current pattern is not without precedent. Updating it to your suggestion leads to a pretty complicated signature compared to the two simple ones. Something like this:

```rust
fn parse<'a>(source: &str, options: impl Into<Option<&'a ParserOptions>>) -> Result<ParseResult, String>
```

Do we know if the Rust code gets much use right now? Maybe it's best to keep the two-function approach for now and see how many more options we add over time. Then we can consider using a Builder pattern in Rust as some libraries do. Something like:

```rust
Parser::new().track_whitespace(true).analyze(false).parse()
```

This is a well-established pattern, here are some examples:

```rust
OpenOptions::new().read(true).open()
RegexBuilder::new().case_insensitive(true).build()
```

javascript/packages/node/test/node.test.ts
citizen428: I added them for `node-wasm/` and `rust/`. When tackling Java, I realized that the current CLI-based snapshot tests have one issue: the Java CLI doesn't really expose these options right now, only the subcommand handling. The same is true for C. How would you feel about tackling this in a follow-up PR, as it may become quite a big change depending on how we want to deal with the options?

citizen428:
@marcoroth I made some/most of the changes requested. Here are things that still need decisions, compiled for your convenience:

1. [Is the current Rust code good enough, and can we delay a more sophisticated approach until we have more options?](https://github.com/marcoroth/herb/pull/1144#discussion_r2796651446)
2. [Can we delay adding Java tests since the CLI doesn't really expose the options yet?](https://github.com/marcoroth/herb/pull/1144#discussion_r2796664109)
3. [Is it ok to change the function signature of `parse` in C in a separate PR?](https://github.com/marcoroth/herb/pull/1144#discussion_r2796585372)

rust/src/herb.rs
marcoroth: That's fine with me, thanks! 🙏🏼

javascript/packages/node/test/node.test.ts
marcoroth: Ah right, I see. That's a good point. Let's keep them as-is now, no need to come up with something for the Java CLI. I think that, ideally, we could use a regular/proper test framework for the Java bindings too, so we don't rely on the `run_tests.sh` script.

rust/src/herb.rs
marcoroth: That's fair, thanks for checking the options. I'm probably more in Ruby/JavaScript mode where it's convenient to have optional arguments with default values.

Currently, I'm not aware of anyone using the Rust bindings, so I'm happy with changing the API if we need to. But I think it's fine for now if we just keep it as-is and maybe later, if we need to, implement a builder pattern-like API for the options.

Thank you!

PR #1145, opened 2026-02-11
CI: Reorder build steps to fail fast on cheap checks

citizen428:
During my work on #1144 I noticed that CI runs some of the more expensive build steps before some of the faster lints. So this PR reorders them to an order that seems a bit more sensible to me.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,14 +44,14 @@ jobs:
       - name: Render Templates
         run: bundle exec rake templates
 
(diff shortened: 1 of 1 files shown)


citizen428:
@marcoroth Ha, interesting. I guess I optimized for cycle time, because I'd always notice locally if something doesn't build. Anyway, no hard feelings if you don't want to merge this, I wasn't aware of your rationale for having it in this order, and most projects I've worked on do the quick things first.

PR #1152, opened 2026-02-13
Java: Add JUnit and proper Java tests

citizen428:
In a [recent discussion](https://github.com/marcoroth/herb/pull/1144#discussion_r2801177249) @marcoroth mentioned the following:

>  think that, ideally, we could use a regular/proper test framework for the Java bindings too, so we don't rely on the `run_tests.sh` script.

This PR removes the snapshot-based CLI tests and replaces them with more conventional tests, though without a full testing framework yet.

**Note:**
The Java part of this project currently doesn't use a build tool like Maven or Gradle, just `javac` and a `Makefile`. I did not want to go down this rabbit hole and the potential associated discussions about the relative merits of the options. Instead, I opted for a couple of simple assertion helpers and a hand-rolled test runner. Since we're not trying to retest the entire API surface but just the wrapping code, this approach seemed sensible to me.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -34,10 +34,10 @@ jobs:
         uses: actions/cache@v4
         with:
           path: java/lib
-          key: ${{ runner.os }}-junit-1.14.2
+          key: ${{ runner.os }}-junit-${{ hashFiles('java/Makefile') }}
 
       - name: Ensure junit jar
-        run: make lib/junit-platform-console-standalone-1.14.2.jar
+        run: make ensure-junit
         working-directory: java
 
       - name: Render Templates
--- a/java/Makefile
+++ b/java/Makefile
@@ -63,6 +63,10 @@ $(JUNIT_JAR): | lib
 		echo "$(JUNIT_JAR) already exists"; \
 	fi
 
+.PHONY: ensure-junit
+ensure-junit: $(JUNIT_JAR)
+	@echo "JUnit standalone jar ensured: $(JUNIT_JAR)"
+
 .PHONY: all
 all: templates herb jni java
 


.github/workflows/java.yml
```diff
@@ -31,23 +31,13 @@ jobs:
...
 
       - name: Render Templates
-        run: bundle exec rake templates
-
-      - name: Compile Herb
-        run: bundle exec rake make
-
-      - name: Build JNI library
-        run: make jni
+        run: make templates
```
citizen428: We already had a `Makefile` target for the templates, so I decided to use that. In my own projects I generally wrap up most things in simple `make` tasks so the easier to remember and I'll always invoke them the same way.

.github/workflows/java.yml
```diff
@@ -31,23 +31,13 @@ jobs:
...
         working-directory: java
 
-      - name: Compile Java classes
-        run: make java
-        working-directory: java
-
-      - name: Run tests
-        run: ./run_tests.sh
+      - name: Compile Herb
+        run: make herb
```
citizen428: I added this target to be consistent with `make templates`.

java/org/herb/HerbTest.java
```diff
@@ -0,0 +1,182 @@
+package org.herb;
+
+// Note: We currently don't have a Java build tool. Instead of going down the
```
citizen428: I figured I'll leave a comment about the rationale here too, so future readers have some context.

java/Makefile
```diff
@@ -48,13 +48,18 @@ JNI_LIB = $(BUILD_DIR)/$(JNI_LIB_PREFIX)herb_jni.$(JNI_LIB_EXT)
 JAVA_SOURCES = $(shell find org -name "*.java" 2>/dev/null)
 JAVA_TIMESTAMP = target/.java_compiled
 
-.PHONY: all clean java jni test
-
-all: templates jni java
+.PHONY: all
```
citizen428, line 70: To be consistent with the main `Makefile`, we're now labeling targets explicitly as `.PHONY`. The list here was already outdated.

java/Makefile
```diff
@@ -48,13 +48,18 @@ JNI_LIB = $(BUILD_DIR)/$(JNI_LIB_PREFIX)herb_jni.$(JNI_LIB_EXT)
...
+all: templates herb jni java
```
citizen428, line 71: I added `herb` here. It's a bit expensive, so if you want, we can remove it again. But technically, this is the full list of deps we need.

java/Makefile
```diff
@@ -75,24 +81,35 @@ $(JAVA_TIMESTAMP): $(JAVA_SOURCES)
 		echo "Compiled Java classes"; \
 	fi
 
+.PHONY: test
+test: jni java
+	$(JAVA_CMD) --enable-native-access=ALL-UNNAMED \
```
citizen428, line 105: The `--enable-native-access=ALL-UNNAMED` gets rid of the following warning, that's caused by native access from classes loaded from the classpath into an unnamed module:

> WARNING: A restricted method in java.lang.System has been called
> WARNING: java.lang.System::loadLibrary has been called by org.herb.Herb in an unnamed module (file:/Users/me/src/herb/java/target/classes/)
> WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
> WARNING: Restricted methods will be blocked in a future release unless native access is enabled

java/Makefile
```diff
@@ -65,34 +71,52 @@ $(JNI_LIB): $(JNI_OBJECTS) $(HERB_OBJECTS)
...
+		$(JAVAC) -cp $(JUNIT_JAR) -d target/classes $(JAVA_SOURCES); \
 		touch $(JAVA_TIMESTAMP); \
 		echo "Compiled Java classes"; \
 	fi
 
+.PHONY: test
+test: jni java
+	$(JAVA_CMD) --enable-native-access=ALL-UNNAMED \
+		-Djava.library.path=$(BUILD_DIR) \
+		-jar $(JUNIT_JAR) execute \
```
citizen428, line 107: This is the fat jar we downloaded into `./lib/`.

citizen428:
@marcoroth Based on our discussion, I restored the snapshots and switched to actually using JUnit. But there's one caveat: I still didn't want to mess around with a build tool, so I downloaded the fat jar version into `lib/java` and called it a day. It's a 3.2 MB or so file, which is not ideal, but I think also not a deal breaker. This seems GoodEnough™ for me now.

marcoroth (approved):
Thank you @citizen428!

I'm happy with JUnit and also that we have a proper test framework for the Java tests now! I slightly updated the way we fetch the junit jar, so that we don't have to commit the jar file to the repo. But otherwise this is great, thank you so much! 🙏🏼

citizen428:
> I slightly updated the way we fetch the junit jar, so that we don't have to commit the jar file to the repo.

I had considered that. I'm not a fan of having a network dependency for tests runs, but I guess that's a moot point on GH actions anyway. 😅

marcoroth:
Yeah, that's a fair point! But that's also why I added the cache step so it doesn't have to fetch the .jar file on every action run! 

I think it should be fine for now, and otherwise we can still add the .jar file 🙌🏼

Issue #1158, opened 2026-02-13
Linter Rule: Require explicit closing tags


marcoroth:
**Rule:** `html-require-closing-tags`

## Description

Disallow the omission of optional closing tags for HTML elements where the closing tag is technically optional according to the HTML specification. This rule flags elements that have an `HTMLOmittedCloseTagNode` as their close tag.

## Rationale

While HTML allows certain closing tags to be omitted (implicitly closed by sibling elements or parent closing), explicit closing tags improve code readability and maintainability. They make the document structure clear at a glance and reduce ambiguity about where elements end.

Explicit closing tags also:
- Make templates easier to understand for developers unfamiliar with HTML's implicit closing rules
- Reduce potential for subtle bugs when refactoring or moving code
- Improve consistency across the codebase
- Make diffs cleaner when adding content to elements

## Elements with Optional Closing Tags

This rule would flag elements that have omitted closing tags:

- `<li>` - list items
- `<dt>`, `<dd>` - definition list terms and descriptions
- `<p>` - paragraphs
- `<option>`, `<optgroup>` - select options and groups
- `<thead>`, `<tbody>`, `<tfoot>` - table sections
- `<tr>` - table rows
- `<td>`, `<th>` - table cells
- `<colgroup>` - table column groups
- `<rt>`, `<rp>` - ruby annotations

### ✅ Good

Explicit closing tags:

```erb
<ul>
  <li>Item 1</li>
  <li>Item 2</li>
  <li>Item 3</li>
</ul>
```

```erb
<dl>
  <dt>Term 1</dt>
  <dd>Definition 1</dd>
  <dt>Term 2</dt>
  <dd>Definition 2</dd>
</dl>
```

```erb
<table>
  <thead>
    <tr>
      <th>Header 1</th>
      <th>Header 2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Cell 1</td>
      <td>Cell 2</td>
    </tr>
  </tbody>
</table>
```

```erb
<select>
  <option>Option 1</option>
  <option>Option 2</option>
  <option>Option 3</option>
</select>
```

```erb
<div>
  <p>Paragraph 1</p>
  <p>Paragraph 2</p>
</div>
```

### 🚫 Bad

Omitted closing tags (implicitly closed):

```erb
<ul>
  <li>Item 1
  <li>Item 2
  <li>Item 3
</ul>
```

```erb
<dl>
  <dt>Term 1
  <dd>Definition 1
[...]

Issue #1157, opened 2026-02-13
Linter Rule: Disallow conditional open tags


marcoroth:
**Rule:** `erb-no-conditional-open-tag`

## Description

Disallow the pattern of using ERB conditionals to vary only the open tag of an element (typically to conditionally change attributes) while sharing the same tag name, body content, and close tag.

## Rationale

This pattern can be difficult to read and maintain. The conditional logic is split from the element's body and closing tag, making it harder to understand the full structure at a glance.

Prefer using:
- `content_tag` with conditional attributes
- A ternary or conditional for the class/attribute value directly
- Separate complete elements in each branch if the differences are significant

### ✅ Good

Using conditional attributes directly:

```erb
<div class="<%= @condition ? 'a' : 'b' %>">
  Content
</div>
```

Using `content_tag` with conditional options:

```erb
<%= content_tag :div, class: (@condition ? 'a' : 'b') do %>
  Content
<% end %>
```

Using `class_names` helper:

```erb
<div class="<%= class_names('base', 'a': @condition, 'b': !@condition) %>">
  Content
</div>
```

Complete separate elements when differences are significant:

```erb
<% if @condition %>
  <div class="a" data-foo="bar">Content</div>
<% else %>
  <div class="b" data-baz="qux">Content</div>
<% end %>
```

Self-closing/void elements in conditionals (each branch is complete):

```erb
<% if @large %>
  <img src="photo.jpg" width="800" height="600">
<% else %>
  <img src="photo.jpg" width="400" height="300">
<% end %>
```

### 🚫 Bad

Conditional open tags with shared body and close tag:

```erb
<% if @condition %>
  <div class="a">
<% else %>
  <div class="b">
<% end %>
  Content
</div>
```

```erb
<% if @with_icon %>
  <button class="btn btn-icon" aria-label="Action">
<% else %>
  <button class="btn">
<% end %>
  Click me
</button>
```

Open tag in conditional without else branch:

```erb
<% if @wrap %>
  <div class="wrapper">
<% end %>
  Content
</div>
```

Missing open tag in else branch:

```erb
<% if @style == "a" %>
  <div class="a">
<% elsif @style == "b" %>
  <div class="b">
<% else %>
<% end %>
  Content
</div>
```

Missing open tag in elsif branch:

```erb
<% if @style == "a" %>
  <div class="a">
<% elsif @style == "b" %>
  <%# no-open-tag %>
<% else %>
  <div class="c">
<% end %>
  Content
</div>
[...]

PR #1168, opened 2026-02-14
Linter: Implement `erb-no-trailing-whitespace` rule

### LINKED ISSUE #547: erb-lint: TrailingWhitespace
https://github.com/Shopify/erb_lint/blob/main/lib/erb_lint/linters/trailing_whitespace.rb
https://github.com/Shopify/erb_lint/blob/main/spec/erb_lint/linters/trailing_whitespace_spec.rb

markokajzer:
closes #547

Implements the `TrailingWhitespace` rule from [ERBLint](https://github.com/Shopify/erb_lint).

This rule disallows whitespace (spaces, tabs, carriage return) at the end of a line.

Unfamiliar with the project, I took inspiration from #557, which implemented a similar rule.

### DIFF
--- a/javascript/packages/linter/docs/rules/erb-no-trailing-whitespace.md
+++ b/javascript/packages/linter/docs/rules/erb-no-trailing-whitespace.md
@@ -4,7 +4,17 @@
 
 ## Description
 
-Disallow trailing whitespace (spaces, tabs, carriage returns) at the end of lines in ERB templates. This rule detects and removes any invisible whitespace characters that appear after the last visible character on a line.
+Disallow trailing whitespace (spaces, tabs, carriage returns, and other whitespace characters) at the end of lines in ERB templates.
+
+## Skipped Content
+
+This rule does **not** flag trailing whitespace inside:
+
+* `<pre>` - Preformatted text where whitespace is significant
+* `<textarea>` - User input where whitespace may be intentional
+* `<script>` - Treated as foreign content
+* `<style>` - Treated as foreign content
+* ERB tags (`<% %>`, `<%= %>`, `<%# %>`) - Ruby code where trailing whitespace could be significant (heredocs, string literals)
 
 ## Rationale
 
@@ -19,13 +29,17 @@ Trailing whitespace is invisible and serves no purpose, but it can cause several
 
 ### ✅ Good
 
-```html
+```erb
 <div>
   <p>Hello</p>
 </div>
+```
 
+```erb
 <%= content %>
+```
 
+```erb
 <div>
   <h1>Title</h1>
 </div>
@@ -35,11 +49,15 @@ Trailing whitespace is invisible and serves no purpose, but it can cause several
 
 ```erb
 <div>···
-  <p>Hello</p>··
+  <p>Hello</p>
 </div>
+```
 
+```erb
 <%= content %>·
+```
 
+```erb
 Hello→→
 World···
 ```
--- a/javascript/packages/linter/src/rules/erb-no-trailing-whitespace.ts
+++ b/javascript/packages/linter/src/rules/erb-no-trailing-whitespace.ts
@@ -1,44 +1,70 @@
-import  { type Node, Location } from "@herb-tools/core"
+import { Location, isHTMLOpenTagNode, isHTMLTextNode, isLiteralNode, Visitor } from "@herb-tools/core"
+import { getTagName, findNodeAtPosition } from "./rule-utils.js"
+import { ParserRule, Mutable, BaseAutofixContext } from "../types.js"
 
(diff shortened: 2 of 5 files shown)


marcoroth (commented):
Hey @markokajzer, thanks for taking a stab at this!

This looks great already and is a really accurate port of the linter rule as it was at play in `erb_lint`.

However, I have a slight concern with autofix for this rule. I feel like we need to be a bit more careful and cannot just strip whitespace at the end of each line, as some of the whitespace might be relevant.

So I feel like we might have to change this linter rule to be a `ParserRule` (opposed to the `SourceRule` we have now) and skip contexts like `<pre>` and `<textarea>` . Maybe even `<script>` and `<style>` too.

But I also realize that the implementation of this would become a lot trickier when we have to operate on AST nodes for a rule like this.

Do you have any thoughts on this? Thank you!

javascript/packages/linter/docs/rules/erb-trailing-whitespace.md
```diff
@@ -0,0 +1,51 @@
+# Linter Rule: Disallow trailing whitespace at end of lines
+
+**Rule:** `erb-trailing-whitespace`
```
marcoroth: I think `erb-no-trailing-whitespace` might be a bit more closly following the naming of the other linter rules we have already.


```suggestion
**Rule:** `erb-no-trailing-whitespace`
```

markokajzer:
I actually fear that might exceed my skills lol

How would be go about this? Traverse all nodes, and if we're at the end of the line, and not within one of the mentioned nodes, delete the whitespace? (Haven't looked at `ParserRule` yet: does a node know where we are in a line?)

On the other hand, does `erb_lint` handle those correctly? If not, people using `erb_lint` should already be familiar with such intricacies? :trollface:

javascript/packages/linter/docs/rules/erb-trailing-whitespace.md
markokajzer: renamed 👍

marcoroth:
> How would be go about this? Traverse all nodes, and if we're at the end of the line, and not within one of the mentioned nodes, delete the whitespace? (Haven't looked at `ParserRule` yet: does a node know where we are in a line?)

Yeah, sadly that's the super tricky part here. When we are operating on AST Nodes we don't really think in terms of lines anymore. But, the AST Nodes would have all required information on them to be able to figure that out again and map it back. It probably would just mean a lot of new rule helpers to make it easier to work with lines in a `ParserRule`. But yeah, this might be a bit overkill for now.

That said, you are right that this is already a problem in `erb_lint` today. But given that we also support an autofix with this rule now, this will automatically be fixed in the Language Server if you open a file with trailing whitespace and save it.

What I would recommend is to tell the rule that it's not auto-correctable set `static autocorrectable = false` and leave the current implementation as-is.

javascript/packages/linter/docs/rules/erb-trailing-whitespace.md
marcoroth: Thank you! 🙏🏼

markokajzer:
> `static autocorrectable = false`

lower severity to warning in the same vein?

marcoroth:
Hey @markokajzer, sorry for pushing up directly to your branch. 

I addressed the concerns by converting the rule from a `SourceRule` to a `ParserRule` with "smart skip zones".

The detection now works in multiple steps: First, we split the source into lines and scan each line for trailing whitespace using a regex, this is pretty much what we had before, but this gives us a list of trailing whitespace candidates with their line and column positions.

Next, we traverse the AST using a visitor once to collect the "skip zones". They are regions where trailing whitespace should be preserved. This includes the `<pre>`, `<textarea>`, `<script>`, `<style>` elements, and all ERB tags. ERB tags are skipped because trailing whitespace inside Ruby code could be significant (heredocs, string literals, etc.). Maybe we can be smarter about this in the future and also enable it for ERB tags.

Finally, we filter the candidates against the skip zones. Only candidates that fall outside all skip zones become offenses. This approach gives us the simplicity of the source-based detection we had before, but with the accurate context-awareness of AST-based filtering.

Since we now skip sensitive contexts, autofix is safe to enable again. I added some more tests to make sure we cover all of the tricky edge cases. 

Thanks for starting this and opening a pull request! 🙏🏼

markokajzer:
Glad we’ll be able to merge something, even if you had to do all the work after all :^)

I see you have a high standard for this project haha

Might try again on another rule 👍

marcoroth:
Sorry to have hijacked it 🙈 

But please don't be discouraged to tackle another rule, I'm happy to help steer you the right way. For example both #1157 and #1158 should be a lot more straightforward, as they are about detecting the presence of a certain node type.

PR #1190, opened 2026-02-16
Parser: Store parser options on `ParseResult`

marcoroth:
This pull request updates the `ParseResult` in all bindings to store the `parser_options_T` that were use to parse the document on the `ParseResult` itself.

**Ruby**
```ruby
result = Herb.parse("<div><%= hello %></div>")
result.options
#=> #<Herb::ParserOptions strict=true track_whitespace=false analyze=true>
```

```ruby
result = Herb.parse("<div><%= hello %></div>", strict: false, track_whitespace: true)
result.options.strict             #=> false
result.options.track_whitespace   #=> true
result.options.analyze            #=> true
```

**JavaScript**

```js
const result = parse("<div><%= hello %></div>")
result.options
// ParserOptions { strict: true, track_whitespace: false, analyze: true }
```

### DIFF
--- a/ext/herb/extension_helpers.c
+++ b/ext/herb/extension_helpers.c
@@ -73,7 +73,7 @@ VALUE create_lex_result(hb_array_T* tokens, VALUE source) {
   return rb_class_new_instance(4, args, cLexResult);
 }
 
-VALUE create_parse_result(AST_DOCUMENT_NODE_T* root, VALUE source, parser_options_T* options) {
+VALUE create_parse_result(AST_DOCUMENT_NODE_T* root, VALUE source, const parser_options_T* options) {
   VALUE value = rb_node_from_c_struct((AST_NODE_T*) root);
   VALUE warnings = rb_ary_new();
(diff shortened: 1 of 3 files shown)


ext/herb/extension_helpers.c
```diff
@@ -73,14 +73,22 @@ VALUE create_lex_result(hb_array_T* tokens, VALUE source) {
   return rb_class_new_instance(4, args, cLexResult);
 }
 
-VALUE create_parse_result(AST_DOCUMENT_NODE_T* root, VALUE source) {
+VALUE create_parse_result(AST_DOCUMENT_NODE_T* root, VALUE source, parser_options_T* options) {
```
citizen428: Do we maybe want this to be `const` to make it explicit that this function will not modify the options?

marcoroth: Good point! Addressed in 464a35212d3fd531e81b95caf0d3afd2d89e736a, thanks!

PR #1199, opened 2026-02-16
Linter: Implement `html-require-closing-tags` rule

### LINKED ISSUE #1158: Linter Rule: Require explicit closing tags

markokajzer:
closes #1158


javascript/packages/linter/test/helpers/linter-test-helper.ts
```diff
@@ -222,7 +224,7 @@ export function createLinterTest(rules: RuleClass | RuleClass[], configOverride?
     })
 
     const linter = new Linter(Herb, ruleClasses, config)
-    const lintResult = linter.lint(html, context)
+    const lintResult = linter.lint(html, context, { strict })
```
markokajzer: In strict mode (which is now the default) the parser produces a `OMITTED_CLOSING_TAG_ERROR` recursive error and we do not even get to the rule evaluation.

So we need to adjust the linter to be able to lint with `strict: false`

javascript/packages/linter/src/linter.ts
```diff
@@ -297,13 +297,13 @@ export class Linter {
...
    */
-  lint(source: string, context?: Partial<LintContext>): LintResult {
+  lint(source: string, context?: Partial<LintContext>, parserOptions?: Partial<ParserOptions>): LintResult {
     this.offenses = []
 
     let ignoredCount = 0
     let wouldBeIgnoredCount = 0
 
-    const parseResult = this.herb.parse(source, { track_whitespace: true })
+    const parseResult = this.herb.parse(source, { track_whitespace: true, ...parserOptions })
```
markokajzer: Similarly, if we wanted to have an autofix for this rule, we'd always have to lint with `strict: false`. I assume that is actually not what we want here?

markokajzer: Could also just be a misconception haha

javascript/packages/linter/test/helpers/linter-test-helper.ts
marcoroth: That's a good point, great catch!

I think we could either have an exception for this rule and let the linter parse the document using `strict: false` or update the `FullRuleConfig` to also allow to specific parser options. Then, we could group all rules by unique parser options and pass each linter the rule the document they expect.

Do I see this right that this works fine now because we are explicitly passing `strict: false` in the tests?

javascript/packages/linter/src/rules/html-require-closing-tags.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const tagName = getTagName(node)
+    if (!tagName) return
+
+    this.addOffense(
+      `Missing explicit closing tag for \`<${tagName}>\`. Use \`</${tagName}>\` instead of relying on implicit tag closing.`,
+      node.open_tag.location
+    )
+  }
+}
```
marcoroth: You might be able to simplify this by directly visiting the `HTMLOmittedCloseTagNode`:
```suggestion
  visitHTMLOmittedCloseTagNode(node: HTMLOmittedCloseTagNode): void {
    const tagName = node.tag_name?.value
    if (!tagName) return

    this.addOffense(
      `Missing explicit closing tag for \`<${tagName}>\`. Use \`</${tagName}>\` instead of relying on implicit tag closing.`,
      node.location
    )
  }
}
```

javascript/packages/linter/test/helpers/linter-test-helper.ts
markokajzer: adding to `FullRuleConfig` sounds reasonable?

> Do I see this right that this works fine now because we are explicitly passing strict: false in the tests?

yes. the test passes `strict` to the test helper, which passes it to the linter.

if we add this info to `FullRuleConfig` or an exception to linter directly, I assume this will be taken care of automatically 👍

markokajzer:
I added the following changes
1. When linter starts, build a “parse cache” for each `parserOptions` combination of the running rules
    1. When a rule tries to lint, use the `parseResult` for its set of `parserOptions`
2. When a rule lints in testing, also retrieve its `parserOptions` before parsing the passed html to avoid having to pass `strict: false` parameter from the test to the test helper to the linter itself

I’m actually not fully convinced of these, especially the first one haha
We now have to parse the same file multiple times depending on the parser options 🤔 

I also tried to add an autofix. We can make this as complicated as we like. I started with a simpler solution at first, that just appends the missing tag, but then also tried to preserve whitespace for nested elements with missing closing tags (e.g. within `table`) which complicates stuff a bit more haha

Maybe you can help coming up with some more edge cases?

This was definitely not as straight forward as I hoped. I’ll try to find the easiest possible rule next time 😸

markokajzer:
thank you for taking care, and sorry to put more work on your plate haha

Issue #162, opened 2025-06-25
Linter Rule: Disallow inline `<style>` tags and `style` attributes


marcoroth:
##### Description

Disallow the use of inline `<style>` tags and `style` attributes in HTML templates. All styling should be applied via external stylesheets or class-based styling.

##### Rationale

Inline CSS makes templates harder to maintain, promotes duplication, and prevents adoption of strict CSP policies (`style-src 'self'`). Centralizing styles via external stylesheets improves maintainability and security.

This rule enforces:

- No `<style>` tags embedded directly in templates.
- No inline `style` attributes on HTML elements.

##### ✅ Good

```html+erb
<head>
  <%= stylesheet_link_tag "application" %>
</head>

<button class="btn btn-primary">Submit</button>
```

##### 🚫 Bad

```html+erb
<head>
  <style>
    .danger { color: red; }
  </style>
</head>

<button style="color: red;">Submit</button>
```

#### References

Inspired by [@pushcx](https://bsky.app/profile/push.cx/post/3lsfddauapk2o)

andy-rootly:
This would be a great one to have, especially to prevent AI agents from trying to use inline styles.

PR #1201, opened 2026-02-17
Linter: Implement `html-allowed-script-type` rule

### LINKED ISSUE #554: erb-lint: AllowedScriptType
https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#allowedscripttype

https://github.com/Shopify/erb_lint/blob/main/lib/erb_lint/linters/allowed_script_type.rb
https://github.com/Shopify/erb_lint/blob/main/spec/erb_lint/linters/allowed_script_type_spec.rb

markokajzer:
closes #554

## questions

1. Do we want the rule to be configurable (like the original)?
2. Do we want the `disallow_inline_scripts` option? It kind of goes a bit further than just checking the types. Maybe it should be its own rule?

### DIFF
--- a/javascript/packages/linter/docs/rules/html-allowed-script-type.md
+++ b/javascript/packages/linter/docs/rules/html-allowed-script-type.md
@@ -4,7 +4,7 @@
 
 ## Description
 
-Restricts which `type` attribute values are permitted on `<script>` tags. Only an approved whitelist of types is allowed: `text/javascript`. A missing `type` attribute is also reported.
+Restricts which `type` attribute values are permitted on `<script>` tags. Only approved types are allowed: `text/javascript`. An empty or valueless `type` attribute is reported.
 
 ## Rationale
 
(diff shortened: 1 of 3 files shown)


marcoroth (commented):
Hey @markokajzer, thanks for taking this on!

1. Sadly we cannot really add custom configuration for linter rules yet, this is something we should add in the future though. 

    That said, I guess we can leave it for now, keep the `ALLOW_BLANK` const in the rule for later, but maybe add a `// TODO` comment so know that it needs some work. 

    I'd also say since it's not configurable, that we set `ALLOW_BLANK` to `true` to begin with, so it's not as aggressive.

    Edit: I opened https://github.com/marcoroth/herb/issues/1204 to track this

3. Yeah, I think we can keep this separate. There is also #162 we can implement separately.

markokajzer:
added comment and flipped the switch 👍

PR #1206, opened 2026-02-19
Engine: Fix ERB expression compilation when code contains a heredoc

pinzonjulian:
When an ERB output tag contains a heredoc (e.g. `<%= method_call <<~GRAPHQL, variables %>`), the closing parenthesis in the compiled Ruby must appear on its own line after the heredoc terminator. Without this, the compiled Ruby is syntactically invalid.

This is a sibling commit of https://github.com/marcoroth/reactionview/pull/78

### DIFF
--- a/lib/herb/engine.rb
+++ b/lib/herb/engine.rb
@@ -213,13 +213,17 @@ module Herb
 
     def add_expression_result(code)
       with_buffer {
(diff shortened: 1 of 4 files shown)


marcoroth (approved):
Thank you so much @pinzonjulian! 🙏🏼

Issue #1183, opened 2026-02-16
Linter Rule: Ensure id is present with data-turbo-permanent attribute


pardeyke:
**Rule:** `[id-on-data-turbo-permanent]`

## Description

Make sure an id is set on elements that also have data-turbo-permanent.
Without an id turbo cant track the element and the permanentness wold not work as expected.

## Rationale

Ensures Turbo is working correctly 

### ✅ Good

```erb
<%= content_tag(:div, "Hello world!",  data: { turbo-permanent: true }, id: "hello-world" %>
```

### 🚫 Bad

```erb
<%= content_tag(:div, "Hello world!",  data: { turbo-permanent: true } %>
```

## References

[Turbo Handbook](https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads)

PR #1229, opened 2026-02-22
Linter: Implement `turbo-permanent-require-id` rule

pardeyke:
#1183


marcoroth:
Hey @pardeyke, thanks for working on this!

We have been waiting on adding Turbo-related rules to the main Herb Linter, similar to how we extracted the Stimulus-related rules to `stimulus-lint`.

The pull request #677 implements a similar rule and a `turbo-lint` package, but sadly the name is not available anymore. Ideally, we could work on adding a plugin system to the Herb Linter, where NPM packages can provide a plugin with a set of linter rules.

Maybe the `stimulus-*` and `turbo-*` rules and/or plugins could be built-in to start with. Would you mind renaming the rule to `turbo-permanent-require-id` so we have the rules "scoped" by the first part of the rule name. Thank you!

pardeyke:
Im not sure if I get what you mean. The rule is names `HTMLTurboPermanentRequireId`. Do you want me to remove the `HTML` part?

marcoroth:
Yeah, so it's kind of implicitly  "scoped" to a `turbo-*` namespace and that we can later extract it to its own plugin/package.

marcoroth (commented):
Hey @pardeyke, looks we got two rules and two test files, would you mind removing the `html-*` one too?

Can we also add a docs page in `javascript/packages/linter/docs/rules/*` as well as short mention in the `javascript/packages/linter/docs/rules/README.md` to align with the other rules? 

Thank you!

brunoprietog:
Hi, sorry for being late to this, but this is not strictly true when used with page refreshes and morphing, where the ID is not required in elements with `data-turbo-permanent`.

PR #1268, opened 2026-03-02
Linter: Make `RuleClass().name` static

### LINKED ISSUE #1264: Linter: Make `RuleClass().name` static
The `name` is a constant string per rule class, it never changes between instances. Making it static would avoid throwaway instantiations just to read the name. There are at least 6 places in `linter.ts` alone that do `new RuleClass().name` purely to look up a rule by name.

citizen428:
Closes #1264

This PR refactors the linter code to use a static `ruleName` property instead of `name` on the instance.

### DIFF
--- a/javascript/packages/linter/src/custom-rule-loader.ts
+++ b/javascript/packages/linter/src/custom-rule-loader.ts
@@ -75,8 +75,14 @@ export class CustomRuleLoader {
       const cacheBustedUrl = `${fileUrl}?t=${Date.now()}`
       const module = await import(cacheBustedUrl)
 
-      if (module.default && this.isValidRuleClass(module.default)) {
+      if (this.isValidRuleClass(module.default)) {
         return [module.default]
+      }
+
+      if (this.hasDeprecatedNameProperty(module.default)) {
+        const name = new module.default().name
+        console.error(`Error: Custom rule in "${filePath}" sets 'name' as an instance property, which is no longer supported. Use 'static ruleName = "${name}"' instead.`)
+        process.exit(1)
       }
 
       if (!this.silent) {
@@ -96,10 +102,22 @@ export class CustomRuleLoader {
    * Type guard to check if an export is a valid rule class
    */
   private isValidRuleClass(value: any): value is RuleClass {
+    if (!value) return false
     if (typeof value !== 'function') return false
     if (!value.prototype) return false
 
     return typeof value.ruleName === 'string' && typeof value.prototype.check === 'function'
+  }
+
+  /**
+   * Check for usage of deprecated 'name' property instead of 'static ruleName'
+   */
+  private hasDeprecatedNameProperty(value: any): boolean {
+    if (!value) return false
+    if (typeof value !== 'function') return false
+
+    const instance = new value()
+    return Object.prototype.hasOwnProperty.call(instance, 'name')
   }
 
   /**
--- a/javascript/packages/linter/test/cli.test.ts
+++ b/javascript/packages/linter/test/cli.test.ts
@@ -685,5 +685,40 @@ describe("CLI Output Formatting", () => {
         }
       }
     })
+
+    test("exits with an error when a custom rule uses the deprecated 'name' instance property", () => {
+      try {
+        mkdirSync(join(tempDir, ".herb/rules"), { recursive: true })
+        mkdirSync(join(tempDir, "app/views"), { recursive: true })
+
+        writeFileSync(join(tempDir, ".herb.yml"), dedent`
+          version: 0.8.10
+          linter:
+            enabled: true
+        `)
+
+        writeFileSync(join(tempDir, ".herb/rules/deprecated-rule.mjs"), dedent`
+          export default class DeprecatedRule {
+            name = "deprecated-rule"
+
+            check(document, context) {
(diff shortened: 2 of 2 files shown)


javascript/packages/linter/src/cli/file-processor.ts
```diff
@@ -51,18 +51,20 @@ export class FileProcessor {
   private isRuleAutocorrectable(ruleName: string): boolean {
     if (!this.linter) return false
 
-    const RuleClass = (this.linter as any).rules.find((rule: any) => {
-      const instance = new rule()
+    const ruleClass = (this.linter as any).rules.find(
```
citizen428, line 54: On top of changing the `find` I also renamed `RuleClass` to `ruleClass`. I think it's confusing to name a variable referencing a class in camel-case like the constructor function.

javascript/packages/linter/src/cli/file-processor.ts
```diff
@@ -51,18 +51,20 @@ export class FileProcessor {
...
+    if (!ruleClass) return false
 
-    if (!RuleClass) return false
-
-    return RuleClass.autocorrectable === true
+    return ruleClass.autocorrectable === true
   }
 
-  async processFiles(files: string[], formatOption: FormatOption = 'detailed', context?: ProcessingContext): Promise<ProcessingResult> {
+  async processFiles(
```
citizen428: I didn't catch that my editor auto-formatted this. Let me know if you want me to roll this back. Same for other format changes.

marcoroth: If you can easily revert it, then that would be great! On that note, I guess we should also enforce formatting for JavaScript/TypeScript files.

javascript/packages/linter/src/types.ts
```diff
@@ -87,11 +87,15 @@ export const DEFAULT_RULE_CONFIG: FullRuleConfig = {
...
 export abstract class ParserRule<TAutofixContext extends BaseAutofixContext = BaseAutofixContext> {
   static type = "parser" as const
+  static ruleName: string
   /** Indicates whether this rule supports autofix. Defaults to false. */
   static autocorrectable = false
   /** Indicates whether this rule supports unsafe autofix (requires --fix-unsafely). Defaults to false. */
   static unsafeAutocorrectable = false
-  abstract name: string
+
+  get ruleName(): string {
```
citizen428, line 96: @marcoroth Having an instance-level getter seemed sensible but if you don't want that we can remove them.

citizen428:
@marcoroth OK, I believe this is ready now. I know you also mentioned potentially doing something similar for `type`, but I think this PR is big enough now.

javascript/packages/linter/src/types.ts
marcoroth: Yeah nice, this is great! Thanks! 🥳  This is way better than having to reference `MyRuleClassName.ruleName`  when passing it to the visitor.

marcoroth:
I think we also need to update the docs: https://herb-tools.dev/projects/linter#custom-rules

Also wondering if we can easily provide a deprecation message if we see somebody is trying to define `name`. I know some people are using custom linter rules already.

citizen428:
> I think we also need to update the docs: [herb-tools.dev/projects/linter#custom-rules](https://herb-tools.dev/projects/linter#custom-rules)

Done in [153de47](https://github.com/marcoroth/herb/pull/1268/commits/153de47f7bdebfbad642ad1e1ac4c6bdf63536e5), please check.

> Also wondering if we can easily provide a deprecation message if we see somebody is trying to define name. I know some people are using custom linter rules already.

I pushed an approach for that in [b5c5020](https://github.com/marcoroth/herb/pull/1268/commits/b5c5020277b8c6fc10ba87419ce0a383b1694144), let me know what you think. I opted for an error and termination to force people into updating their custom rules,. but maybe that's too aggressive.

javascript/packages/linter/src/custom-rule-loader.ts
```diff
@@ -75,10 +75,16 @@ export class CustomRuleLoader {
       const cacheBustedUrl = `${fileUrl}?t=${Date.now()}`
       const module = await import(cacheBustedUrl)
 
-      if (module.default && this.isValidRuleClass(module.default)) {
+      if (this.isValidRuleClass(module.default)) {
```
citizen428, line 78: I felt this `module.default &&` is a bit superfluous, I moved it into `isValidRuleClass` as a `if (!value)` check and then followed the same pattern in `hasDeprecatedNameProperty`.

javascript/packages/linter/src/custom-rule-loader.ts
```diff
@@ -75,10 +75,16 @@ export class CustomRuleLoader {
...
         return [module.default]
       }
 
+      if (this.hasDeprecatedNameProperty(module.default)) {
```
citizen428, line 82: So when we get here, we already know that we're not dealing with a valid rule class (e.g. not satisfying `typeof value.ruleName === 'string'`). So here we check if we should output the more specific deprecation message.

marcoroth: Perfect, this is great, thanks!

PR #1295, opened 2026-03-04
Formatter: Preserve user newlines for block elements with ERB children

### LINKED ISSUE #1181: Formatter: Leave children tags on their own lines
Right now, ERB tags will be formatted such that the whole available line length is used greedily. This is in contrast to other tags, e.g. the `<p>` tag, which stays as is.

If this is intended behavior, it'd be awesome if there was a formatter "rule" (or some config) that allows to customize it. In my opinion, the example is more readable if not the full line length is used. While setting max line length to 80, I set it to 100 on purpose since from my experience, _80 soft / 100 hard_ works quite well. _80 hard_ usually has some annoying edge cases, that's why I usually go for _100 hard_ but try to only use up to 80 chars where possible.

**Input:**
```erb
<div class="abcdefg abcdefg abcdefg abcdefg">
  <% t("registration.policies.edit.title") %>
</div>

<div class="abcdefg abcdefg abcdefg abcdefg">
  <p>Test</p>
</div>

```

**Output:**
```erb
<div class="abcdefg abcdefg abcdefg abcdefg"><% t("registration.policies.edit.title") %></div>

<div class="abcdefg abcdefg abcdefg abcdefg">
  <p>Test</p>
</div>

```

**Expected:**

```erb
<div class="abcdefg abcdefg abcdefg abcdefg">
  <% t("registration.policies.edit.title") %>
</div>

<div class="abcdefg abcdefg abcdefg abcdefg">
  <p>Test</p>
</div>

```

**Playground:**

See [this link](https://herb-tools.dev/playground?tab=format&formatterOptions=%7B%22maxLineLength%22%3A100%7D#DwEwlgbgBAxgNgQwM5ILwCIECMYgKYBmA5lNroSWfsaTtUegHwBQUUwApFAC4AU6AJzxEwSbgITcwAewB2AOgAO0uGBhg8SeXnDd5U7nDzoAlFA4tgAenAQWzUJFiIUGKhVrka773QpNWdkVGABVNbmtghxtIRiA) and click on `Format`. Note that I changed the max line length to 100.

https://github.com/user-attachments/assets/5d6d2fba-af32-419a-b5a8-02b08841b5c3

marcoroth:
Previously, the formatter would inline ERB children in block elements when the content fit within `maxLineLength`, even when the user intentionally placed them on separate lines. 

This pull request extends the existing mixed-content newline preservation (#1279) to also cover block elements with ERB children.

Resolves #1181

### DIFF
--- a/javascript/packages/formatter/test/html/inline-elements.test.ts
+++ b/javascript/packages/formatter/test/html/inline-elements.test.ts
@@ -179,10 +179,159 @@ describe("@herb-tools/formatter - inline elements", () => {
     `)
   })
 
-  test("block element with single ERB child on separate line preserves format (issue #1181)", () => {
+  test("block element with ERB output child on separate line preserves format (issue #1181)", () => {
     expectFormattedToMatch(dedent`
       <div class="form-inputs">
         <%= f.input :password, hint: false %>
+      </div>
+    `)
+  })
+
+  test("block element with ERB output child on same line stays inline (issue #1181)", () => {
+    expectFormattedToMatch(dedent`
+      <div class="form-inputs"><%= f.input :password, hint: false %></div>
+    `)
+  })
+
+  test("block element with ERB execution child on separate line preserves format (issue #1181)", () => {
+    expectFormattedToMatch(dedent`
+      <div class="form-inputs">
+        <% f.input :password, hint: false %>
+      </div>
+    `)
+  })
+
+  test("block element with ERB execution child on same line stays inline (issue #1181)", () => {
+    expectFormattedToMatch(dedent`
+      <div class="form-inputs"><% f.input :password, hint: false %></div>
+    `)
+  })
+
+  test("block element with ERB child on separate line preserves format with maxLineLength 100 (issue #1181)", () => {
+    const wideFormatter = new Formatter(Herb, { indentWidth: 2, maxLineLength: 100 })
+
+    const source = dedent`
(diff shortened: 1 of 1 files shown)


javascript/packages/formatter/test/erb-formatter/erb-formatter-compat.test.ts
```diff
@@ -1,9 +1,11 @@
 import { describe, test, expect, beforeAll } from "vitest"
 import { Herb } from "@herb-tools/node-wasm"
 import { Formatter } from "../../src"
+import { createExpectFormattedToMatch } from "../helpers"
 import dedent from "dedent"
 
 let formatter: Formatter
+let expectFormattedTotMatch: ReturnType<typeof createExpectFormattedToMatch>
```
natematykiewicz, line 8: @marcoroth this looks like a typo

```suggestion
let expectFormattedToMatch: ReturnType<typeof createExpectFormattedToMatch>
```

marcoroth: Whoops, thanks for catching that @natematykiewicz! 😅

marcoroth: Fixed in https://github.com/marcoroth/herb/commit/88e8e763f14147c900ac1ffd9194756f765516d2

Issue #1023, opened 2025-12-29
ERB output tags (<%= %>) are not allowed in attribute position


bonkrat:
`herb analyze` is failing on compilation errors with the following error:

` Herb::Engine::SecurityError: ERB output tags (<%= %>) are not allowed in attribute position. - Suggestion: Use control flow (<% %>) with static attributes instead.`

Unfortunately our project uses helpers extensively that trigger this security error. Are there any suggestions here on disabling this security error or a workaround?

marcoroth:
Hey @bonkrat, thanks for reporting!

I think we want to keep these ` Herb::Engine::SecurityError` on by default, but I think it's fair to also allow people to opt out, so that it will ease the adoption and improve backwards compatibility.

I'm imagining that you could put something like this in the `.herb.yml` file:

```yaml
engine:
  security: "error"  # or "warn" or "ignore"
```

jcoyne:
I suspect things like this are fairly commonplace idioms:

```erb
<div class="dark"<%= ' hidden' if index != 0 %>> ...</div>
```

Should this really be a security error?

joelhawksley:
I agree with @jcoyne. In our application, we currently get almost 1800 of these errors.

joelhawksley:
@marcoroth - @issyl0 and I see that this error is _not_ raised by the latest linter (with `parser-no-errors: true`), but is raised when compiling. Which behavior is correct? https://github.com/marcoroth/herb/issues/1251

marcoroth:
@joelhawksley from a parser perspective it's valid syntax and valid structure which is why the parser (nor the linter rule for parser errors) is raising that issue. 

However, I also dislike that this is mismatching between the linter and the engine on the Ruby side. That said, there is an open issue to also implement the same validation check the engine is doing as a linter rule (see #546).

joelhawksley:
@marcoroth that makes sense. Do you still see us being able to disable this error? This is a backwards incompatible change that I think will be a big pain if/when Herb is upstreamed.

marcoroth:
Yes totally! I was also considering having them turned off by default. Not sure that makes sense for most new apps though.

But I definitely want to make them configurable, so that you control the behavior like with the linter rules.

joelhawksley:
@marcoroth would you accept a PR? I would be happy to try writing one to make it configurable.

marcoroth:
@joelhawksley 100%, yes that would be great! 🙏🏼

[...]

PR #1315, opened 2026-03-06
Engine: Add `engine.security` configuration option

joelhawksley:
Fixes: #1023

This PR was written by Claude Opus 4.6, but I reviewed every line ❤️


lib/herb/engine.rb
```diff
@@ -55,6 +55,9 @@ def initialize(input, properties = {})
       @content_for_head = properties[:content_for_head]
       @validation_error_template = nil
       @validation_mode = properties.fetch(:validation_mode, :raise)
+      @security_mode = properties.fetch(:security) {
+        Herb.configuration.dig("engine", "security") || "error"
+      }
```
marcoroth: Since `Herb::Engine.new` is an API accessible from pure-Ruby, can we cast the `@security_mode` to a symbol, so it matches `@validation_mode` etc.


```suggestion
      @security_mode = properties.fetch(:security) {
        Herb.configuration.dig("engine", "security") || :error
      }.to_sym
```

I think it might be fine to keep the `Herb::Configuration` one to continue to return a string

lib/herb/engine.rb
```diff
@@ -72,6 +75,11 @@ def initialize(input, properties = {})
               "validation_mode must be one of :raise, :overlay, or :none, got #{@validation_mode.inspect}"
       end
 
+      unless ["error", "warn", "ignore"].include?(@security_mode)
+        raise ArgumentError,
+              "security must be one of \"error\", \"warn\", or \"ignore\", got #{@security_mode.inspect}"
+      end
```
marcoroth:
```suggestion
      unless [:error, :warn, :ignore].include?(@security_mode)
        raise ArgumentError,
              "security must be one of :error, :warn, or :ignore, got #{@security_mode.inspect}"
      end
```

joelhawksley:
@marcoroth I think this is in good shape now ❤️

joelhawksley:
> If you use Herb via ReActionView in Rails wouldn't you expect to configure this in config/initializers/reactionview.rb?

I think you need both, in case you are using Herb without ReActionView. How the two should interplay is something I would need to give more thought.

PR #1327, opened 2026-03-07
C: Migrate `hb_narray` to use new allocator system

timkaechele:
This MR migrates the `hb_narray` implementation to use the `hb_allocator` interface for allocating memory. 

Related #1323 and #1326

### DIFF
--- a/test/c/test_hb_narray.c
+++ b/test/c/test_hb_narray.c
@@ -3,8 +3,10 @@
 
 TEST(test_hb_narray_init)
   hb_narray_T array;
+  hb_allocator_T allocator;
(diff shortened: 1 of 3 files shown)


test/c/test_hb_narray.c
```diff
@@ -1,10 +1,13 @@
 #include "include/test.h"
 #include "../../src/include/util/hb_narray.h"
+#include <alloca.h>
```
marcoroth: What do we need this for? 🤔

timkaechele: That's a good question, I think that was a wrong include.

Issue #1282, opened 2026-03-03
Language Server: Implement additional LSP features


marcoroth:
The language server currently implements diagnostics, formatting, range formatting, and code actions. There are many more LSP features we could leverage given Herb's rich AST, token stream, linter, and Ruby analysis.

| Feature | LSP Method | Herb Opportunity |
|---|---|---|
| Document Symbols | `textDocument/documentSymbol` | Outline + breadcrumbs for HTML elements, ERB control flow blocks |
| Folding Ranges | `textDocument/foldingRange` | Fold HTML elements, ERB `if`/`each`/`case`/`begin` blocks, comments |
| Semantic Tokens | `textDocument/semanticTokens` | Rich highlighting using lexer tokens (tags, attributes, ERB delimiters, keywords) |
| Document Highlight | `textDocument/documentHighlight` | Highlight matching open/close tag pairs and ERB control flow pairs |
| Linked Editing Range | `textDocument/linkedEditingRange` | Edit `<div>` and `</div>` simultaneously |
| Completions | `textDocument/completion` | HTML tags, attributes, attribute values, ERB snippets, closing tags, rule names |
| Hover | `textDocument/hover` | Element/attribute info, ARIA guidance, ERB tag type, linter rule descriptions |
| Selection Range | `textDocument/selectionRange` | Smart expand selection via AST (content → tag → element → parent) |
| Go to Definition | `textDocument/definition` | Navigate `render` calls to partials/components |
| Document Links | `textDocument/documentLink` | Clickable partial paths and asset `src`/`href` attributes |
| Rename | `textDocument/rename` | Rename tag pairs, `id` attributes, ERB locals |
| Code Lens | `textDocument/codeLens` | Inline offense counts, disabled rule counts |
| On-Type Formatting | `textDocument/onTypeFormatting` | Auto-indent after ERB/HTML opening tags |
| Inlay Hints | `textDocument/inlayHint` | ERB tag type labels, trim indicators |
| Workspace Symbols | `workspace/symbol` | Project-wide search for IDs, classes, partials |
| Document Color | `textDocument/documentColor` | Color swatches for CSS colors in `style` attributes |

PR #1328, opened 2026-03-08
Language Service: Implement Inlay Hints

marcoroth:
This pull request implements inlay hints for the language server, annotating the closing tag of a longer block with what it closes.

image

An ERB `<% end %>` shows the expression that opened the block, and an HTML closing tag names the element the way you would select it, so a `</div>` reads as `<!-- #main.card -->`. Hints are emitted for `if`, `unless`, `case`, `case`/`in`, `while`, `until`, `for`, `begin` and block nodes, and for any element carrying a static `id` or `class`.

Inlay hints are a reading preference rather than something a project should decide, so they live with the editor and not in `.herb.yml`.

VS Code contributes the settings as regular preferences. Editors that keep language server settings under their own key pass the same options nested under `initialization_options`, which the Zed documentation now covers.

Changing a setting refreshes the hints straight away through `workspace/inlayHint/refresh`, guarded by a new `supportsInlayHintRefresh` capability so clients that do not advertise it are left alone.

The inlay hints can be disabled with the `languageServerHerb.inlayHints.enabled` setting, but they can also be turned off in the editor itself for all files, like in VS Code:

image

Related #1282


eirvandelden:
I would love this! 😍

Issue #1251, opened 2026-02-27
Mismatch between `herb compile` and linter


joelhawksley:
I am seeing inconsistent behavior between the linter and the compiler. Shouldn't passing the linter mean that the compiler will succeed?

```
npx @herb-tools/linter@0.8.10 .../show.html.erb
✓ .../show.html.erb - No issues found
```

 Summary:
  Checked      1 file
  Offenses     0 offenses
  Start at     10:27:14
  Duration     148ms (47 rules)

```
bundle exec herb compile ..,/show.html.erb
Unexpected error: Herb::Engine::SecurityError: .../show.html.erb:19:12 - ERB output tags (<%= %>) are not allowed in attribute position. - Suggestion: Use control flow (<% %>) with static attributes instead.
/workspaces/github/vendor/gems/4.0.1/ruby/4.0.0/gems/herb-0.8.10/lib/herb/engine.rb:319:in 'Herb::Engine#handle_validation_errors'
/workspaces/github/vendor/gems/4.0.1/ruby/4.0.0/gems/herb-0.8.10/lib/herb/engine.rb:110:in 'Herb::Engine#initialize'
/workspaces/github/vendor/gems/4.0.1/ruby/4.0.0/gems/herb-0.8.10/lib/herb/cli.rb:271:in 'Herb::CLI#compile_template'
/workspaces/github/vendor/gems/4.0.1/ruby/4.0.0/gems/herb-0.8.10/lib/herb/cli.rb:125:in 'Herb::CLI#result'
/workspaces/github/vendor/gems/4.0.1/ruby/4.0.0/gems/herb-0.8.10/lib/herb/cli.rb:31:in 'Herb::CLI#call'
```

Approximate template:

```html+erb
<tr class="tolor-fg-muted" <%= test_selector("org") %>>
```

Issue #1058, opened 2026-01-13
SecurityValidator differs from erb-no-output-control-flow


Jcambass:
I'm aware that the CLI does not use the `.herb` config yet. However, with code like

```erb
style=<%= 'display:none' unless xyz %>
```

I get no lint errors even with `erb-no-output-control-flow` enabled. While `herb analyze` gives me  

```
RB output tags (<%= %>) are not allowed in attribute position. - Suggestion: Use control flow (<% %>) with static attributes instead.
```

I also get this error which isn't present in the lint output itself.

```
ERB output in attribute names is not allowed for security reasons. - Suggestion: Use static attribute names with dynamic values instead.
```

for code like

```erb
<button class="btn"
  <%= x ? "" : "disabled" %>
  type="submit">
    Save and Continue
</button>
```

marcoroth:
Thanks for reporting this one, @Jcambass!

Sadly this is one of the inconsequences of writing the linter in JavaScript and the rendering engine itself in Ruby. 

Ideally, we could also have linter rules for the runtime errors the engine produces when trying to compile a template, as we are technically also able to figure that out at static analysis time.

Issue #546, opened 2025-09-30
erb-lint: ErbSafety


marcoroth:
https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#erbsafety

https://github.com/Shopify/erb_lint/blob/main/lib/erb_lint/linters/erb_safety.rb
https://github.com/Shopify/erb_lint/blob/main/spec/erb_lint/linters/erb_safety_spec.rb

## Linter Rules

Based on [Shopify/better-html](https://github.com/Shopify/better-html) safety checks:

- [x] `erb-no-output-in-attribute-position` — Disallow ERB output in attribute position (formerly `erb-safety`)
- [x] `erb-no-output-in-attribute-name` — Disallow ERB output in attribute names (formerly `erb-safety`)
- [x] `erb-no-unsafe-raw` — Disallow `raw()` and `.html_safe` in ERB output
- [x] `erb-no-raw-output-in-attribute-value` — Disallow `<%==` in attribute values
- [x] `erb-no-unsafe-js-attribute` — Disallow unsafe ERB output in JavaScript attributes (require `.to_json`, `j()`, or `escape_javascript()`)
- [x] `erb-no-unsafe-script-interpolation` — Disallow unsafe ERB output inside `<script>` tags (require `.to_json`)
- [x] `erb-no-statement-in-script` — Disallow ERB statements inside `<script>` tags (likely meant `<%= %>`)
- [x] `erb-no-javascript-tag-helper` — Disallow deprecated `javascript_tag do` helper
- [x] `html-allowed-script-type` — Restrict allowed `type` attributes for `<script>` tags (already existed)

markokajzer:
What's the expectation here?

If I look at https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#erbsafety, it says
> Runs the checks provided by [better-html's erb safety test helper](https://github.com/Shopify/better-html#testing-for-valid-html-and-erb).

If I check [the `better-html` repo](https://github.com/Shopify/better-html), it actually points back to `herb` now lol

> This project is no longer actively maintained. **It has been superseded by the tools and guidance available at https://herb-tools.dev/.**

so... noop?

marcoroth:
I think we want to provide the same checks we have in the `Herb::Engine` as linter rule checks, so that the errors in the linter match with the errors the engine reports at runtime: See: https://github.com/marcoroth/herb/tree/main/lib/herb/engine/validators

Related: #1058

PR #1330, opened 2026-03-08
Linter: Implement `erb-safety` family of linter rules

marcoroth:
This pull request implements the full set of ERB safety linter rules, covering all checks from [better-html](https://github.com/Shopify/better-html) `SafeErb` test helpers and [erb_lint](https://github.com/Shopify/erb_lint) `ErbSafety` linter.

This now also brings the linter in-sync with the `SecurityValidator` the `Herb::Engine` has been enforcing.

Resolves https://github.com/marcoroth/herb/issues/546
Resolves https://github.com/marcoroth/herb/issues/1058
Resolves https://github.com/marcoroth/herb/issues/1251

#### New Rules

| Rule                                   | Description                                                                                        |
|----------------------------------------|----------------------------------------------------------------------------------------------------|
| `erb-no-statement-in-script`           | Avoid `<% %>` tags inside `<script>`. Use `<%= %>` to interpolate values into JavaScript.          |
| `erb-no-javascript-tag-helper`         | Avoid `javascript_tag`. Use inline `<script>` tags instead.                                        |
| `erb-no-unsafe-script-interpolation`   | ERB output in `<script>` tags must use `.to_json` to safely serialize values.                      |
| `erb-no-raw-output-in-attribute-value` | Avoid `<%==` in attribute values. Use `<%= %>` instead.                                            |
| `erb-no-unsafe-raw`                    | Avoid `raw()` and `.html_safe` in ERB output (skips raw-text elements like `<script>`, `<style>`). |
| `erb-no-unsafe-js-attribute`           | ERB output in `on*` attributes must use `.to_json`, `j()`, or `escape_javascript()`.               |
| `erb-no-output-in-attribute-position`  | Avoid `<%= %>` in attribute position. Use `<% if ... %>` with static attributes instead.           |
| `erb-no-output-in-attribute-name`      | Avoid ERB output in attribute names. Use static attribute names with dynamic values instead.       |

#### Mapping to better-html / erb_lint

| better-html checker | Our rule(s) |
|---|---|
| `AllowedScriptType` | `html-allowed-script-type` (pre-existing) |
| `NoStatements` | `erb-no-statement-in-script` |
| `NoJavascriptTagHelper` | `erb-no-javascript-tag-helper` |
| `ScriptInterpolation` | `erb-no-unsafe-script-interpolation` |
| `TagInterpolation` | `erb-no-raw-output-in-attribute-value`, `erb-no-unsafe-raw`, `erb-no-unsafe-js-attribute`, `erb-no-output-in-attribute-position`,   `erb-no-output-in-attribute-name` |


frederikspang:
Late to the party here. For `erb-no-unsafe-script-interpolation` - should we not allow for `j()` helpers in script tag, such as:

```
const url = '<%= j @poll_path %>';
```

Not sure if I'm overlooking something here.

marcoroth:
Hey @frederikspang, thats a good point, thanks for mentioning!

I think your example is technically fine, but the following wouldn't, since it's not quoted:
```html+erb
const url = <%= j @poll_path %>
```

I think without parsing the JavaScript we don't really have a good understanding of the surrounding JavaScript. I wonder if there is a different way around this 🤔

marcoroth:
Maybe we should detect the usage of `j` and `escape_javascript`, and then specifically encourage to use `.to_json` instead, as this is going to be valid and safe in any position:

```diff
- const url = '<%= j @poll_path %>';
+ const url = <%= @poll_path.to_json %>;
```

PR #1363, opened 2026-03-11
Engine: Inline comments on `<% end %>` on output blocks produce invalid Ruby

joelhawksley:
## Version

herb 0.8.10 (commit `3a36f815b05b`)

## Summary

When `<% end # comment %>` closes a block opened inside a `<%= %>` output tag, Herb appends the closing `))` *after* the comment text in compiled Ruby. Since `#` starts a Ruby comment, the `))` become part of the comment, leaving `__herb.h((` unclosed.

## Minimal reproduction

```erb
<%= render Foo.new do %>
  hello
<% end # comment %>
```

### Actual compiled Ruby (broken)

```ruby
_buf << __herb.h((render Foo.new do; _buf << '  hello
'.freeze; end # comment ))
```

The `))` after `# comment` is commented out — parens never close.

### Expected

The `))` should be placed *before* the comment, or the comment should be stripped:

```ruby
_buf << __herb.h((render Foo.new do; _buf << '  hello
'.freeze; end )) # comment
```

### Prism parse errors

```
- unexpected end-of-input, assuming it is closing the parent top level context (line 4)
- expected a matching `)` (line 5)
- unexpected end-of-input; expected a `)` to close the arguments (line 4)
```

## Scope

- **Affected:** `<% end # ... %>` closing a block started in `<%= %>` (where Herb wraps in `__herb.h((...))`)
- **Not affected:** `<% end # ... %>` inside non-output `<% %>` tags — no wrapping parens, so comment is harmless

## Workaround

Remove inline comments from `end` tags that close output-tag blocks.

### DIFF
--- a/test/snapshots/engine/engine_block_test/test_0014_inline_comment_on_end_inside_output_block_does_not_break_parens_1c140392798ee0e334ac05bafcebe13b.txt
+++ b/test/snapshots/engine/engine_block_test/test_0014_inline_comment_on_end_inside_output_block_does_not_break_parens_1c140392798ee0e334ac05bafcebe13b.txt
@@ -2,6 +2,6 @@
 source: "Engine::EngineBlockTest#test_0014_inline comment on end inside output block does not break parens"
 input: "{source: \"<%= render Foo.new do %>hello<% end # comment %>\", options: {}}"
 ---
-_buf = ::String.new; _buf << (render Foo.new do; _buf << 'hello'.freeze; end)
-# comment
+_buf = ::String.new; _buf << (render Foo.new do; _buf << 'hello'.freeze; end # comment
+);
 _buf.to_s


marcoroth:
I think we can keep this pragmatic and just forcefully insert a newline if we see a `#`.

lib/herb/engine.rb
```diff
@@ -286,14 +286,25 @@ def add_expression_block_end(code, escaped: false)
...
+      else
+        @src << " " << code_stripped
+        @src << closing
+
+        @src << if trailing_newline
+                  "\n"
+                else
+                  ";"
+                end
+      end
```
marcoroth: I appreciate the attempt to keep the compiled output pretty ❤️

But I think we could keep this super pragmatic and just insert the newline directly after the code:

```suggestion
      @src << " " << code_stripped
      @src << "\n" if self.class.comment?(code_stripped)
      @src << (escaped ? "))" : ")")
      @src << (trailing_newline ? "\n" : ";")
```

PR #1379, opened 2026-03-14
C: Use 128 KB pages by default for arenas

timkaechele:
This PR increases the default arena page size to 128 KB, which is a tradeoff between reducing page allocations and not requiring too much memory per parse run.

Additionally it starts using the already existing macros to define the default page size.

It also keeps the WASM builds on 16 KB pages, since the tradeoff runs the other way there.

Measured against [herb-corpus](https://github.com/marcoroth/herb-corpus): roughly 3% lower RSS across the board, and far fewer page allocations, with a large view going from 146 pages down to 18. Parse time is unchanged, so this is a memory change rather than a speedup. The extra mapped-but-unused bytes are not a real cost on native builds, because anonymous `mmap` commits lazily.

WASM is the exception. Emscripten has no virtual memory, so a mapping there is allocated and zeroed for real and the unused tail gets paid for twice. Since most templates fit in a single 128 KB page, that would be a clear regression. `HB_ALLOCATOR_DEFAULT_ARENA_SIZE` is already `#ifndef`-guarded, so `wasm/Makefile` just defines it back to 16 KB.


marcoroth (approved):
Updated the pull request and description to keep the WASM build on 16 KB, thanks @timkaechele! 🙏🏼

PR #1381, opened 2026-03-14
C: Use more sensible start capacity for `hb_array`

timkaechele:
This PR changes the start capacity of an array that previously had no capacity.

Instead of using 1, which only allocates a single slot, we allocate 8, which allows more room to grow. 

Given that the `hb_array` only holds pointers to its items, the memory usage is negligible.

### DIFF
--- a/src/util/hb_array.c
+++ b/src/util/hb_array.c
@@ -39,7 +39,7 @@ bool hb_array_append(hb_array_T* array, void* item) {
     size_t new_capacity;
 
     if (array->capacity == 0) {
-      new_capacity = 1;
+      new_capacity = 8;
     } else if (array->capacity > SIZE_MAX / (2 * sizeof(void*))) {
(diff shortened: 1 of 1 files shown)


src/util/hb_array.c
```diff
@@ -39,7 +39,7 @@ bool hb_array_append(hb_array_T* array, void* item) {
     size_t new_capacity;
 
     if (array->capacity == 0) {
-      new_capacity = 1;
+      new_capacity = 8;
```
marcoroth, line 42: I'm wondering if this is really the right approach. Ideally we could have a lazy and/or shared empty array that only uses memory once for all instances. And only when we try to append errors, we convert it to a proper full array with a capacity. Otherwise on the happy path we would allocate a lot of memory that is never going to be used.

marcoroth:
Thanks @timkaechele 🙏🏼 

I'm going to close this in favor of https://github.com/marcoroth/herb/pull/1425

PR #1384, opened 2026-03-15
Linter: Implement `html-require-script-nonce` rule

### LINKED ISSUE #543: erb-lint: RequireScriptNonce
https://github.com/Shopify/erb_lint/tree/main?tab=readme-ov-file#requirescriptnonce

https://github.com/Shopify/erb_lint/blob/main/lib/erb_lint/linters/require_script_nonce.rb
https://github.com/Shopify/erb_lint/blob/main/spec/erb_lint/linters/require_script_nonce_spec.rb

markokajzer:
closes #543

handles three things
1. html: `<script>` tags
2. erb: Rails javascript helpers, i.e. `javascript_tag`, `javascript_include_tag`, ~and `javascript_pack_tag`~
3. erb: Rails tag helpers, i.e. `tag.script`

uses #1374 to handle both `<script>` and `javascript_tag` helpers via `BaseRuleVisitor#visitHTMLElementNode`.

> [!NOTE]
> Does not handle `nil` values, e.g. `tag.script nonce: nil` based on the specs of the inspiration. If wanted, we can add more specs and handlers.


marcoroth:
Just merged https://github.com/marcoroth/herb/pull/1374 🙌🏼

markokajzer:
Thank you for your tireless work @marcoroth 

I'll rebase, and give it another shot! 👍

javascript/packages/linter/src/rules/html-require-script-nonce.ts
```diff
@@ -0,0 +1,72 @@
...
+    return getAttribute(node, name)
+  }
+}
+
+export class HTMLRequireScriptNonceRule extends ParserRule {
+  static ruleName = "html-require-script-nonce"
+
+  get defaultConfig(): FullRuleConfig {
+    return {
+      enabled: true,
```
markokajzer, line 54: just noticed it's not enabled by default in erb_lint. do we want the same here?

javascript/packages/linter/src/rules/html-require-script-nonce.ts
```diff
@@ -0,0 +1,72 @@
...
+      severity: "error"
```
markokajzer, line 55: what's the correct severity?

markokajzer:
@marcoroth I think I got it now! Given that #1374 did not include parsing of `javascript_pack_tag`, I removed the handling of it here as well, so that we can simplify and only rely on `visitHTMLElementNode`. (I think it's also deprecated / removed in Rails as well nowadays?)

Let me know if any other changes are necessary, I wasn't sure about severity and whether we want to enable by default! 🙏

javascript/packages/linter/src/rules/html-require-script-nonce.ts
marcoroth: I think its fine to keep this one enabled for now 🙌🏼

javascript/packages/linter/src/rules/html-require-script-nonce.ts
marcoroth: We still haven't really balanced all the severities yet. So I think we can keep this one as `error` for now, even though it might be a bit harsh/aggressive.

marcoroth:
Implemented a small follow up on https://github.com/marcoroth/herb/pull/1452 🙌🏼

PR #1388, opened 2026-03-16
C: Skip unnecessary `token_copy()` in AST node constructors

### LINKED ISSUE #1339: C: Skip unnecessary `token_copy()` in AST node constructors
Since we are using `hb_string_T` now we might be able to drop the `token_copy()` here:

https://github.com/marcoroth/herb/blob/000dbff52b9155b17fe6656b1374b8d25b259811/templates/src/ast_nodes.c.erb#L30

citizen428:
Closes #1339 

@marcoroth I followed the `node`/`borrowed_node` code for this. But I'm not entirely sure if this is what you wanted because we only seem to use `borrowed_node` for `HTMLConditionalElementNode` (there are 2 occurrences of `borrowed_node` and 50 of `node`).


citizen428:
@marcoroth Should we just close this? It was small when I originally opened it but things happened and now that I had time to look at it again, it's a bit of a mess 😅

PR #1409, opened 2026-03-17
VS Code: Don't register Toggle Comment as global shortcut

### LINKED ISSUE #1405: Cursor: Commenting broken in last version of HERB.
Somehow, I think the commenting feature "gets broken" after some time with the latest herb version.
Mihai had an issue inside a Ruby file [yesterday](https://cleanshot.com/share/gvXfSwRK) and I have an issue today inside CSS (which is weird).

I don't know how to debug it further, but happy to send more guided info or jump on a call if need be.

Thanks,

Image

marcoroth:
This pull request fixes the Herb extension from globally overriding VS Code's built-in comment commands (`editor.action.commentLine` and `editor.action.blockComment`). 

Previously, the extension hijacked these commands for all file types and tried to fall back to the default behavior via `default:editor.action.commentLine` for non-ERB files, but that fallback mechanism isn't reliable. Instead we register our comment actions and assign them with a shortcut in ERB files.

Resolves #1405

### DIFF
--- a/javascript/packages/vscode/package.json
+++ b/javascript/packages/vscode/package.json
@@ -140,6 +140,16 @@
         "when": "false"
       },
       {
+        "command": "herb.toggleLineComment",
+        "title": "Herb: Toggle Line Comment",
+        "when": "editorLangId == erb"
+      },
+      {
+        "command": "herb.toggleBlockComment",
+        "title": "Herb: Toggle Block Comment",
+        "when": "editorLangId == erb"
+      },
+      {
         "command": "herb.toggleLinter",
         "title": "Herb: Toggle Linter",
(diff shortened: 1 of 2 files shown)


adrianthedev:
You are the best!
I bet you need to cut a release in order to get the fix, right?

marcoroth:
Yeah, I got one more fix I want do and then cut a 0.9.1 release shortly 🙌🏼

marcoroth:
Should be available as v0.9.1 on both the VS Code and Cursor extension marketplaces 🙌🏼

adrianthedev:
Already on it.
I appreciate you fixing it so fast 🫶

Issue #1204, opened 2026-02-18
Linter: Support per-rule custom configuration options


marcoroth:
Allow linter rules to define custom configuration options beyond `enabled`, `severity`, `include`, `exclude`, and `only`.

### Use case

Rules like `html-allowed-script-type` need configurable options:
- `allowedTypes`: `["text/javascript", "application/json"]`
- `allowBlank`: `true`/`false`

### Example config

```yaml
linter:
  rules:
    html-allowed-script-type:
      enabled: true
      allowedTypes:
        - text/javascript
        - application/json
      allowBlank: true
```

### Related

- #1201
- #554

marcoroth:
We should add a list of side effect methods for ` erb-no-unused-expressions` (see https://github.com/marcoroth/herb/pull/1586)

chrise86:
This would be great. I'd facing issues with `erb-no-unused-expressions` and the `breadcrumb` helper from https://github.com/kzkn/gretel

drjayvee:
I just ran into another specific use case for customization:

```ruby
class ApplicationHelper
  def test_id(id)
    tag.attributes("data-test-id": value) if Rails.env.test?
  end
end
```

```erb
<div <%= test_id :thing %>>
```

This currently creates
```
[error] Avoid <%= %> in attribute position. Use <% if ... %> with static attributes instead. (erb-no-output-in-attribute-position)
```

Currently, the Herb-compliant solutions are awkward workarounds:
```erb
<div <%= test_id :thing %>> <%# herb:disable erb-no-output-in-attribute-position %>

<div <% if Rails.env.test? %>data-test-id="thing"<% end %>>

<div <%= tag.attributes(**test_id_attrs(:thing)) %>>
```

# Proposal
Therefore, it would be great if particular helpers could be allow-listed for `erb-no-output-in-attribute-position`.

The same could be used for `erb-no-unused-expressions`.

We currently have to
```erb
<% set_meta_tags noindex: true %> <%# herb:disable erb-no-unused-expressions %>
```

Issue #1359, opened 2026-03-11
Linter: Rename Action View specific Linter Rules


marcoroth:
With #1358, we added a first `actionview-*` linter rule. We should go through the existing linter rules and make sure that all Action View specific rules use the `actionview-*` prefix. 

We also want to include redirects for the docs page so that the old links are pointing to the new renamed rule.

The main motivation here is, that you can disable all `actionview-*` linter rules with one config in the `.herb.yml` when you aren't using Herb with Action View/Rails.

Related https://github.com/marcoroth/herb/issues/480
Related #1310

Issue #1310, opened 2026-03-06
Linter: Consider renaming all accessibility-related rule names to `a11y-*`


marcoroth:
Currently, most accessibility-related rules have the prefix `html-*` as we only have been really differentiating between `html-*`, `erb-*` and `svg-*`.

Maybe it would make sense to categorize/namespace the accessibility-related rules have the `a11y-*` prefix to make it more obvious.

We also have been slowly introducing `herb-*` and `turbo-*` as prefixes, and I think some rules could also benefit from other namespaces like `actionview-*`, `actionmailer-*`, `stimulus-*` etc.

Related: #1066

written-fresh:
Wonder if it would be worth putting a doc together outlining all of these?

marcoroth:
@written-fresh yeah, I think we need to do a pass to go over them all at once. Since we also need to handle the redirection logic, and making sure if people disabled the rule with their old name that they have a migration path.

Though, for new rules we can already start using the new `a11y-*` prefix so we don't have to rename them later.

PR #1444, opened 2026-03-21
Linter: Implement `source-indentation` rule

### LINKED ISSUE #548: SpaceIndentation

markokajzer:
closes #548

I added an autofix in a separate commit because I wasn't sure about the implementation. The formatter provides an `indentWidth` config option, and optimally we would want to use the same configured value without having to specify a separate value for each rule that needs it.

To do this, I added some plumbing in the linter, to pipe the config from the formatter to all rules. If this is not what we want, we can simply drop/undo the specific commit.

another option is to wait for custom per-rule configuration options, see #1204.

we could also hardcode the configuration, e.g. to 2m, until then, similar to what we did with `html-allowed-script-type`, see #1201.


marcoroth (approved):
Thank you @markokajzer (and sorry for pushing to your branch again) 🙈 

I just merged in main and updated the rule to work with the changes from https://github.com/marcoroth/herb/pull/1453.

Also renamed it to `source-*` to make sure we already align them (similar to https://github.com/marcoroth/herb/issues/1359 and https://github.com/marcoroth/herb/issues/1310), so we don't have to rename them later 🙌🏼

marcoroth:
Hmm, I wonder if we should call it just `source-indentation` so that we could also support something like `style: "tabs"` at a later point 🤔

Issue #1205, opened 2026-02-18
`erblint-github` compatibility/parity


marcoroth:
The Herb Linter could become a viable alternative for the existing [`erblint-github`](https://github.com/github/erblint-github) gem which provides accessibility-focused linting rules.

This issue tracks rule parity with `erblint-github`.

### Accessibility Rules

- [x] [`aria_label_is_well_formatted`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/aria_label_is_well_formatted.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/aria-label-is-well-formatted.md)
- [x] [`avoid_both_disabled_and_aria_disabled`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/avoid_both_disabled_and_aria_disabled.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/avoid-both-disabled-and-aria-disabled.md)
- [x] [`iframe_has_title`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/iframe_has_title.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/iframe-has-title.md)
- [x] [`navigation_has_label`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/navigation_has_label.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/navigation-has-label.md)
- [x] [`no_aria_hidden_on_focusable`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/no_aria_hidden_on_focusable.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-aria-hidden-on-focusable.md)
- [x] [`no_positive_tab_index`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/no_positive_tab_index.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-positive-tab-index.md)
- [x] [`no_title_attribute`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/no_title_attribute.rb) — [docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-title-attribute.md)

### Related

[...]

PR #1455, opened 2026-03-22
docs: add `erblint-github` to inspiration section

markokajzer:
some rules are already in, see #414, and we're aiming for full compatibility, see #1205, so worthing pointing that out :^)


marcoroth (approved):
Great catch, thanks! 🙏🏼

Issue #177, opened 2025-06-26
Linter Rule: No autofocus attribute


NotGrm:
##### Description

Prevent usage of the `autofocus` attribute

##### Rationale

The autofocus attribute is a global attribute that indicates an element should be focused on page load. Autofocus reduces accessibility by moving users to an element without warning and context. Its use should be limited to form fields that serve as the main purpose of the page, such as the search input on the Google home page.

##### ✅ Good

```html+erb
<input type="text" />
```

##### 🚫 Bad

```html+erb
<input type="text" autofocus />
```

```html+erb
<input type="password" autofocus="autofocus" />
```

#### References

Inspired by [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-autofocus-attribute.md)

brunoprietog:
Well, I don't quite agree with this rule. You can and should use autofocus on a popover that has focusable controls, such as a menu. It's also good practice in SPAs to bring the first h1 title to the user when navigating, since screen readers mostly don't report a navigation only when you replace the body element.

I personally use this attribute quite a lot exactly to improve accessibility.

natematykiewicz:
I use it quite a bit to improve accessibility. For example, our OTP page (whose entire purpose is to enter a 6-digit number) autofocuses you to the OTP textbox, so screen users don't have to fumble tabbing through all sorts of UI just to find the textbox.

PR #1462, opened 2026-03-22
Linter: group a11y-related rules

### LINKED ISSUE #1310: Linter: Consider renaming all accessibility-related rule names to `a11y-*`

markokajzer:
closes #1310

Group a11y-related rules together and change their prefix from `html-*` to `a11y-*`

I was looking at adding more a11y linter rules, so decided to do this cleanup first, so that they're all grouped together.

There might be some additional things that we need to do, as existing config might not work anymore for these rules. technically this is _somewhat of a breaking change_. what would be the preferred migration strategy?
+ codemod for existing configs
+ reading previous config keys (e.g. `html-input-require-autocomplete`)
+ skip?
+ ...?

---

somewhat related question: i noticed that most of these only operate on `html`. however, with the introduction of #1350, their scope could be widened to include ERB tags as well.

is that something we would want to look into before introducing new rules? is that something we would want to pay attention to when introducing more a11y rules, like in #177?


marcoroth:
Thanks @markokajzer!

I think I want to wait with this a little bit before we do a big rename of the linter rule names.

But I think for the migration story, we want to utilize the `--upgrade` command we introduced in #1453:

CleanShot 2026-03-22 at 23 07 27@2x

So that running `herb-lint --upgrade` will go through all files, renames any `<%# herb:disable  %>` comments, and updates the `.herb.yml` from the old to the new rule names.

markokajzer:
i'm happy to close this for now then 👍 

i'll start adding new related rules under `ally-*` already, like `ally-no-autofocus-attribute` in #177 👍 

thank you mate! 🙏

PR #1487, opened 2026-03-24
Linter: Add `a11y-no-autofocus-attribute` rule

### LINKED ISSUE #177: Linter Rule: No autofocus attribute

markokajzer:
closes #177

# Linter Rule: No `autofocus` attribute

**Rule:** `a11y-no-autofocus-attribute`

## Description

Prevent usage of the `autofocus` attribute on HTML elements, Action View Form Tag Helpers and Action View Form Builder methods.

## Rationale

The `autofocus` attribute is a global attribute that indicates an element should be focused on page load. It reduces accessibility by moving users to an element without warning and context.

Its use should be limited to form fields that serve as the main purpose of the page, such as a search input on a search page.

### DIFF
--- a/javascript/packages/linter/test/__snapshots__/cli.test.ts.snap
+++ b/javascript/packages/linter/test/__snapshots__/cli.test.ts.snap
@@ -310,7 +310,7 @@ test/fixtures/parser-errors.html.erb:2:16
  Summary:
   Checked      1 file
   Offenses     1 error | 0 warnings (1 offense across 1 file)
-  Fixable      0 offense
+  Fixable      0 offenses
   Rules        72 enabled | 8 not enabled | 5 skipped (version)
 
  TIP: Run herb-lint --init to create a .herb.yml and lock the version.


javascript/packages/linter/src/rules/a11y-no-autofocus-attribute.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+export class A11yNoAutofocusAttributeRule extends ParserRule {
+  static ruleName = "a11y-no-autofocus-attribute"
+  static introducedIn = this.version("unreleased")
+
+  get defaultConfig(): FullRuleConfig {
+    return {
+      enabled: true,
```
marcoroth: I think we should disable this rule by default, given the legitimate concerns brought up in #177 🙌🏼 


```suggestion
      enabled: false,
```

People can still explicitly enable the rule if they feel like it makes sense for their app

markokajzer:
some issues downloading vscode for the `vscode:test` step right now... unlucky :'(

javascript/packages/linter/src/rules/a11y-no-autofocus-attribute.ts
```diff
@@ -0,0 +1,107 @@
+import { hasAttribute, PrismVisitor } from "@herb-tools/core"
+
+import { BaseRuleVisitor } from "./rule-utils.js"
+import { ParserRule } from "../types.js"
+
+import type { HTMLOpenTagNode, ParseResult, ERBContentNode, ParserOptions, PrismNode } from "@herb-tools/core"
+import type { UnboundLintOffense, LintContext, FullRuleConfig } from "../types.js"
+
+const FORM_TAG_HELPERS = new Set([
```
marcoroth, line 10:
```suggestion
// TODO: migrate to `action_view_helpers: true` once we support them all
const FORM_TAG_HELPERS = new Set([
```

marcoroth:
Thank you @markokajzer! 🙏🏼

PR #1492, opened 2026-03-24
Engine: Strip whitespace between consecutive end tags after expression blocks

joelhawksley:
## Fix: Strip whitespace between consecutive `<% end %>` tags after expression blocks

_Co-written with Claude Opus 4.6_ - I've reviewed the fix and generally think it's right, but there might be a simpler solution? 

### Problem

Herb emits `@output_buffer.safe_append=` for whitespace between consecutive `<% end %>` tags that Erubi (with `trim: true`) strips. This breaks `CaptureHelper#capture` in Rails 8.2 for helpers like `form_tag` that return HTML via the block's return value rather than calling `concat`.

When Herb adds `@output_buffer.safe_append='  '` between `end` statements:

1. The block's Ruby return value changes from the form HTML to `@raw_buffer` (the return of `safe_append=`)
2. The capture buffer gets whitespace (e.g. `"  "`)
3. `buffer.presence` returns `nil` (whitespace is `.blank?`)
4. The `value` fallback is just `@raw_buffer` (which only contains whitespace)
5. The form HTML is completely lost

**Triggering template:**

```erb
<% outer do %>
  <% form_tag("/t") do %>
    <%= tag.p do %>
      text
    <% end %>
  <% end %>
<% end %>
```

Herb compiled (before fix):

```ruby
    end )
 _buf << '  '.freeze; end; _buf << '\n'.freeze; end
```

Erubi compiled (and Herb after fix):

```ruby
    end )
   end 
 end
```

### Root Cause

In [preceding_token_ends_with_newline?](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), `:expr_block_end` was grouped with expression tokens in a blanket return false list. This prevented [at_line_start?](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) from recognizing that a `<% end %>` tag following an expression block's closing `end` is at line start — so the subsequent code tag was treated as inline, and the whitespace between them was emitted as buffer appends instead of being trimmed.

### Fix

Remove `:expr_block_end` from the blanket return false list and instead check whether the `expr_block_end` token's value actually ends with `"\n"` (which it does when the end tag was on its own line). This lets [at_line_start?](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) correctly trigger trim behavior for subsequent code-only tags.


lib/herb/engine/compiler.rb
```diff
@@ -10,6 +10,9 @@ def initialize(engine, options = {})
 
         @engine = engine
         @escape = options.fetch(:escape) { options.fetch(:escape_html, false) }
+        @attrfunc = options.fetch(:attrfunc, @escape ? "__herb.attr" : "::Herb::Engine.attr")
```
marcoroth: Were these changes intentional? It looks like this might be from a bad merge.

You probably only need the changes in `preceding_token_ends_with_newline?`, right?

joelhawksley: Yeah, was just looking at cleaning that part up. I'm still working to validate that this fix does what I need it to do. I also have another fix in the works to test alongside it. I'll ping you when things are ready to review!

marcoroth: Sounds good, thank you! 🙏🏼

joelhawksley:
@marcoroth green!

test/engine/whitespace_trimming_test.rb
```diff
@@ -194,5 +194,19 @@ class WhitespaceTrimmingTest < Minitest::Spec
...
+        <% outer do %>
+          <% form_tag("/t") do %>
+            <%= tag.p do %>
+              text
+            <% end %>
+          <% end %>
+        <% end %>
+      ERB
+
+      assert_compiled_snapshot(template)
```
marcoroth, line 209: @joelhawksley I think the fix looks good! I'd love to have an evaluation test here too so we can make sure this also matches the existing Erubi behavior in Action View.

```suggestion
      template = <<~ERB
        <% 1.times do %>
          <% form_tag("/t") do %>
            <%= tag.p do %>
              text
            <% end %>
          <% end %>
        <% end %>
      ERB

      assert_compiled_snapshot(template)
      assert_evaluated_snapshot(template, {},  enforce_actionview_erubi_equality: true)      
```

marcoroth:
Thank you @joelhawksley! 🙏🏼

PR #1504, opened 2026-03-26
Ruby: Introduce `enforce_actionview_erubi_equality` test helper

marcoroth:
This pull request introduces a new `enforce_actionview_erubi_equality` keyword argument for the `assert_evaluated_snapshot` test helper. This option makes sure the `Herb::Engine` evaluated output matches the `Erubi::Engine` evaluated output through ActionView.

The `enforce_actionview_erubi_equality` is automatically enabled when `enforce_erubi_equality` is being passed, so this now runs automatically on all existing test that have `enforce_erubi_equality: true`.

If, for some reason, the output should differ we can opt-out of `enforce_actionview_erubi_equality` by doing the following:
```ruby
assert_evaluated_snapshot(template, {}, enforce_erubi_equality: true, enforce_actionview_erubi_equality: false)
```

Obviously, just passing `enforce_actionview_erubi_equality: true` also works. In that case we only enforcing the equality of the ActionView evaluated engine output:

 ```ruby
assert_evaluated_snapshot(template, {},  enforce_actionview_erubi_equality: true)
```

Introducing this option allows us to properly test https://github.com/marcoroth/herb/pull/1492.

### DIFF
--- a/test/snapshot_utils.rb
+++ b/test/snapshot_utils.rb
@@ -71,6 +71,7 @@ module SnapshotUtils
     require "prism"
 
     enforce_erubi_equality = kwargs.delete(:enforce_erubi_equality) || false
+    enforce_actionview_erubi_equality = kwargs.key?(:enforce_actionview_erubi_equality) ? kwargs.delete(:enforce_actionview_erubi_equality) : enforce_erubi_equality
     engine_options = options.merge(kwargs)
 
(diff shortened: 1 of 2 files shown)


joelhawksley:
@marcoroth you beat me to it! I literally had the idea to write this PR while falling asleep last night 😂

marcoroth:
@joelhawksley I'm really happy we think alike 😅

PR #1508, opened 2026-03-27
Engine: Ignoring vendor directory considered harmful

joelhawksley:
By default, Herb currently ignores the `vendor` directly. This is potentially quite harmful, as it could expose consuming applications to runtime exceptions.

For example. We vendor https://github.com/primer/view_components in our application. The dependency includes ViewComponents with ERB files, including a few that will not compile with Herb. If we don't compile the templates in our test environment, those Herb compilation errors will not be caught and would raise in production.

I see similar risks for other gems that provide UI, such as as dashboards mounted via an engine.

I'm wary of blankly excluding the directory silently by default, especially when we describe the behavior of `herb analyze` as `Running herb analyze without arguments now defaults to the current directory` (https://herb-tools.dev/blog/whats-new-in-herb-v0-9#improved-herb-analyze-command). This could make one believe that all subdirectories are being analyzed based on the config in the user's `.herb.yml` file, but in fact it is the combination of `defaults.yml` and the user's configuration.

It also appears that adding `'vendor/**/*.erb'` to `include` _still_ results in the directory being excluded. I had to explicitly call `bundle exec herb analyze vendor` to see errors from the vendor folder!

Of course, this does nothing to avoid similar issues for gems that provide UI that are not vendored. I believe we should include a boot-time safe-guard for this issue in ReActionView at the very least. Or, we could look into ways of not using Herb to compile ERB from gems, but that sounds like a mess to me.


joelhawksley:
@marcoroth while I've opened https://github.com/marcoroth/herb/pull/2406 as a less-obtrusive fix, I don't think it's the best long-term fix. Excluding `vendor` opens up the possibility for Herb to introduce runtime exceptions when migrating from Erubi. I don't believe that is acceptable.

marcoroth:
Closing this in favor of #2406 and https://github.com/marcoroth/reactionview/pull/168

PR #1517, opened 2026-03-28
Linter: Implement `a11y-no-aria-unsupported-elements` rule

### LINKED ISSUE #176: Linter Rule: No aria on unsupported elements
##### Description

Prevent usage of ARIA roles, states and properties on certain DOM elements.

##### Rationale

These elements include:

- html
- meta
- script
- style

This often because these elements are not visible

##### ✅ Good

```html+erb
<meta charset="UTF-8" />
```

```html+erb
<html lang="en"></html>
```

```html+erb
<script></script>
```

##### 🚫 Bad

```html+erb
<meta charset="UTF-8" aria-hidden="false" />
```

```html+erb
<html lang="en" role="application"></html>
```

```html+erb
<script aria-hidden="false"></script>
```

#### References

Inspired by [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-aria-unsupported-elements.md)

markokajzer:
closes #176

## Description

Prevent usage of ARIA roles, states, and properties on certain DOM elements that do not support them.

## Rationale

Certain HTML elements do not support ARIA roles, states, or properties because they are not visible or interactive. Adding ARIA attributes to these elements has no effect and indicates a misunderstanding of the ARIA specification.

These elements include:

- `<html>`
- `<meta>`
- `<script>`
- `<style>`

Also handles `javascript_tag` helper with `aria_label` option key.

### DIFF
--- a/javascript/packages/linter/src/rules/a11y-no-aria-unsupported-elements.ts
+++ b/javascript/packages/linter/src/rules/a11y-no-aria-unsupported-elements.ts
@@ -13,41 +13,27 @@ const UNSUPPORTED_ELEMENTS = new Set([
 ])
 
 class NoAriaUnsupportedElementsVisitor extends BaseRuleVisitor {
-  visitHTMLElementNode(node: HTMLElementNode): void {
-    const tagName = getTagLocalName(node)
-
-    if (tagName && UNSUPPORTED_ELEMENTS.has(tagName)) {
-      this.checkAttributes(node, tagName)
-    }
-
+  private currentTagName: string | null = null
+  
+  visitHTMLElementNode(node: HTMLElementNode): void {  
+    this.currentTagName = getTagLocalName(node)
     super.visitHTMLElementNode(node)
+    this.currentTagName = null
   }
-
-  private checkAttributes(node: HTMLElementNode, tagName: string): void {
-    const attributes = this.getAttributeNodes(node)
-
-    for (const attributeNode of attributes) {
-      const attributeName = getAttributeName(attributeNode)
-      if (!attributeName) continue
-      if (!attributeName.startsWith("aria-") && attributeName !== "role") continue
-
-      this.addOffense(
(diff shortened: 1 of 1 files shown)


javascript/packages/linter/src/rules/a11y-no-aria-unsupported-elements.ts
```diff
@@ -0,0 +1,78 @@
...
+    const openTag = node.open_tag
+
+    if (isERBOpenTagNode(openTag)) {
+      return filterHTMLAttributeNodes(openTag.children)
+    } else if (isHTMLOpenTagNode(openTag)) {
+      return filterHTMLAttributeNodes(openTag.children)
+    } else {
+      return []
+    }
+  }
```
marcoroth: We can simplify this rule a bit if we collect the tag name in the `visitHTMLElementNode` and then check for it directly in the `visitHTMLAttributeNode`. This way, we don't need to know the details and specifics about the `HTMLElementNode.open_tag` being a `HTMLOpenTagNode` or `ERBOpenTagNode`  🙌🏼 

```suggestion
  private currentTagName: string | null = null
  
  visitHTMLElementNode(node: HTMLElementNode): void {  
    this.currentTagName = getTagLocalName(node)
    super.visitHTMLElementNode(node)
    this.currentTagName = null
  }
  
  visitHTMLAttributeNode(node: HTMLAttributeNode): void {
    if (!this.currentTagName || !UNSUPPORTED_ELEMENTS.has(this.currentTagName)) return
  
    const attributeName = getAttributeName(node)
    if (!attributeName) return
    if (!attributeName.startsWith("aria-") && attributeName !== "role") return
  
    this.addOffense(
      `The \`${attributeName}\` attribute is not supported on the \`<${this.currentTagName}>\` element. ARIA roles, states, and properties should not be used on elements that are not visible or not interactive.`,
      node.location,
    )
  
    super.visitHTMLAttributeNode(node)
  }
```

markokajzer: nice much simpler! 🙏

PR #1520, opened 2026-03-28
Linter: Implement `a11y-no-accesskey-attribute` rule

### LINKED ISSUE #174: Linter Rule: No accesskey attribute
##### Description

Enforce no `accesskey` attribute on element. Access keys are HTML attributes that allow web developers to assign keyboard shortcuts to elements. 

##### Rationale

Inconsistencies between keyboard shortcuts and keyboard commands used by screenreader and keyboard only users create accessibility complications so to avoid complications, access keys should not be used.
##### ✅ Good

```html+erb
<div></div>
```

##### 🚫 Bad

```html+erb
<div accesskey="h"></div>
```

#### References

Inspired by [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-accesskey-attribute.md)

markokajzer:
closes #174

## Description

Enforce no `accesskey` attribute on elements. Access keys are HTML attributes that allow web developers to assign keyboard shortcuts to elements.

## Rationale

Inconsistencies between keyboard shortcuts and keyboard commands used by screen readers and keyboard-only users create accessibility complications. To avoid these complications, access keys should not be used.

### DIFF
--- a/javascript/packages/linter/src/rules/a11y-no-accesskey-attribute.ts
+++ b/javascript/packages/linter/src/rules/a11y-no-accesskey-attribute.ts
@@ -1,30 +1,19 @@
-import type { ParseResult, ParserOptions, HTMLElementNode } from "@herb-tools/core"
-import { hasAttribute, isERBOpenTagNode, findAttributeByName } from "@herb-tools/core"
+import { type ParseResult, type ParserOptions, type HTMLAttributeNode, getAttributeName } from "@herb-tools/core"
 
 import { BaseRuleVisitor } from "./rule-utils.js"
 import { ParserRule } from "../types.js"
 import type { UnboundLintOffense, LintContext, FullRuleConfig } from "../types.js"
 
 class NoAccesskeyAttributeVisitor extends BaseRuleVisitor {
-  visitHTMLElementNode(node: HTMLElementNode): void {
-    if (this.hasAccesskey(node)) {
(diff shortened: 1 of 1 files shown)


javascript/packages/linter/src/rules/a11y-no-accesskey-attribute.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  private hasAccesskey(node: HTMLElementNode): boolean {
+    const openTag = node.open_tag
+
+    if (isERBOpenTagNode(openTag)) {
+      return findAttributeByName(openTag.children, "accesskey") !== null
+    } else {
+      return hasAttribute(node, "accesskey")
+    }
+  }
```
marcoroth: I think you can directly visit the `HTMLAttributeNode` so we don't need the decision logic in `hasAccesskey` to know if it's a `ERBOpenTagNode` or not:

```suggestion  
  visitHTMLAttributeNode(node: HTMLAttributeNode): void {  
    if (getAttributeName(node) === "accesskey") {
      this.addOffense(
        "Avoid using the `accesskey` attribute. Inconsistencies between keyboard shortcuts and keyboard commands used by screen readers and keyboard-only users create accessibility complications.",
        node.location,
      )
    }
  
    super.visitHTMLAttributeNode(node)
  }    
```

PR #1528, opened 2026-03-28
Linter: Implement `erb-no-unused-literals` rule

### LINKED ISSUE #179: Linter Rule: Disallow Ruby literals in ERB without output
##### Description

Disallow Ruby literal values inside ERB tags that do not produce output or side effects. Writing literals like `<% "string" %>` or `<% 123 %>` has no effect. The value is evaluated and discarded, making the line functionally useless.

This rule detects and warns about ERB tags containing:

- Literal strings
- Numeric values
- `true` and `false`
- `nil`
- Symbols
- Arrays, hashes, and ranges
- Regular expressions

##### Rationale

These expressions are evaluated but unused, they don't produce output (not `<%= ... %>`), and they don't perform side effects (no assignments, method calls, or control flow). They're likely accidental leftovers, debugging artifacts, or misunderstandings of ERB syntax.

##### ✅ Good

```html+erb
<% if logged_in? %>
  <p>Welcome!</p>
<% end %>

<%= "Hello, #{user.name}" %>
```

##### 🚫 Bad

```html+erb
<% "Logged in" %>
<% 42 %>
<% true %>
<% false %>
<% nil %>
<% [:foo, :bar] %>
<% /pattern/ %>
<% { key: "value" } %>
<% :symbol %>
```

#### References

\-

markokajzer:
closes #179

## Description

Disallow Ruby literal values inside ERB tags that do not produce output or side effects. Writing literals like `<% "string" %>` or `<% 123 %>` has no effect. The value is evaluated and discarded, making the line functionally useless.

This rule detects and warns about ERB tags containing:

- Literal strings
- Numeric values (integers, floats, rationals, imaginary)
- `true` and `false`
- `nil`
- Symbols
- Arrays, hashes, and ranges
- Regular expressions

## Rationale

These expressions are evaluated but unused, they don't produce output (not `<%= ... %>`), and they don't perform side effects (no assignments, method calls, or control flow). They're likely accidental leftovers, debugging artifacts, or misunderstandings of ERB syntax.

### DIFF
--- a/javascript/packages/linter/src/rules/erb-no-unused-literals.ts
+++ b/javascript/packages/linter/src/rules/erb-no-unused-literals.ts
@@ -15,7 +15,12 @@ class LiteralCollector extends PrismVisitor {
 
   override visit(node: PrismNode): void {
     if (!node) return
-    if (isAssignmentNode(node) || isCallNode(node)) return
+    if (isAssignmentNode(node)) return
+
+    if (isCallNode(node)) {
+      this.visit(node.receiver)
+      return
+    }
 
     super.visit(node)
   }
@@ -61,6 +66,10 @@ class LiteralCollector extends PrismVisitor {
   }
 
   visitStringNode(node: PrismNodes.StringNode): void {
+    this.literals.push(node)
+  }
+
+  visitInterpolatedStringNode(node: PrismNodes.InterpolatedStringNode): void {
     this.literals.push(node)
   }
 
--- a/javascript/packages/linter/test/rules/erb-no-unused-literals.test.ts
+++ b/javascript/packages/linter/test/rules/erb-no-unused-literals.test.ts
@@ -121,6 +121,109 @@ describe("ERBNoUnusedLiteralsRule", () => {
     `)
   })
 
+  test("fails for interpolated string literals", () => {
+    expectWarning('Avoid using silent ERB tags for literals. `"Hello #{name}"` is evaluated but never used or output.')
+
+    assertOffenses(dedent`
+      <% "Hello #{name}" %>
+    `)
+  })
+
+  test("fails for arithmetic on literal receivers", () => {
+    expectWarning('Avoid using silent ERB tags for literals. `10` is evaluated but never used or output.')
+
+    assertOffenses(dedent`
+      <% 10 / 1 %>
+    `)
+  })
+
+  test("fails for string method calls on literal receivers", () => {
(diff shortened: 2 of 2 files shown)


javascript/packages/linter/test/rules/erb-no-unused-literals.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      <% :symbol %>
+    `)
+  })
+
+  test("fails for true literals", () => {
+    expectWarning("Avoid using silent ERB tags for literals. `true` is evaluated but never used or output.",)
+
+    assertOffenses(dedent`
+      <% true %>
+    `)
```
marcoroth, line 121: Can we make sure we don't flag assignments? Like:
```erb
<% names = ["Alice", "Bob"] %>
```

javascript/packages/linter/src/rules/erb-no-unused-literals.ts
```diff
@@ -0,0 +1,72 @@
...
+  "SymbolNode",
+  "TrueNode",
+])
+
+function isLiteralPrismNode(prismNode: PrismNode): boolean {
+  const type: string = prismNode?.constructor?.name
+  if (!type) return false
+
+  return LITERAL_NODE_TYPES.has(type)
+}
```
marcoroth: I think we could implement this using a `PrismVisitor`, so we can catch things that aren't directly on the `prismNode` top-level itself.

It might also help to catch more untypical things, like:
```erb
<% valid? && "success" %>
```

javascript/packages/linter/test/rules/erb-no-unused-literals.test.ts
markokajzer: added an array and a hash example to the `passes for assignments` test

javascript/packages/linter/src/rules/erb-no-unused-literals.ts
markokajzer: great idea!

markokajzer: changed to
1. setup a collector via `PrismVisitor`
2. use the visit methods, e.g. `visitArrayNode` to register literals
3. ignore assignment nodes and call nodes
4. add an additional spec for `<% valid? && "success" %>`

marcoroth: Thank you!

PR #1531, opened 2026-03-29
Linter: Implement `erb-no-unused-expressions` rule

marcoroth:
Disallow expressions inside silent ERB tags (`<% %>`) whose return values are discarded. Writing expressions like `<% @user.name %>` or `<% helper_method(arg) %>` evaluates the expression but discards the result, making the line functionally useless.

This rule detects and warns about silent ERB tags containing:

- Method calls without blocks
- Instance variable reads (`@user`)
- Class variable reads (`@@count`)
- Global variable reads (`$global`)
- Constant reads (`CONSTANT`, `Foo::Bar`)


brunoprietog:
This could result in some false positives, such as when using the `turbo_refreshes_with` helper, which doesn't require `<%=`.

marcoroth:
Thanks @brunoprietog great catch! I opened #1586 to allow for those Turbo methods that have side effects 🙌🏼

PR #1538, opened 2026-03-29
Linter: Fix `erb-no-unused-literals` false positives

marcoroth:
The `LiteralCollector` relied on a `visit()` override to prevent traversal into `CallNode` arguments. However, `PrismVisitor` dispatches via `node.accept()` which calls `visitCallNode()` directly, bypassing the `visit()` override.

This caused literals used as arguments (e.g., `:key` in `@hash[:key] = value`) to be incorrectly flagged.

This pull request fixes this by overriding `visitCallNode` instead, which is always called regardless of the dispatch path.


markokajzer:
oh... sorry about that mate! feel free to just let me know of edge cases you find the rules we're adding, you're probably using the tool the most. i'm more than happy to fix my buggy code 🙇

marcoroth:
@markokajzer no worries, I just found out about them too late. But also, I think this "fix" actually broke some of the other cases that were previously working, see #1547 🙈

PR #1546, opened 2026-03-30
Engine: Add Erubi line parity tests for comment tags

joelhawksley:
I'm still working to better isolate this issue, but I ran into a case where error propagation in Herb did not match Erubi. Do these failing tests ring a bell to you @marcoroth?

### DIFF
--- a/test/engine/rails_compatibility_test.rb
+++ b/test/engine/rails_compatibility_test.rb
@@ -112,5 +112,50 @@ module Engine
 
       assert_equal rails_result, reactionview_result
     end
+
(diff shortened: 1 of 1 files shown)


joelhawksley:
I found a different underlying cause that was specific to our codebase. Closing!

marcoroth:
@joelhawksley are you saying Herb is already compliant with Erubi for those cases? If so, would it still make sense to add the tests to lock in the behavior?

joelhawksley:
@marcoroth I'm not sure yet. I'll follow up when I'm more confident in my observation of the issue we're seeing ❤️

PR #1548, opened 2026-03-30
Engine: Add remaining failing test from Erubi test suite

joelhawksley:
As a fun experiment, I had Claude re-create the tests from https://github.com/jeremyevans/erubi/blob/master/test/test.rb inside Herb and see if there were any gaps. The original run surfaced 8 failing tests.

Re-verifying against current `main` (2026-09-14), **7 of the 8 tests now pass** on the back of recent engine work, so this PR is now down to the last remaining gap.

### Tests fixed on `main` since the original run

| Test | Fixed by |
| --- | --- |
| `<%%` literal escape (3 tests) | #2392 (guard → validator) + #2628 (escaped control flow as literal) |
| Trailing content after `%>` | #2642 (keep indentation of a first-line tag) |
| `trim: false` (2 tests) | #2391 (support Erubi's `trim` option) |
| `literal_prefix` / `literal_postfix` | #2647 (support Erubi's `literal_prefix` / `literal_postfix`) |

### Remaining failure

| Category | Result on current `main` | Issue |
| --- | --- | --- |
| `<%` inside a Ruby string | `Herb::Engine::ParseError` | `<% x = '<%' %>` — the parser terminates the outer ERB tag at the `<%` inside the string literal instead of treating it as string content. |


test/engine/erubi_test_suite_test.rb
```diff
@@ -0,0 +1,88 @@
...
+# Each test verifies that Herb::Engine produces the exact same evaluated output
+# as Erubi::Engine for the given template and options.
+module Engine
+  class ErubiTestSuiteTest < Minitest::Spec
+    include SnapshotUtils
+
+    ERUBI_OPTS = { enforce_erubi_equality: true, enforce_actionview_erubi_equality: false }.freeze
+
+    test "handles literal erb escape with double percent" do
+      template = <<~'ERB'.chomp
```
joelhawksley:
```suggestion
      template = <<~ERB.chomp
```

test/engine/erubi_test_suite_test.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    test "code tag with string containing erb-like content" do
+      template = "<% x = '<%' %><%= x %>"
+
+      assert_evaluated_snapshot(template, {}, **ERUBI_OPTS)
+    end
+
+    test "literal percent with different prefix and postfix" do
+      template = <<~'ERB'.chomp
```
joelhawksley:
```suggestion
      template = <<~ERB.chomp
```

joelhawksley:
@marcoroth I've refreshed this PR and it looks like we are still failing the same checks, FWIW.

PR #1553, opened 2026-03-31
Engine: Fix leading whitespace before inline control tags after trimmed code lines

joelhawksley:
When a code-only line (e.g. `<% end %>`) is followed by a line starting with an inline control tag (`<% unless %><%= expr %><% end %>`), the leading whitespace before the control tag is incorrectly stripped.

**Root cause:** `leading_space_follows_newline?` doesn't recognize whitespace-only text tokens that follow a trim-consumed newline. After a code-only line trims its trailing newline, the remaining whitespace token no longer matches `/\n[ \t]+\z/` because the `\n` was consumed.

**Fix:** Also check `@last_trim_consumed_newline && /\A[ \t]+\z/` to recognize whitespace tokens that effectively follow a consumed newline.


joelhawksley:

PR #1554, opened 2026-03-31
Engine: Fix right-trim on expression block opening tags

joelhawksley:
Expression blocks with right-trim (`<%= ... do -%>`) did not trim the trailing newline. Erubi trims it correctly.

**Root cause:** `visit_erb_block_node` never checks for `-%>` on the tag closing when processing expression blocks. The right-trim logic that exists for regular output and control tags was missing for expression blocks.

**Fix:** After emitting the expression block token, check `node.tag_closing&.value == "-%>"` and set `@trim_next_whitespace = true` if present.

### DIFF
--- a/lib/herb/engine/compiler.rb
+++ b/lib/herb/engine/compiler.rb
@@ -259,10 +259,9 @@ module Herb
                      else
                        [:expr_block, code, current_context]
                      end
-          @last_trim_consumed_newline = false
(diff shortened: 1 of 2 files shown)


joelhawksley:

marcoroth (approved):
Thank you! 🙏🏼

PR #1555, opened 2026-03-31
Engine: Add Erubi line parity test for ERB comment lines

joelhawksley:
Follow up on #1546.

This pull request keeps the test case from this branch and drops the engine change. Herb now emits a line for each standalone ERB comment, so a compiled template evaluated with `module_eval` and a line offset attributes errors to the same line Erubi would. What is left worth landing is the regression test.

The test compiles the template through both engines and asserts they agree on the line count, using a new `assert_erubi_line_parity` helper alongside the existing snapshot assertion.

### DIFF
--- a/javascript/packages/config/src/config.ts
+++ b/javascript/packages/config/src/config.ts
@@ -124,12 +124,12 @@ export class Config {
 
   public readonly path: string
   public config: HerbConfig
-  public readonly configVersion: string
+  public readonly configVersion: string | undefined
 
   constructor(projectPath: string, config: HerbConfig, configVersion?: string) {
     this.path = Config.configPathFromProjectPath(projectPath)
     this.config = config
-    this.configVersion = configVersion ?? config.version
+    this.configVersion = configVersion
   }
 
   get projectPath(): string {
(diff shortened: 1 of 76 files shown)


joelhawksley:

marcoroth:
@joelhawksley I'm happy to merge this, but just want to understand: The evaluated output was already correct and this is about more about "cosmetics", right?

marcoroth:
Actually, does Erubi really try to make an effort to keep the code on the same line as it was in the source?

joelhawksley:
@marcoroth As it turns out, we didn't need this change to pass CI. We are green on `0.9.4`! I still think it's a good idea though, as it's nice to have error messages raised on the same lines as the original template file.

marcoroth:
@joelhawksley awesome, that's great to hear and thanks for sharing! 🎉 

I agree, it would be nice to have it aligned and behave the same way. Is there a reason why you closed https://github.com/marcoroth/herb/pull/1556? Doesn't that serve a similar purpose?

joelhawksley:
@marcoroth I closed https://github.com/marcoroth/herb/pull/1556 because I didn't need it to pass CI. I left this one open because you commented on it ❤️ 

I'll let you decide whether to take any of them 😄

marcoroth:
Oh my bad then, I only commented on one because they looked so similar. But if https://github.com/marcoroth/herb/pull/1556 is actually also valid on its own then lets re-open it 🙌🏼

PR #1556, opened 2026-03-31
Engine: Add Erubi line parity test for multi-line code blocks

joelhawksley:
Follow up on #1546 and #1555.

This pull request keeps the test case from this branch and drops the engine change. Multi-line `<% ... %>` blocks with the delimiters on their own lines already compile to the same number of lines as Erubi, so what is left worth landing is the regression test.

The test compiles the template through both engines and asserts they agree on the line count, using the `assert_erubi_line_parity` helper from #1555 alongside the existing snapshot assertion.


joelhawksley:

marcoroth:
Whoops, sorry. A bad push closed the PR. I'm going to open a new one.

marcoroth:
Resolved via https://github.com/marcoroth/herb/pull/2520

PR #1559, opened 2026-03-31
Linter: Allow `content_for` and `provide` in `erb-no-unused-expressions`

marcoroth:
This pull request updates the `erb-no-unused-expressions` linter rule to allow side effect method calls like `content_for` without blocks. This was previously flagged:

```erb
<% content_for :title, "My Site" %>
```

### DIFF
--- a/javascript/packages/linter/src/rules/erb-no-unused-expressions.ts
+++ b/javascript/packages/linter/src/rules/erb-no-unused-expressions.ts
@@ -25,6 +25,12 @@ const MUTATION_METHODS = new Set([
   "concat",
 ])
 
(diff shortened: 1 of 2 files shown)


janko:
Would it make sense to also include Turbo helpers that internally call `content_for`/`provide`:

* `turbo_refreshes_with`
* `turbo_exempts_page_from_cache`
* `turbo_exempts_page_from_preview`
* `turbo_page_requires_reload`
 
Or would that be out of scope for Herb?

marcoroth:
@janko I think it's also fair to add them 👍🏼

marcoroth:
Opened https://github.com/marcoroth/herb/pull/1586

PR #1571, opened 2026-04-01
Formatter: Fix trailing whitespace in `<pre>` with inline child elements

hakanensari:
After upgrading to the latest Herb version, formatting `<pre><code>` blocks nested inside another element became non-idempotent.

Each formatter run appends trailing whitespace before the closing tag, which then triggers lint errors.

I added a failing test.


marcoroth:
Released with [v0.9.5](https://github.com/marcoroth/herb/releases/tag/v0.9.5) 🙌🏼

Issue #1111, opened 2026-01-30
Debug overlay wrappers break CSS sibling combinators (~, +) used by Tailwind peer-*   variants


thomaswitt:
## Problem

Herb's development-mode debug overlay wraps each <%= tag.* %> ERB output in <span  data-herb-debug-outline-type="erb-output component" style="display: contents">. While display: contents correctly makes these wrappers invisible to CSS layout, it does not affect the DOM tree structure. This breaks CSS sibling combinators (~, +), which operate on DOM siblings, not layout siblings.

This is particularly impactful for Tailwind CSS peer-* variants (e.g., peer-checked:, peer-focus:), which rely on the general sibling combinator (~) to style elements based on a sibling's state.

## Reproduction

Minimal ERB using Rails tag.* helpers and Tailwind's Flowbite toggle pattern:

```
<%= tag.label(class: 'relative inline-flex cursor-pointer items-center') do %>
  <%= tag.input(type: 'checkbox', class: 'sr-only peer', checked: true) %>
  <%= tag.div(class: 'peer h-6 w-11 rounded-full bg-gray-200 peer-checked:bg-blue-600') %>
<% end %>
```

**Expected DOM (without Herb)**

```
<label class="relative inline-flex cursor-pointer items-center">
  <input type="checkbox" class="sr-only peer" checked>
  <div class="peer h-6 w-11 rounded-full bg-gray-200 peer-checked:bg-blue-600"></div>
</label>
```

input.peer:checked ~ div.peer-checked\:bg-blue-600 matches → toggle shows blue.

**Actual DOM (with Herb debug overlay)**

```
<label class="relative inline-flex cursor-pointer items-center">
  <span data-herb-debug-outline-type="erb-output component" style="display: contents">
    <input type="checkbox" class="sr-only peer" checked>
  </span>
  <span data-herb-debug-outline-type="erb-output component" style="display: contents">
    <div class="peer h-6 w-11 rounded-full bg-gray-200 peer-checked:bg-blue-600"></div>
  </span>
</label>
```

input and div are no longer DOM siblings → CSS ~ combinator doesn't match → toggle always
appears gray/off, even when checked.

## Impact

- Visual state is lost: Checked checkboxes appear unchecked because peer-checked: never activates
- Affects all peer-* variants: peer-focus:, peer-hover:, peer-disabled:, etc.
- Also affects group-* variants if tag.* helpers are used (same ~ combinator pattern)
- Adjacent sibling combinator (+) is equally affected
[...]

PR #1613, opened 2026-04-05
Engine: Compile-time optimizations for Action View tag helpers

### LINKED ISSUE #653: Engine: Optimize Action View Tag Helpers when compiling template
We can optimize static or partially static Action View Tag Helpers, since the Herb Parser is soon going to understand and transform Action View Tag Helpers to HTML-elements.

## Fully Static with dynamic content

For example, for fully static Tag Helpers:

```erb
<%= tag.p class: "flex" do %>
  <span><%= content %></span>
<% end %>
```

Currently gets compiled to:

```ruby
__herb = ::Herb::Engine; _buf = ::String.new;
 _buf << __herb.h(tag.p class: "flex" do); _buf << '
  <span>'.freeze; _buf << __herb.h((content)); _buf << '</span>
'.freeze; end; _buf << '
'.freeze;
_buf.to_s
```

Since everything in this Tag-helper call is static, we could optimize it to avoid the helper-related method calls at runtime:
```ruby
__herb = ::Herb::Engine; _buf = ::String.new;
 _buf << '<p class="flex">
  <span>'.freeze; _buf << __herb.h((content)); _buf << '</span>
</p>
'.freeze;
_buf.to_s
```

## Fully Static with static content

For fully static tag helper calls with only static content:

```erb
<%= tag.p class: "flex" do %>
  <span>Content</span>
<% end %>
```

We can inline everything into a single string:

```ruby
__herb = ::Herb::Engine; _buf = ::String.new;
 _buf << '<p class="flex">
  <span>Content</span>
</p>
'.freeze;
_buf.to_s
```

## Partially Static

For partially static Tag Helpers like this:
```erb
<%= tag.p class: classes do %>
  <span><%= content %></span>
<% end %>
```

We could optimize it to:

```ruby
__herb = ::Herb::Engine; _buf = ::String.new;
 _buf << '<p class="'.freeze; _buf << ::Herb::Engine.attr((classes)); _buf << '">
  <span>'.freeze; _buf << __herb.h((content)); _buf << '</span>
</p>
'.freeze;
_buf.to_s
```

marcoroth:
This pull request adds compile-time optimization support for Action View tag helpers to `Herb::Engine`.

When `optimize: true` is passed to `Herb::Engine`, the parser transforms Action View helper calls (`tag.*`, `content_tag`, `link_to`, `image_tag`, `javascript_tag`, `javascript_include_tag`, `turbo_frame_tag`) into their HTML equivalents at compile time, producing optimized Ruby output that avoids runtime helper dispatch.

When compile-time optimizations are enabled, the engine checks whether the template contains supported Action View helpers via the `Herb::ActionView::HelperRegistry.supported` method. If helpers are detected, the parser is invoked with `action_view_helpers: true` and `transform_conditionals: true`, which also handles postfix if/unless and ternary conditionals containing helper calls.

The pull request also includes comprehensive snapshot-based tests for all supported helpers, an `assert_optimized_output_match` helper that validates optimized output against both standard Herb and ActionView+Erubi to ensure the rendered output matches.

Additionally, we added a YAML-driven benchmark framework in `bench/action_view/` with a `bin/bench` CLI for compile-time and render-time comparisons. Early benchmarks show render-time improvements of 3-22x depending on the template, with a realistic 192-line page layout rendering 22x faster, and a helper-heavy template with all static values reaching up to 150x faster. 

The trade-off is compile time: Herb's parser is currently 10-90x slower than Erubi at compile time, so optimization adds overhead that needs to be amortized over multiple renders. But this is typically won back with 15-100 renders depending on template complexity and could also be done ahead of time, instead of the current on-the-fly approach used in Action View today.

  > [!WARNING]
  > Compile-time optimizations are experimental. Output may differ from standard ActionView rendering.

CleanShot 2026-04-05 at 23 58 07@2x

<br />

Resolves https://github.com/marcoroth/herb/issues/653
Enables https://github.com/marcoroth/herb/issues/1111


joelhawksley:
@marcoroth did you test this against cases where the consumer re-defines an ActionView helper, say to add additional security features to `csrf_meta_tags`? I'll need to track down what our cases look like, but I believe we have a few places where we add additional functionality. It could be pretty unfortunate if we stripped out a re-defined helper silently.

marcoroth:
Hey @joelhawksley, this feature is very experimental and opt-in. This currently also only applies for the helpers that the registry marks as "supported":
```ruby
❯ bin/console
irb(main):001> require "herb/action_view/helper_registry"
=> true
irb(main):002> Herb::ActionView::HelperRegistry.supported.map(&:name)
=> ["image_tag", "javascript_include_tag", "javascript_tag", "content_tag", "tag", "link_to", "turbo_frame_tag"]
irb(main):003>
```

But yes, there is currently a gap when you overwrite `content_tag` or any other helper. As of right now they would get compiled away. The idea I had in mind to catch cases like these was to introduce a new development-only transform visitor that compiles in some runtime inspections whenever we compile away a helper. So that it could tell you at runtime that the helper method that we compiled away wasn't the original one from ActionView.


So if you had this:

```erb
<%= tag.div do %>
  Content
<% end %>
```

it would currently compile this template with the compile-time optimizations to this:

**`bin/herb compile test.html.erb --optimize`**
```ruby
__herb = ::Herb::Engine; _buf = ::String.new; _buf << '<div>
  Content
</div>'.freeze;
_buf.to_s
```
and I was thinking of adding a pre- or postamble in development/test to either raise, or have it collect the warnings and show them like we just did in #1660:
```diff
__herb = ::Herb::Engine; _buf = ::String.new; _buf << '<div>
  Content
</div>'.freeze;
_buf.to_s
+
+ raise "Helper was overwritten" if method(:tag).owner != Herb::ActionView::HelperRegistry.get(:tag).source
```

joelhawksley:
Ah yep, that makes sense. I just poked around a little bit and we have overrides for about a dozen or so of the helpers, so it's definitely something we'd want to leave turned off by default until we can be 100% confident in avoiding a mismatch.

In general, I'd argue that a runtime check is insufficient, as it assumes that every line of ERB is executed in CI.

marcoroth:
@joelhawksley I just opened https://github.com/marcoroth/herb/issues/1664 so we don't loose track of this, thanks for the input!

> In general, I'd argue that a runtime check is insufficient, as it assumes that every line of ERB is executed in CI.

Yeah, I agree. Though, it wouldn't need to execute every line of ERB since it's either added to the start/end of the file, so rendering the template once in any way would be sufficient to fire the check at least once. 

But there might also be a potential other way to do this in a more static manner by instantiating an empty `ActionView::ViewContext` and check the Source Location of the defined helpers that way and add that to something like `herb actionview check` (introduced via #1663) or `herb analyze`.

PR #1622, opened 2026-04-10
Language Server: Report language server version on initialization

janko:
Some language servers I use report versions, such as [Ruby LSP](https://github.com/Shopify/ruby-lsp/blob/d4a9dfa3df4e597830f2260f5b1dd40285e7a8ba/lib/ruby_lsp/server.rb#L295-L298) and `vtsls`, which Zed then shows in its summary card.

Screenshot 2026-04-10 at 16 23 04

### DIFF
--- a/javascript/packages/language-server/src/server.ts
+++ b/javascript/packages/language-server/src/server.ts
@@ -45,7 +45,7 @@ export class Server {
 
       const result: InitializeResult = {
         serverInfo: {
-          name: "Herb LSP",
+          name: "Herb Language Server",
           version,
         },
         capabilities: {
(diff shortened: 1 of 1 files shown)


javascript/packages/language-server/src/server.ts
```diff
@@ -43,6 +44,10 @@ export class Server {
       })
 
       const result: InitializeResult = {
+        serverInfo: {
+          name: "Herb LSP",
```
marcoroth:
```suggestion
          name: "Herb Language Server",
```

marcoroth (approved):
Great idea, thank you @janko! 🙏🏼

PR #1632, opened 2026-04-15
Config: Support YAML anchors and aliases in `.herb.yml`

### LINKED ISSUE #1631: Configuration: YAML anchors/aliases in .herb.yml raise Psych::AliasesNotEnabled
## Summary

`Herb::Configuration#load_config` calls `YAML.safe_load_file` without `aliases: true`, which raises `Psych::AliasesNotEnabled` at runtime when a project's `.herb.yml` uses YAML anchors and aliases — a common way to DRY up repeated rule definitions.

## Reproduction

`.herb.yml`:

```yaml
files:
  include: &patterns
    - "**/*.custom.erb"
  exclude: *patterns
```

Then:

```ruby
require "herb"
Herb::Configuration.load(Dir.pwd)
```

Raises:

```
Psych::AliasesNotEnabled: Alias parsing was not enabled.
To enable it, pass `aliases: true` to `Psych::load` or `Psych::safe_load`.
```

This surfaces in any runtime that triggers `Configuration.load` — for example, when [reactionview](https://github.com/marcoroth/reactionview) is installed with `config.intercept_erb = true`, every request renders a page that crashes with this error. Standalone `bundle exec herb lint` appears unaffected because it uses a different code path (presumably Psych's default `load`, which enables aliases).

## Root cause

[`lib/herb/configuration.rb#L228`](https://github.com/marcoroth/herb/blob/v0.9.6/lib/herb/configuration.rb#L228):

```ruby
user_config = YAML.safe_load_file(@config_path, permitted_classes: [Symbol]) || {}
```

`Psych.safe_load` defaults `aliases` to `false` to mitigate the "billion laughs" DoS, which is appropriate for untrusted YAML. But `.herb.yml` is developer-authored local config — same threat model as `database.yml`, `credentials.yml`, or `.rubocop.yml`, all of which enable aliases.

## Proposed fix

Pass `aliases: true` to the `YAML.safe_load_file` call.

## Affected versions

Reproduced on `herb` 0.9.5 and 0.9.6 (current latest).

## Pull request

I've opened #1632 with the one-line fix plus a regression test covering anchor/alias loading.

matiasleidemer:
## Summary

Fixes #1631 — `Herb::Configuration#load_config` raises `Psych::AliasesNotEnabled` at runtime when `.herb.yml` uses YAML anchors and aliases.

The fix is a one-character addition: pass `aliases: true` to `YAML.safe_load_file`. Added a regression test that exercises the anchor/alias case end-to-end through `Herb::Configuration.load`.

## Rationale

`Psych.safe_load` defaults `aliases: false` to mitigate the "billion laughs" DoS against untrusted YAML. But `.herb.yml` is developer-authored local config — same trust model as `database.yml`, `credentials.yml`, and `.rubocop.yml`, all of which enable aliases. Anchors/aliases are a common ergonomics pattern for config files with repeated rule definitions (the project I hit this on has ~60 aliases in its `.herb.yml`).

## Reproduction

Before this PR, loading a config like:

```yaml
files:
  include: &patterns
    - "**/*.custom.erb"
  exclude: *patterns
```

raises `Psych::AliasesNotEnabled`. After this PR, it loads cleanly and both `include` and `exclude` resolve to the aliased list. See the new test `loads configuration using YAML anchors and aliases` in `test/configuration_test.rb`.

## Changes

- `lib/herb/configuration.rb`: add `aliases: true` to the `YAML.safe_load_file` call
- `test/configuration_test.rb`: new regression test

## Test plan

- [x] `bundle exec rake test` locally for `ConfigurationTest` (full suite needs native build; maintainer CI will exercise it end-to-end)
- [x] Verified underlying Psych call with and without `aliases: true` against a minimal YAML fixture


marcoroth:
Hey @matiasleidemer, thanks for opening this pull request.

Can you help me understand why this is needed? I'm not sure I follow how this is beneficial in terms of the example you gave here. Why would you want to both `include` and `exclude` the same pattern here?

marcoroth (approved):
I see, the example is not representative.

I added two tests, one for Rust one for JavaScript each to make sure it works in all bindings, thank you! 🙏🏼

matiasleidemer:
Thank you, @marcoroth 😄

PR #1671, opened 2026-04-22
Linter: Implement `a11y-no-visually-hidden-interactive-elements` rule

### LINKED ISSUE #1225: Linter: Implement `no-visually-hidden-interactive-elements` rule
Implement the [`GitHub::Accessibility::NoVisuallyHiddenInteractiveElements`](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/no_visually_hidden_interactive_elements.rb) rule from [`erblint-github`](https://github.com/github/erblint-github).

### Rule Details

This rule guards against visually hiding interactive elements. If a sighted keyboard user navigates to an interactive element that is visually hidden they might become confused and assume that keyboard focus has been lost.

Note: `input` elements are not flagged at this time as some visually hidden inputs might cause false positives (e.g. file inputs).

### Examples

**Incorrect** code for this rule:

```html
<button class="sr-only">Submit</button>
```

**Correct** code for this rule:

```html
<h2 class="sr-only">Welcome to GitHub</h2>
```

### References

- [erblint-github source](https://github.com/github/erblint-github/blob/main/lib/erblint-github/linters/github/accessibility/no_visually_hidden_interactive_elements.rb)
- [erblint-github docs](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-visually-hidden-interactive-elements.md)

### Parent Issue

- #1205

joelhawksley:
Implements the `a11y-no-visually-hidden-interactive-elements` rule, inspired by erblint-github's `NoVisuallyHiddenInteractiveElements`.

This rule flags keyboard-focusable elements that use the `sr-only` class without a focus reveal class. Sighted keyboard users navigating to a visually hidden element may think keyboard focus has been lost.

The rule accounts for `href`, `disabled`, and `tabindex`, and allows reveal classes such as `focus:not-sr-only` and `focus-visible:not-sr-only`.

`input` elements are intentionally not flagged to avoid false positives from common patterns such as visually hidden file inputs and custom-styled form controls.

The current implementation recognizes `sr-only` and `not-sr-only` conventions like those Tailwind defines. The class names are not yet configurable; configurable rule options are planned as a follow-up.

The rule is disabled by default because the behavior of `sr-only` depends on the application’s CSS. Projects that reveal visually hidden elements globally on focus do not need to enable it.

Closes #1225


brunoprietog:
Wouldn't this rule incorrectly flag things like `sr-only focus:not-sr-only`?

joelhawksley:
@brunoprietog that's a good point! The original rule from rubocop-github did not account for those patterns. I went ahead and added tests and a fix and put your name on the commit ❤️

brunoprietog:
Great! ❤️

I'm not sure if it should be enabled by default, since we're assuming you'd be using Tailwind here, right?

Generally, when I define the class in my projects, I do it like this:

```css
.visually-hidden:not(:focus):not(:active):not(:focus-within) {
  clip-path: inset(50%);
  height: 1px;
  overflow: clip;
  position: absolute;
  white-space: nowrap;
  width: 1px;
}
```

And that way, you always avoid the problem. You could call it sr-only, too.

But if you use Bootstrap, it could also apply to `visually-hidden`, because the other version is `visually-hidden-focusable`.

Or maybe make the class name configurable in some way?

joelhawksley:
@brunoprietog forgive me if I'm not following, but I think we can the accommodation for Tailwind and not worry whether folks are using it or not. As for whether to allow for configuring other classes, I think that's a good idea but maybe @marcoroth can decide? I assume we'd just allow a list to be passed to the rule configuration.

brunoprietog:
Oh I'm sorry, I think I didn't explain myself clearly. What I meant was that this actually depends a lot on whether you're using Tailwind or not, in the sense that you could create your own `sr-only` class, which would be the same as the class I showed above with the `:not(:focus):not(:active):not(:focus-within)` pseudo classes, and in that case, it wouldn't really be a problem to apply that class to interactive elements.

Similarly, if we added the visually-hidden Bootstrap class to the list along with `sr-only`, it would be somewhat similar.

joelhawksley:
@brunoprietog I see what you mean now. Yes, this rule depends on what the class actually does in your application. I think it's fine to leave this as disabled by default, but maybe we should add some docs to clarify your concern. Would you be up for suggesting changes on this PR?

javascript/packages/linter/docs/rules/a11y-no-visually-hidden-interactive-elements.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Description
+
+Enforce that interactive elements are not visually hidden using classes like `sr-only`.
+
+## Rationale
+
+Visually hiding interactive elements can be confusing to sighted keyboard users as it appears their focus has been lost when they navigate to the hidden element.
+
+Note: `input` elements are not flagged at this time as some visually hidden inputs might cause false positives (e.g. file inputs).
```
brunoprietog:
```suggestion
Note: `input` elements are not flagged at this time as some visually hidden inputs might cause false positives (e.g. file inputs).

## When to enable

This rule is specific to Tailwind in practice: it detects Tailwind's `sr-only` class and its `*:not-sr-only` focus-reveal variants. It is disabled by default because whether `sr-only` is a problem depends on how the class is defined in CSS.

In Tailwind, `sr-only` keeps the element permanently hidden, so revealing it on focus has to be added per element (`focus:not-sr-only`), which is what this rule checks for. If instead you define your own `sr-only` that reveals on focus (e.g. `.sr-only:not(:focus):not(:active):not(:focus-within)`), then using it on interactive elements is not a problem and you can keep the rule disabled.

Class names also differ across frameworks (e.g. Bootstrap's `visually-hidden` / `visually-hidden-focusable`), which this rule does not currently handle.
```

javascript/packages/linter/src/rules/a11y-no-visually-hidden-interactive-elements.ts
```diff
@@ -0,0 +1,59 @@
+import { type ParseResult, type ParserOptions, type HTMLElementNode, getTagLocalName, getStaticAttributeValue } from "@herb-tools/core"
+
+import { BaseRuleVisitor } from "./rule-utils.js"
+import { ParserRule } from "../types.js"
+import type { UnboundLintOffense, LintContext, FullRuleConfig } from "../types.js"
+
+const INTERACTIVE_ELEMENTS = ["a", "button", "summary", "select", "option", "textarea"]
+
+const VISUALLY_HIDDEN_CLASSES = ["sr-only"]
+const VISUALLY_HIDDEN_UNDO_CLASSES = ["not-sr-only", "focus:not-sr-only", "focus-within:not-sr-only"]
```
brunoprietog: Since the rule already works on Tailwind's conventions, it might be better to handle its focus variants more completely. Instead of an explicit list, matching the semantics (any focus-triggered `not-sr-only`) would be more robust:

```ts
const VISUALLY_SHOWN_ON_FOCUS = /(?:^|:)(?:group-)?focus(?:-visible|-within)?:not-sr-only$/
const isUndoClass = (cls: string) => cls === "not-sr-only" || VISUALLY_SHOWN_ON_FOCUS.test(cls)
```

And then `!classes.some(isUndoClass)` below. This covers `focus`, `focus-visible`, `focus-within`, `group-focus-within`, and responsive prefixes like `md:focus:not-sr-only` at once. It still excludes `hover:` and `active:`, because those do not reveal the element for keyboard users, so the offense should still fire.

brunoprietog (commented):
I have been thinking about this more and I have some doubts I wanted to share.

Whether `sr-only` is a problem depends on the CSS, not the HTML. If the class is defined to reveal the element on focus (like the example I shared before, with `:not(:focus):not(:active):not(:focus-within)`), then using it on interactive elements is never a problem, no matter where it is used. So the real fix is to define the class correctly once in CSS, instead of checking every usage in the templates. If you control the CSS, this rule does not add much.

However, Tailwind is very widely used, and there `sr-only` keeps the element permanently hidden, and you have to reveal it per element (`focus:not-sr-only`). In that case the fix really is per usage, so the rule is useful.

But this also means it is a Tailwind-specific rule, not a general a11y one. It will not catch Bootstrap, plain CSS, or a custom `sr-only` that reveals on focus, while the name `a11y-no-visually-hidden-interactive-elements` looks general.

So I have two questions:

- Is it worth keeping, considering the CSS fix is better when you control the CSS? I think yes, mainly because Tailwind is so widely used.
- If we keep it, should it be scoped and named to make clear that it is really about Tailwind's `sr-only`, instead of looking like a general rule? There is already precedent for context-specific rules (`actionview-*`).

What do you think?

joelhawksley:
> * If we keep it, should it be scoped and named to make clear that it is really about Tailwind's `sr-only`, instead of looking like a general rule? There is already precedent for context-specific rules (`actionview-*`).
> 
> What do you think?

Funny you should mention this re: Tailwind, when we don't use it at GitHub! It just so happens that our implementation is similar in this regard.

What do you think about defaulting to sr-only and allowing the class to be configured?

brunoprietog:
Yes I suppose so, but is that possible?

Just out of curiosity, is there a reason why you haven't fixed that at the source in the CSS on GitHub instead of using the rule? Or maybe there's something I'm not taking into account

joelhawksley:
> Just out of curiosity, is there a reason why you haven't fixed that at the source in the CSS on GitHub instead of using the rule? Or maybe there's something I'm not taking into account

I'm on paternity leave at the moment, so I can't answer that question 😄 

@marcoroth how do you think we should proceed here?

javascript/packages/linter/src/rules/a11y-no-visually-hidden-interactive-elements.ts
marcoroth: Addressed in the latest update. I replaced the closed reveal-class list with the semantic matcher you suggested and added coverage for prefixed variants such as `md:focus:not-sr-only` and `dark:lg:focus-visible:not-sr-only`, while keeping `hover:` and `active:` as offenses. Thanks for the concrete suggestion!

javascript/packages/linter/docs/rules/a11y-no-visually-hidden-interactive-elements.md
marcoroth: Addressed in the latest update. The docs now explain when the rule applies, show the focus-aware CSS alternative, call out the current class-name limitations, and note that configurable hidden and reveal classes are planned as a follow-up. Thanks for helping clarify the scope here!

marcoroth:
@brunoprietog @joelhawksley I revisited the rule based on your discussion and incorporated the feedback in [this follow-up commit](https://github.com/marcoroth/herb/pull/1671/changes/838de3c99126c12ec2c0e2fe5e5feaee96b17ab4).

The rule remains disabled by default and the docs now make its scope explicit. It recognizes `sr-only` conventions like those Tailwind defines, cannot infer the application's CSS, and is unnecessary when the visually hidden class already reveals focused elements globally. Configurable hidden and reveal class names remain a follow-up for #1204.

The implementation now checks actual keyboard focusability instead of a fixed tag list. It accounts for `href`, `disabled`, and `tabindex`, supports custom focusable elements, no longer reports `<option>` independently, and continues to exclude inputs.

Following [Bruno's suggestion](https://github.com/marcoroth/herb/pull/1671#discussion_r3378971897), focus reveal classes are matched semantically. Prefixed variants such as `md:focus:not-sr-only` work, while `hover:` and `active:` still report an offense. Dynamic ERB class values remain conservatively skipped.

The offense message now identifies the tag and ends with an actionable fix.

Thank you! The feedback really clarified both the scope and behavior of the rule. 🙏🏼

brunoprietog:
Amazing! ❤️

Issue #1670, opened 2026-04-22
html-no-unescaped-entities triggered on (content_)tag helper


bdewater-thatch:
Running Herb 0.9.6 via Yarn I noticed https://herb-tools.dev/linter/rules/html-no-unescaped-entities was triggered on this ERB:
```erb
<%= content_tag :p, "2025 & On" %>
```

[`content_tag`](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag) and `tag` have default `escape = true` arguments so this shouldn't get flagged.

bdewater-thatch:
Err, I think my LLM hallucinated this I had it work auto-fixing all of our disabled linters. Sorry for the noise!

PR #1672, opened 2026-04-22
Linter: Fix `html-no-unescaped-entities` false positive on ActionView tag helpers

bdewater-thatch:
https://github.com/marcoroth/herb/issues/1670 was actually correct, the agent hallucinated the rule being enabled or not 😂 

The html-no-unescaped-entities rule flags `&` characters inside string arguments to ActionView tag helpers that auto-escape their content by default. Following the suggested fix (`&amp;`) causes double-escaping and the user sees literal `&amp;` instead of `&`.

Decided to let it take a stab at fixing it because I don't want people to get too comfortable with disabling this rule.


bdewater-thatch:
@marcoroth rebased and tests should be fixed now, could you kick off another CI run please?

javascript/packages/linter/src/rules/html-no-unescaped-entities.ts
marcoroth, line 162: Hey @bdewater-thatch, thanks for taking a stab at fixing this!

I'd need to check, but couldn't we just disable the `action_view_tag_helpers` option on this linter rule, so that the linter doesn't even see the converted ActionView helpers as HTML nodes?

Or is the automatic escaping only happening on some of the Action View helpers?

marcoroth: Actually, I wonder if we should handle this in the transformation logic in the parser itself where we translate the Action View Helpers to the raw HTML nodes to directly escape what needs to be escaped.

bdewater-thatch: Thanks for the comments @marcoroth - let me explore that direction. Automatic escaping is conditional on arguments passed in or whether the block form is used.

PR #1678, opened 2026-04-23
Config: Warn when `.herb.yml` instead of `.herb.yml` is found

krystof-k:
Love the work you're doing on Herb — it's a great tool and I'm enjoying using it!

I ran into a small friction point: when I created `.herb.yaml` instead of `.herb.yml`, it didn't work — it was actually actively rejected with an error asking me to rename it. Since `.yaml` is [the official YAML extension](https://yaml.com/resources/faq/), it should be the supported (and preferred) variant.

This PR:

- Removes the .herb.yaml rejection
- Supports both .herb.yaml and .herb.yml across all three runtimes (JS, Ruby, Rust)
- Prefers .herb.yaml over .herb.yml if both exist (since it's the official extension)
- Updates all user-facing strings, docs, READMEs and `--init`
- Fully backwards compatible — existing .herb.yml files continue to work without any changes
- Tests updated across all three languages, including precedence tests

Hope this absolutely unimportant but preventing-many-confusions change will be helpful!


marcoroth (commented):
Hey @krystof-k, thanks for the pull request.

I have been following the convention established by most Ruby tools and configuration files that also Rails ships with, which all use `.yml`. 

So I'm not sure if we should be using `.yaml`, as that might be unusual and unexpected for most users.

krystof-k:
Hi @marcoroth, so at least someone needs to have it right ;)

Jokes aside, fair point — I understand the Ruby/Rails convention around `.yml`. My thinking here was less about forcing a switch and more about reducing friction: `.yaml` is the official YAML extension and some people (like me) reach for it first and are surprised it doesn't work. Supporting both feels like the least surprising behavior, while keeping full backwards compatibility with existing `.yml` files.

Whether Herb should prefer `.yaml` in generated examples/docs is totally your call. I intentionally kept those changes isolated in a separate commit so they can be reverted easily.

sethhorsley:
@krystof-k isn't `yml` the default in rails why would you need to change this here?

marcoroth:
Hey @krystof-k, thanks for proposing this change. 

I don't think I want to land this pull request as-is. To keep it simple, I think we should never read and support both extensions as the same time. But I'm happy to make the changes so that any/or all of the CLIs warn you that you have `.herb.yaml` defined instead of `.herb.yml` to make it easier to discover that the file should be `.yml`. Does that make sense?

marcoroth:
Superseded by #1996

PR #1711, opened 2026-05-05
Add sig/manifest.yaml to declare stdlib dependencies

tk0miya:
`sig/herb/token_list.rbs` declares a class inheriting from `SimpleDelegator`, which comes from the `delegate` standard library. Since `delegate` is a default gem, it does not appear in the gemspec, and `rbs collection install` cannot resolve it on the consumer side.

With Steep 2.0, this previously-silent issue surfaces as a hard error:

    Type checking failed due to error in library RBS file
    Cannot find type `SimpleDelegator`
    (.../herb-0.10.1/sig/herb/token_list.rbs:4:2)
    Diagnostic ID: Ruby::LibraryRBSError

Declare the implicit standard library dependencies in `sig/manifest.yaml` so downstream users get correct type resolution.

Refs: https://github.com/ruby/rbs/blob/master/docs/gem.md

### DIFF
--- a/herb.gemspec
+++ b/herb.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |spec|
     "lib/**/*.rb",
     "lib/**/*.yml",
     "sig/**/*.rbs",
+    "sig/manifest.yaml",
     "src/**/*.{c,h}",
     "ext/**/*.{c,h}",
     "templates/**/*.{rb,erb}",
--- a/sig/manifest.yaml
+++ b/sig/manifest.yaml
@@ -1,0 +1,13 @@
+dependencies:
+  - name: delegate
+  - name: did_you_mean
+  - name: fileutils
+  - name: json
+  - name: optparse
+  - name: pathname
+  - name: socket
+  - name: stringio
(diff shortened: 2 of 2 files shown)


marcoroth (commented):
Thank you @tk0miya! 🙏🏼 

Is there a way we can check in CI that Herb is bundling/exposing the RBS signatures properly so we don't introduce any regressions?

tk0miya:
@marcoroth we can check the validity of types via `rbs validate` command. As a trial, I implemented the validation on https://github.com/marcoroth/herb/compare/main...tk0miya:herb:rbs_validate?expand=1.

At present, some `serialized_*` type aliases are not defined. So we need to define them before validating them on CI.

marcoroth:
Awesome, thank you for investigating @tk0miya!


If we move the `bundle exec rbs validate` command to the existing `build.yml` we should also have the `serialized_* ` types available as they are getting generated automatically from the ERB files in `templates/` via the `rake templates` command here: https://github.com/marcoroth/herb/blob/main/templates/sig/serialized_ast_nodes.rbs.erb

PR #1745, opened 2026-05-23
skills: Add Herb install and ERB lint/format skills for LLM agents

lucianghinda:
Adds two skills under skills/ that help LLM agents work with Herb in Rails projects. 

- skills/herb-install-rails/ — idempotent install/configure flow with bin/herb-rails-doctor (preflight), bin/herb-rails-setup (Gemfile + .herb.yml + CI workflow with both Ruby and Node setup, plus uninstall), and bin/herb-rails-write-config.rb (deep-merge writer that preserves user-set keys, Ruby 2.6+ compatible).

- skills/herb-erb-lint-format/ — inner-loop contract for editing ERB: lint + actionview check + format after every edit, with explicit npx fallback when the gem binary is missing and a note that actionview check has no npm equivalent.

### DIFF
--- a/skills/herb-erb-lint-format/SKILL.md
+++ b/skills/herb-erb-lint-format/SKILL.md
@@ -1,0 +1,206 @@
+---
+name: herb-erb-lint-format
+description: Use when editing, creating, refactoring, renaming, or deleting `.html.erb`, `.turbo_stream.erb`, `.html.herb`, or `.rhtml` files in a Rails project that has a `.herb.yml` at the root. Also use when auditing Rails views for accessibility, security, or broken partial references, or when writing template codemods.
+---
+
+# Lint and format ERB with Herb
+
+## Overview
+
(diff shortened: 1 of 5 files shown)


lucianghinda:
@marcoroth here is a draft of adding two skills to Herb following our conversation in Rimini. I split the idea of skills in two: 
1. One script to help setup Herb for a Rails app - this one has some scripts inside
2. One script that should be invoked by any LLM (with skill support) and that teaches the LLM how to use Herb while implementing features. 

I marked this as draft because I need a bit of feedback from you in a couple of directions: 
1. In case you agree with the idea of a skill that teaches an LLM to install herb that has few additional scripts and here I have a question if those scripts should be inside the skill or be part of Herb itself. 
2. In case the scripts for installing Herb remains  remain part of the skill I need to write some tests for them and so the question will be where to put those tests in the `test` folder (which path)? 
3. Of course any other feedback about the wording of the skill will be helpful

marcoroth:
Thanks for the work here @lucianghinda! 

I think that we might want to wait a bit until we get a more stable ground to derive these skills from and also see how they end up being exposed in Rails 8.2+.

PR #1753, opened 2026-05-23
Linter: Fix `html-no-self-closing` to preserve indentation

fgkolf:
:wave:  I came across the following issue with the `html-no-self-closing` rule:

When autofixing nested elements whose closing tag appears on a new line, the self-closing tag is correctly fixed, but the indentation is lost and the closing bracket is placed at column 0.

This can be verified with the following example:
#### Before
```
<figure>
  <img
    src="/image.png"
    alt="An image"
  />
</figure>
``` 

#### After
```
<figure>
  <img
    src="/image.png"
    alt="An image"
>
</figure>
``` 

#### With this fix
```
<figure>
  <img
    src="/image.png"
    alt="An image"
  >
</figure>
``` 

P.S. I wasn't sure whether I should open an issue first. Let me know if you'd prefer that for completeness' sake.

### DIFF
--- a/javascript/packages/linter/src/rules/html-no-self-closing.ts
+++ b/javascript/packages/linter/src/rules/html-no-self-closing.ts
@@ -55,6 +55,13 @@ export class HTMLNoSelfClosingRule extends ParserRule<NoSelfClosingAutofixContex
     return { action_view_helpers: true }
   }
 
+  private isIndentation(precedingNode: Node | null, node: Node): boolean {
+    return !!precedingNode &&
+      isWhitespaceNode(node) &&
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Awesome, thanks for this great catch @fgkolf! 🙏🏼 

> P.S. I wasn't sure whether I should open an issue first. Let me know if you'd prefer that for completeness' sake.

Whatever is more comfortable for you, but you definitely don't need to open an issue first just for the sake of opening an issue if you already have a proposed solution for the fix, thank you! 🙌🏼

PR #1773, opened 2026-06-03
Docs: Updates custom linter rules documentation

jkogara:
- Custom linter rule examples use `this.name` when constructing visitors, but `ParserRule` exposes the rule name via `this.ruleName`. 
- Using `this.name` results in undefined being passed to the visitor, causing the rule name to show as undefined in CLI output and breaking (at least my nvim) LSP diagnostics.


marcoroth (approved):
Thanks so much for catching this @jkogara! 🙏🏼

PR #1776, opened 2026-06-05
Language Server: Fix comments inside multiline ERB tags

### LINKED ISSUE #1774: Commenting inside a ruby tag in an erb file is broken (comments as HTML)
I have this weird case where HERB comments the Ruby inside a Ruby tag as HTML instead of Ruby.

```erb
<%
  some_ruby_inside_erb
%>
<%= link_to some_ruby_inside_erb %>
```

https://github.com/user-attachments/assets/319685bc-4965-4496-a60b-f159b1b0a8f6

RSeven:
### Summary
This fixes comment handling for Ruby lines inside multiline ERB tags.

### Issues resolved
Before, commenting the internal Ruby line with **toggleLineComment** produced HTML comments inside Ruby:

```erb
<%
  <!-- a -->
%>
```

After:
```erb
<%
  # a
%>
```
___
Before, block-commenting only the internal Ruby line with **toggleBlockComment** inserted nested ERB tags:

```erb
<%
  <% if false %>
  a
  <% end %>
%>

```

After:

```erb
<%
  # a
%>

```

Closes #1774

### DIFF
--- a/javascript/packages/language-server/src/comment_ast_utils.ts
+++ b/javascript/packages/language-server/src/comment_ast_utils.ts
@@ -70,6 +70,12 @@ export function determineStrategy(erbNodes: ERBContentNode[], lineText: string):
     if (!node.tag_opening || !node.tag_closing) return "html-only"
 
     const nodeStart = node.tag_opening.location.start.column
+    const spansMultipleLines = node.tag_closing.location.end.line > node.tag_opening.location.start.line
+
+    if (spansMultipleLines) {
+      return lineText.substring(0, nodeStart).trim() === "" ? "single-erb" : "whole-line"
+    }
+
     const nodeEnd = node.tag_closing.location.end.column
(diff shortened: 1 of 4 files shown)


RSeven:
Not sure about how to handle toggleBlockComment for multiline ERB tags. I thought about keeping the if block but it creates some strange behaviours, for example:

1) Calling toggleBlockComment on the same selection twice doesn't really toggle the comment off, it just generates a new if block:
```erb
<%
  if false
    if false
      a
      b
    end
  end
%>
```

2. Calling toggleBlockComment on a selection that includes the `if` clause and not the `end` will result in the following code that, even being valid, is probably not the expected behaviour:
```erb
<%
  if false
    if false
      a
  end
    b
  end
%>
```

marcoroth (approved):
Thank you @RSeven! 🙏🏼

PR #1821, opened 2026-06-25
Parser: Allow inline `in` expressions outside of `case` statements

### LINKED ISSUE #1758: Parser: `ERB_CONTROL_FLOW_SCOPE_ERROR` when using pattern matching for conditional display
Below is an example of using pattern matching to detect the nature of an object and conditionally display it

```erb
  <% if @object.attribute in /...+/ # Not blank, not "NA", more than 3 chars %>
    <%= @object.attribute %>
  <% end %>
```

This is being flagged with:

```
`<% in %>` appears outside its control flow block. Keep ERB control flow statements together within the same HTML scope (tag, attribute, or content). (Herb Parser [ERB_CONTROL_FLOW_SCOPE_ERROR])
```
and
```
`<% end %>` appears outside its control flow block. Keep ERB control flow statements together within the same HTML scope (tag, attribute, or content). (Herb Parser [ERB_CONTROL_FLOW_SCOPE_ERROR])
```

I've tried including parenthesis in various places to no avail. I also don't really understand what this error is looking to highlight. As far as I can tell, it is in the same HTML scope. I suspect the keyword `in` is throwing it off. When I remove keyword `in` and use `=~` for example, the linting error goes away.

All cases render as I'd expect without errors.

ttilberg:
Currently using `value in pattern` without an accompanied case statement raises `ERBControlFlowScopeError`.

```erb
<% if value in pattern %>
  True
<% end %>
```

```
`<% in %>` appears outside its control flow block. Keep ERB control flow statements together within the same HTML scope (tag, attribute, or content). (Herb Parser [ERB_CONTROL_FLOW_SCOPE_ERROR])
```

Now a pattern match predicate only increments in_node_count when it's accompanied by a case statement. 

I used Claude to help me understand how to make this change -- so please look it over and determine if this is the correct type of fix.

Fixes #1758

### DIFF
--- a/examples/in_as_expression.html.erb
+++ b/examples/in_as_expression.html.erb
@@ -1,0 +1,7 @@
+<h1>
+  <% if 1 in Numeric %>
+    <div>Text1</div>
+  <% else %>
+    <div>Text2</div>
+  <% end %>
+</h1>
--- a/src/analyze/analyze_helpers.c
+++ b/src/analyze/analyze_helpers.c
@@ -277,7 +277,13 @@ bool search_in_nodes(const pm_node_t* node, void* data) {
   analyzed_ruby_T* analyzed = (analyzed_ruby_T*) data;
 
   if (node->type == PM_IN_NODE) { analyzed->in_node_count++; }
-  if (node->type == PM_MATCH_PREDICATE_NODE) { analyzed->in_node_count++; }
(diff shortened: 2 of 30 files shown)


marcoroth (approved):
Awesome catch, thank you so much @ttilberg! 🙏🏼

PR #1825, opened 2026-06-27
Linter: Add WAI-ARIA Graphics roles to `html-aria-role-must-be-valid`

### LINKED ISSUE #1726: Linter Rule: `html-aria-role-must-be-valid` missing valid roles
I think the mdn list for roles or something like that was used, but this means that valid roles such as `graphics-symbol` are marked as incorrect.

See: https://www.w3.org/TR/graphics-aria-1.0/#roles

There could be more valid specs that are missed.

miyatis:
This pull request adds the three roles from the [W3C WAI-ARIA Graphics Module 1.0](https://www.w3.org/TR/graphics-aria-1.0/#roles) spec — `graphics-document`, `graphics-object`, and `graphics-symbol` — to the `VALID_ARIA_ROLES` set so that `html-aria-role-must-be-valid` no longer flags them as unrecognized.

Closes #1726


marcoroth (approved):
Thank you @miyatis! 🙏🏼

Issue #928, opened 2025-11-26
Formatter: Dangling > and odd formatting for html tags with lengthy / multiple attributes


ron-shinall:
I'm not sure if the formatting is intentional, but I wanted to document it just in case. Formatting html tags with lengthy and/or multiple attributes produces unexpected results.

**Input:**
```erb
<div class="bg-white border-l-4 shadow-lg rounded-lg p-6 mb-6" role="alert" style="border-left-color: #B9D400; background-color: #F0F8E8;">
```

**Output:**
```erb
<div
  class="bg-white border-l-4 shadow-lg rounded-lg p-6 mb-6"
  role="alert"
  style="border-left-color: #B9D400; background-color: #F0F8E8;"
>
```

**Expected:**
```erb
<div class="bg-white border-l-4 shadow-lg rounded-lg p-6 mb-6" role="alert" style="border-left-color: #B9D400; background-color: #F0F8E8;">
```

...or possibly:
```erb
<div class="bg-white border-l-4 shadow-lg rounded-lg p-6 mb-6" role="alert" 
     style="border-left-color: #B9D400; background-color: #F0F8E8;">
```

marcoroth:
Hey @ron-shinall, at this current stage this is intentional. 

[...]

PR #1834, opened 2026-06-28
Formatter: Wrap attributes based on `maxLineLength` only

andy-rootly:
## Summary

- Removes the "wrap when more than 3 attributes" rule from [`attribute-renderer.ts:91`](https://github.com/marcoroth/herb/blob/main/javascript/packages/formatter/src/attribute-renderer.ts#L91). Wrapping is now driven purely by `maxLineLength`.
- Updates tests and the `maxLineLength` doc note added in #1823.

Follow-up to #1823 / #928. The count heuristic was surprising — `<div a="1" b="2" c="3" d="4">` wrapped at 36 chars under the 80-char default, with no way for the user to opt out. After this PR, the same tag stays inline; long tags still wrap.

## Test plan

- [x] CI passes (full formatter test suite — not run locally)
- [x] Existing snapshot fixtures still match, or are updated where the 4-attribute rule was load-bearing


marcoroth (approved):
Thank for this pull request, @andy-rootly!  I'll keep this one setting here for a little bit more so we can bring this on main for the next bigger Herb V0.11.0 release.

Issue #1393, opened 2026-03-16
Linter: script type module and importmap incorrectly marked as wrong


tvdeyen:
The linter marks `<script type="module">` and `<script type="importmap">` as incorrect syntax. 

<img width="1744" height="254" alt="Image" src="https://github.com/user-attachments/assets/d771f7d8-bfa2-4a09-9a0a-f3aded17dece" />

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type

marcoroth:
Very good point, thanks for reporting!

This rule was ported 1:1 from `erb_lint` and I guess that was written at a time where `type="module"` and `type="importmap"` weren't really popular or a thing yet.

We should expand the list to also allow `module`, `importmap`, and even `speculationrules` 👍🏼

marcoroth:
Opened https://github.com/marcoroth/herb/pull/1413 to address this 🙌🏼

PR #1844, opened 2026-07-08
Linter: Allow `application/ld+json` for `html-allowed-script-type` rule

xymbol:
`application/ld+json` is the standard mechanism for embedding JSON-LD structured data, and the HTML specification treats it as an inert data block. Same fix as #1393 / #1413 made for `module` and `importmap`.

https://html.spec.whatwg.org/multipage/scripting.html#data-block


marcoroth (approved):
Great catch, thank you @xymbol! 🙏🏼

PR #1845, opened 2026-07-08
Linter: Treat `breadcrumb` as a side-effecting helper in `erb-no-unused-expressions`

6temes:
## What

Adds `breadcrumb` to `SIDE_EFFECT_METHODS` in `erb-no-unused-expressions`, so `<% breadcrumb :key %>` calls are no longer reported as unused expressions.

## Why

[gretel](https://github.com/lassebunk/gretel) — a widely used breadcrumbs gem — exposes `breadcrumb` as a view helper that is called purely for its side effect. It sets the current breadcrumb trail and is meant to be used in a silent tag:

```erb
<% breadcrumb :product, @product %>
```

Under the hood it just assigns the renderer and returns it:

```ruby
def breadcrumb(key = nil, *args)
  @_gretel_renderer = Gretel::Renderer.new(self, key, *args)
end
```

This is the same shape as `content_for`, `provide`, and the `turbo_*` helpers already on the allowlist: a bare, receiver-less call used for a side effect, whose return value is intentionally discarded (you can't meaningfully `<%= %>` it — that would print a `Gretel::Renderer`). Today the rule flags every `<% breadcrumb %>`, so an app that renders breadcrumbs on most pages gets one offense per view, which in practice forces disabling the rule project-wide.

## Verification

- Added a `valid cases` test mirroring the existing `provide` / Turbo tests.
- Behaviorally confirmed against the published `0.10.1` build: with `breadcrumb` allowlisted, `<% breadcrumb :key %>` passes while a genuine `<% @foo %>` is still reported.

I wasn't able to run the full JS test suite locally (the workspace build pulls in the native/Emscripten toolchain), so I'd appreciate CI confirming the added test.

### DIFF
--- a/javascript/packages/linter/src/rules/erb-no-unused-expressions.ts
+++ b/javascript/packages/linter/src/rules/erb-no-unused-expressions.ts
@@ -27,6 +27,7 @@ const MUTATION_METHODS = new Set([
 ])
 
 const SIDE_EFFECT_METHODS = new Set([
+  "breadcrumb",
   "content_for",
   "provide",
   "flush",
--- a/javascript/packages/linter/test/rules/erb-no-unused-expressions.test.ts
+++ b/javascript/packages/linter/test/rules/erb-no-unused-expressions.test.ts
@@ -131,6 +131,13 @@ describe("ERBNoUnusedExpressionsRule", () => {
       `)
     })
 
+    test("passes for breadcrumb", () => {
+      expectNoOffenses(dedent`
+        <% breadcrumb :page %>
+        <% breadcrumb :product, @product %>
+      `)
+    })
+
     test("passes for assert_valid_keys", () => {
       expectNoOffenses(dedent`
         <%


brunoprietog:
`erb-no-unused-expressions` has the potential to become a massive set of allowed expressions, and it actually ignores the fact that you yourself could create helpers that modify instance variables or use the `provide` helper under the hood. I said that in [this comment](https://github.com/marcoroth/herb/pull/1531#issuecomment-4149811743), using Turbo as an example, but I don't think adding a fixed list of allowed expressions is really the way to go. Having this rule enabled by default basically means prohibiting these kinds of custom helpers, which are sometimes quite useful for dynamically enabling or disabling things, or changing a view’s behavior using natural language. Not all helpers need to use the <%= syntax. What do you think?

6temes:
@brunoprietog you're right that a fixed allowlist doesn't scale — and more importantly, it can't cover app-defined helpers at all. If someone wraps `provide` in their own `set_page_title` helper, or uses a gem helper that isn't on the list (this PR's `breadcrumb` being one), the rule flags every call and there's no recourse short of disabling it project-wide. That's the same concern you raised on #1531, and hardcoding each new case just defers it.

What if the defaults stayed limited to the **framework-owned** helpers — Rails' `content_for`/`provide` and the `turbo_*` helpers — and everything app- or gem-specific was **user-configured** instead?

```yaml
erb-no-unused-expressions:
  allowedMethods:
    - breadcrumb
    - set_page_title
```

That draws the line where you're pointing: herb ships the helpers it can reasonably know about (Rails + Turbo), and each app allowlists its own side-effecting helpers — gretel's `breadcrumb`, custom helpers, anything. The default list stops needing to grow with every gem, and custom helpers become a one-line config instead of an upstream PR. It's also the shape ESLint's `no-unused-expressions` and most RuboCop cops take.

One thing I noticed while looking into this: no rule in herb currently reads a custom config option — the per-rule surface is just `enabled`/`severity`/`include`/`only`/`exclude`, and `RuleConfigBaseSchema` isn't `passthrough`, so an unknown key gets stripped today. So this would be the first per-rule option, which is a convention worth you deciding on (naming, additive-vs-replace, schema vs passthrough). Happy to implement it whichever way you'd prefer.

marcoroth:
Hey @6temes, thanks for opening this pull request. I agree, that we shouldn't be adding more of these gem-specific helpers to the ignore list.

That being said, I think we should have it be part of the configuration. Sadly, we don't support specific linter rules configuration yet. But we have this tracked in https://github.com/marcoroth/herb/issues/1204.

6temes:
Thanks, @marcoroth. 

I will wait for this issue to get solved, then.

Issue #574, opened 2025-10-04
Formatter: Need a way to ignore the warnings or surpress it


mnkhod:
**Input:**
```erb
    <ul class="grid gap-y-3 font-[Pixelify_Sans]">
      <%= render "nav_link", title: "Dance of Ombres" %>
      <%= render "nav_link", title: "Invitations" %>
      <%= render "nav_link", title: "Leaderboard" %>
      <%= render "nav_link", title: "Tasks" %>
         <%= render "nav_link", title: "Market" %>
    </ul>
```

**Output:**
```erb
    ⚠️ Experimental Preview: The formatter is in early development. Please
    report any unexpected behavior or bugs to
    https://github.com/marcoroth/herb/issues/new?template=formatting-issue.md
    <ul class="grid gap-y-3 font-[Pixelify_Sans]">
      <%= render "nav_link", title: "Dance of Ombres" %>
      <%= render "nav_link", title: "Invitations" %>
      <%= render "nav_link", title: "Leaderboard" %>
      <%= render "nav_link", title: "Tasks" %>
      <%= render "nav_link", title: "Market" %>
    </ul>
```

**Expected:**
```erb
    <ul class="grid gap-y-3 font-[Pixelify_Sans]">
      <%= render "nav_link", title: "Dance of Ombres" %>
      <%= render "nav_link", title: "Invitations" %>
      <%= render "nav_link", title: "Leaderboard" %>
      <%= render "nav_link", title: "Tasks" %>
      <%= render "nav_link", title: "Market" %>
    </ul>
```

**Playground:**

**Additional context:**
Using conform nvim hence my configuration is
```lua
formatters = {
  injected = { options = { ignore_errors = true } },
  herb_format = {
    command = "herb-format",
    --  --indent-width 2
    append_args = { "--max-line-length", "200" },
  },
},
```

marcoroth:
Hey @mnkhod, thanks for reporting!

Oh, I guess the formatter CLI wasn't intended to be used like this! The idea would be to use the [Herb Language Server](https://herb-tools.dev/projects/language-server#herb-language-server) and then let the language server handle the `textDocument/formatting` request.

But I guess we could print the `Experimental Preview` warning non on stdout so it would also work the use it.

mnkhod:
[...]

PR #1851, opened 2026-07-12
Formatter: Fix disabled formatter stdin output

hschne:
Came across this with my Neovim setup in the Rubyevents repo, where the formatter is disabled via `.herb.yml`.

Editing any erb file results in the buffer being replaced with:
```
Formatter is disabled in .herb.yml configuration.
To enable formatting, set formatter.enabled: true in .herb.yml
Or use --force to format anyway.
``` 

The relevant configuration, in case you're interested, is: 

```lua
return {
  "stevearc/conform.nvim",
  opts = {
    formatters = {
      herb_format = {
        command = "herb-format",
        stdin = true,
      },
    },
  },
}
```

This is the same class of bug as #574, which moved the experimental-preview warning to stderr for exactly this reason.

 ## Fix

 - Print the disabled notice to stderr instead of stdout.
 - When input is coming from stdin (- or piped with no file arg), echo the original source back to stdout so the formatter is a clean no-op.
 - Exit code stays 0 — a disabled formatter is an intentional no-op, not an error.

 ## Testing

 - yarn workspace @herb-tools/formatter test --run test/cli.test.ts — 67 passing, including a new case asserting stdin is returned verbatim and the notice lands on stderr.
 - Manual: printf '<div>  x </div>\n' | node bin/herb-format - in a project with formatter.enabled: false now prints the input unchanged on stdout, notice on stderr.

### DIFF
--- a/javascript/packages/formatter/src/cli.ts
+++ b/javascript/packages/formatter/src/cli.ts
@@ -156,6 +156,7 @@ export class CLI {
       this.determineProjectPath(positionals)
 
       const file = positionals[0]
+      const isUsingStdin = (!file && !process.stdin.isTTY) || file === "-"
 
       if (isInitMode) {
         const configPath = configFile || this.projectPath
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Great catch, thanks for fixing this @hschne! 🙏🏼

Issue #564, opened 2025-10-01
Formatter: duplicated lines in multilines context


colinux:
Hi, thanks for Herb !

**Input:**
```erb
<span>
  Par défaut, un dossier déposé peut être complété
  ou corrigé par le demandeur jusqu'à sa mise en instruction.
  <br>
  Dans une démarche déclarative, une fois déposé, un dossier ne peut
  plus être modifié. Soit il passe immédiatement en instruction
</span>

```

**Output:**
```erb
<span>
  Par défaut, un dossier déposé peut être complété ou corrigé par le demandeur jusqu'à sa mise en instruction.
  <br>
  Par défaut, un dossier déposé peut être complété ou corrigé par le demandeur
  jusqu'à sa mise en instruction.
  <br>
  Dans une démarche déclarative, une fois déposé, un dossier ne peut plus être
  modifié. Soit il passe immédiatement en instruction
</span>

```

**Expected:**
text not duplicated

**Additional context:**

Seems related with a combination of text length, multilines, and presence of <br>
With multiple `<br>` in input, several duplications occurs.

Version 0.7.0 & 0.7.4 are affected.

jbfeldis:
Hello,

we are having the same issue (also probably the same here : https://github.com/marcoroth/herb/issues/436#issuecomment-3356738094).

Here's a simple playground with a paragraph that does not check and duplicate on every format run : https://herb-tools.dev/playground?tab=format#DwBwfAygFg9grgGwCYAIB2MAuKBGBTFJOEBASwGMBDTPJAOmBwCcUB6MAFTwA9sB3UpigpKKUmjJoCeBHgC2eNJmCioTPADMAvACI5lUgkwwAXAGcYCoeIDmAAQtWotuuUs6wjvNbT2vPm1dLYFZKMBE0VC8UORh1FBpeEQ0aJhDwAChQSFhEVAxsfEJiMioaekYWdi4kgSERMQlxaVkFJRUUNU1dSgAvPCZMAE84UhgQAEczJA0Pf2dfB0tvBcC3ORCwiKjlmLiCROxKFIH0sCzwaHhkdCxcAiISCmpaBmY2Th5+QWFRcUkWvJFMpVOptDo+gNhqNxlMZjY5ssAksnC51ptwpRIihorF4odkqkzkA

It looks like the parsing is correct but the formatted output is not. 
[...]

Issue #436, opened 2025-08-25
Formatter: Duplicating content multiple times


sudhanshug16:
**Input:**
```erb
<!DOCTYPE html>
<html xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office" data-theme="light">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="x-ua-compatible" content="ie=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta name="format-detection" content="telephone=no, date=no, address=no, email=no, url=no">
    <meta name="x-apple-disable-message-reformatting">
    <title><%= message.subject %> | NoScrubs</title>
    <%= stylesheet_link_tag "mailer", media: "all" %>
    <style>
      @import url('https://fonts.googleapis.com/css2?family=DM+Sans:wght@400;500;600;700&display=swap');
    </style>
    <style>
      @media only screen and (max-width: 600px) {
        .hidden-mobile {
          display: none !important;
        }
      }
      @media only screen and (min-width: 601px) {
        .hidden-desktop {
          display: none !important;
        }
      }
    </style>
  </head>
  <body  style="font-family: 'DM Sans', Arial, sans-serif;">
    <% if content_for?(:preheader) %>
      <div class="hidden">
        <%= yield :preheader %>
      </div>
    <% end %>
    <div role="article" aria-roledescription="email" aria-label="<%= message.subject %>" lang="en">
      <table border="0" cellpadding="0" cellspacing="0" width="100%" class="bg-background leading-normal text-base">
        <tr>
          <td height="32"></td>
        </tr>
        <tr>
          <td align="center">
            <table border="0" cellpadding="0" cellspacing="0" class="w-full max-w-[684px] px-0 sm:w-[684px] sm-px-2" width="100%">
              <tr>
                <td>
                  <div class="shadow-xl border border-border-secondary rounded-lg overflow-hidden">
                    <div class="space-y-2 bg-primary-foreground p-4"><% if content_for?(:header) %>
                        <%= yield :header %>
                      <% else %>
                        <%= hosted_image_tag('mailer/header-logo.png', class: 'h-[35px] block mb-3') %>
                      <% end %>
                      <%= yield %>
                    </div>
                    <div class="p-4 bg-primary-foreground text-center border-t border-border-secondary">
                      <%= hosted_image_tag('mailer/footer-logo.png', class: 'h-[48px] mb-1') %>
[...]

PR #1863, opened 2026-07-20
Formatter: Don't add or remove whitespace at glued content boundaries

### LINKED ISSUE #469: Formatter: inserts whitespace around interpolations and elements
**Input:**
```erb
<%= @user.translated_greeting %>,<br>
```

**Output:**
```erb
<%= @user.translated_greeting %>
,
<br>
```

**Expected:**
```erb
<%= @user.translated_greeting %>,<br>
```

**Playground:**

https://herb-tools.dev/playground?tab=format#DwUgvABAAgrgzgUwE4DoBmSCWCB2ATAGwE8B9AcyQQQBdMcyIQA+AGmACMkmg

### LINKED ISSUE #1729: Formatter: adds newlines that produce undesired whitespace
The formatter will sometimes add newlines that cause undesired whitespace in the rendered HTML.

### Examples

Here's a pretty simple one:

**Input:**
```erb
<p>
  Hello
  <% if user %>
    <%= user.name %>'s dog
  <% end %>
</p>
```

**Output:**
```erb
<p>
  Hello
  <% if user %>
    <%= user.name %>
    's dog
  <% end %>
</p>
```

**Expected:** (no change)
```erb
<p>
  Hello
  <% if user %>
    <%= user.name %>'s dog
  <% end %>
</p>
```

This renders as
```
Hello John 's dog
```

with an extra space before the `'`, rather than the expected

```
Hello John's dog
```

Here's a slightly more complicated one:

**Input:**
```erb
<p>
  Hello<% if owner %> <%= owner.name %>'s dog<% end %>!
  It's time for your walk.
</p>
```

**Output:**

```erb
<p>
  Hello
  <% if owner %>
    <%= owner.name %>
    's dog
  <% end %>!
  It's time for your walk.
</p>
```

**Expected:**

No change, though not picky about it as long as it doesn't change what gets rendered.

When `owner` is present, it renders as

```
Hello John 's dog ! It's time for your walk.
```

with extra spaces before the `'` and `!`, rather than

```
Hello John's dog! It's time for your walk.
```

When `owner` is falsy, it renders as

```
Hello ! It's time for your walk.
```
with an extra space before the `!`, instead of

```
Hello! It's time for your walk.
```

**Playground:**

https://herb-tools.dev/playground/#DwBwfAUABFASCmAbRB7aVgFIoEsBmUArgM7wBOUmkMMWAvEaWQHQB2AhgLbyVgDkxKABMUAc3RYo8VkN4RgAenAR5ymAmQpJ+KAAFSM8r3p6DQ8my48qA4WMnTZVAIToAkgBdbHnNyh4UCgBPFEIKAHd2RABrZnklMCA

**Additional context:**

```sh
$ bundle exec herb -v
herb gem v0.10.1, libprism v1.9.0, libherb v0.10.1 (Ruby C native extension)
```

Somewhat related to #855.

joaoGabriel55:
## Overview

The formatter is free to turn a space into a newline and back — HTML collapses both to a single space. What it must never do is **introduce** whitespace where the source had none, or **remove** whitespace where the source had some, because either changes the rendered output.

Three open issues turned out to be instances of that one invariant:

```erb
<p>Hello<% if owner %> <%= owner.name %>'s dog<% end %>!</p>
```

rendered `Hello Peggy's dog!` before formatting and `Hello Peggy's dog !` after — and `Hello!` became `Hello !` when the branch wasn't taken.

The original approach in this PR gated on a "glued boundary" and treated a non-output `<% … %>` tag as *breaking* the glue chain, on the reasoning that the tag renders nothing so a line break next to it is safe. That reasoning conflates two different things: the tag emits no characters (true), and a newline placed *next to* the tag emits no characters (false). Everything outside `<% %>` delimiters is literal template output, so a newline before `<% end %>` still lands in the rendered stream. That's why only one side of the boundary got fixed.

## Related Issue

Closes #1729.
Closes #469.

Partially addresses #1883 — the whitespace *between* children of an inline element is fixed here. The leading/trailing case (`x<span> y </span>z` → `xyz`) is a separate, context-dependent problem and remains open.

## Affected Component(s)

- [x] 💅 Formatter
- [x] 📦 JavaScript / Node.js / WebAssembly

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Main Technical Changes

**Non-output ERB is modelled as transparent, not as a separator.** `flowRole()` classifies a node as `visible`, `transparent` or `separator`. A `<% if %>` / `<% end %>` / `<% i += 1 %>` tag is zero-width, so glue propagates *across* it: in `dog<% end %>!` the `dog` and the `!` are adjacent in the output and neither boundary can take a break. `hasGluedTextFlowBoundary` is removed in favour of `isGluedControlFlowNode`, and a glued control-flow node becomes a single atomic token in text flow (`tryRenderControlFlowInline`), falling back to the previous block layout when it wraps a block element or multiline ERB.

**ERB control-flow bodies use the same path as element bodies.** `visitStatements` routes `if` / `unless` / `when` / `in` / `begin` / `while` / `until` / `for` / `rescue` / `ensure` / `else` through `visitElementChildren` instead of bare `visitAll`, which had been skipping `shouldAppendToLastLine` and text-flow runs entirely. `ERBBlockNode` and `ERBRenderNode` already did this; the other eleven were the exception. Blank-line insertion is disabled on that path — the "rule of three" is a layout choice for sibling *elements* and would start inserting blank lines into ERB that never had them.

**Void elements stay attached to glued content.** `<br>` / `<hr>` have no children, so `tryRenderInlineFull` reported them as un-inlineable and they broke the surrounding flow, detaching the comma in `<%= greeting %>,<br>` (#469). They are now inlineable when glued to what precedes them, and still end the line they land on. When whitespace separates them they keep their own line as before.

**Whitespace between children of an inline element is preserved.** `filterSignificantChildren` only keeps a whitespace text node that is exactly `" "`, so the `"\n  "` between two `<em>`s was dropped before rendering and `a b` became `ab` (#1883). Inline elements now render from the unfiltered body.

### Changed expectations

Seven existing tests asserted output that injected whitespace the source did not have. Worth reviewing closely:

| test | was asserting |
|---|---|
| `while` / `until` loops | `<b><%= i %></b>` split from a glued `<% i += 1 %>`, putting a space between loop iterations |
| `<br> in <p>` / `<hr> in <p>` | `One` split from its glued `<br>` |
| `erb.test.ts` #564-comment, #436 Example 6 | same, for `.<br />` and `Details:<br>` |
| `glued-text-flow.test.ts` punctuation tests | the comma detached from the `<br>` — the #469 symptom itself |
| `if with multiple consecutive ERB outputs` | ERB outputs not wrapping, unlike every other body |

In three of these the **document-root variant of the same test already asserted the corrected form** — for example the `<br>` test directly above `<br> in <p>` expects `One<br>` on one line. Element and control-flow bodies were the outliers, not the new behaviour.

## How to Test

```bash
cd javascript/packages/formatter
yarn vitest run
```

New suite: `test/erb/whitespace-preservation.test.ts` — glue on either or both sides of a control-flow node, across `unless` / `case` / `while` / nested control flow, overflow staying on one line, whitespace-separated control flow still laid out as a block, fallback when a block element or multiline ERB is inside, the three #469 cases from the issue, and the #1883 inner-whitespace cases.

Three `test.fails` document what is deliberately left open: two leading/trailing-whitespace cases from #1883, plus a pre-existing comment-indent case.

Quick manual check:

```erb
<p>  Hello<% if owner %> <%= owner.name %>'s dog<% end %>!</p>
```

should format to, and stay at:

```erb
<p>
  Hello<% if owner %> <%= owner.name %>'s dog<% end %>!
</p>
```

## Checklist

- [x] 🧪 Tests added or updated
- [x] 🔨 Ran the relevant test suites
- [ ] 📚 Documentation updated (if applicable)
- [x] 🧭 Commit messages follow the `Component: Description` convention

### DIFF
--- a/javascript/packages/formatter/src/format-printer.ts
+++ b/javascript/packages/formatter/src/format-printer.ts
@@ -719,6 +719,26 @@ export class FormatPrinter extends Printer implements TextFlowDelegate, Attribut
     }
   }
 
+  /**
+   * Visit ERB control-flow statements (the body of an `if`, `each`, `case`, …).
+   *
+   * When the statements contain rendered output glued directly together with no
+   * whitespace between (e.g. `<%= user.name %>'s dog` or `<%= greeting %>,`),
+   * route them through the text-flow engine so the glued content isn't split
+   * onto separate lines — which would inject whitespace into the rendered HTML
+   * (#1729). When the statements are only ever separated by whitespace (e.g. a
+   * list of standalone `<%= … %>` outputs), or the gluing only involves a
+   * non-output `<% … %>` statement (which renders nothing), breaking across
+   * lines is rendering-safe, so we keep the existing per-node visiting that
+   * breaks at ERB boundaries.
+   */
+  private visitStatements(statements: Node[]) {
+    if (this.textFlow.hasGluedTextFlowBoundary(statements)) {
+      this.textFlow.visitTextFlowChildren(statements)
+    } else {
+      this.visitAll(statements)
+    }
+  }
 
   /**
    * Visit element children with intelligent spacing logic
@@ -1036,7 +1056,7 @@ export class FormatPrinter extends Printer implements TextFlowDelegate, Attribut
   visitERBInNode(node: ERBInNode) {
     this.trackBoundary(node, () => {
       this.printERBNode(node)
-      this.withIndent(() => this.visitAll(node.statements))
+      this.withIndent(() => this.visitStatements(node.statements))
     })
   }
 
@@ -1111,7 +1131,7 @@ export class FormatPrinter extends Printer implements TextFlowDelegate, Attribut
         this.printERBNode(node)
 
         this.withIndent(() => {
-          this.visitAll(node.statements)
+          this.visitStatements(node.statements)
         })
 
         if (node.subsequent) this.visit(node.subsequent)
@@ -1126,13 +1146,13 @@ export class FormatPrinter extends Printer implements TextFlowDelegate, Attribut
     if (this.inlineMode) {
       this.visitAll(node.statements)
     } else {
-      this.withIndent(() => this.visitAll(node.statements))
+      this.withIndent(() => this.visitStatements(node.statements))
     }
   }
 
   visitERBWhenNode(node: ERBWhenNode) {
     this.printERBNode(node)
-    this.withIndent(() => this.visitAll(node.statements))
+    this.withIndent(() => this.visitStatements(node.statements))
   }
 
   visitERBCaseNode(node: ERBCaseNode) {
@@ -1150,7 +1170,7 @@ export class FormatPrinter extends Printer implements TextFlowDelegate, Attribut
   visitERBBeginNode(node: ERBBeginNode) {
     this.trackBoundary(node, () => {
       this.printERBNode(node)
-      this.withIndent(() => this.visitAll(node.statements))
+      this.withIndent(() => this.visitStatements(node.statements))
 
       if (node.rescue_clause) this.visit(node.rescue_clause)
       if (node.else_clause) this.visit(node.else_clause)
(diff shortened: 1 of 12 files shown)


marcoroth (approved):
Thank you for the pull request @joaoGabriel55! 

I touched it up a bit so we can also resolve #1729, #469 and partially #1883, as they are closely related, thank you! 🙏🏼

PR #1866, opened 2026-07-21
Linter: Fix Prism byte-offset handling for non-ASCII source

### LINKED ISSUE #1761: Auto Fix Issue : herb-lint `--fix` corrupts source when a multi-byte UTF-8 char precedes a second `<%= "#{...}" %>` offence
**Input:**
```erb
<%= "#{first}  – " %>
<%= "#{second} - " %>
```

**Output:**
```erb
<%= first %>  –
<%= cond} %> -
```

**Expected:**
```erb
<%= first %>  –
<%= second %> -
```

**Playground:**

That input snippet, when put in the playground shows the issue when you "Auto Fix" it.

**Additional context:**

The trigger is a multi-byte UTF-8 char (the em-dash `–`, of course claude would add that!) in the first ERB; prism reports byte offsets  but `String.prototype.substring` indexes UTF-16 code units, so any expression after a multi-byte char gets sliced from the wrong window. 

## Other call sites with same issue

Same `source.substring(prismNode.location.startOffset, ...)` pattern elsewhere in `@herb-tools/linter`:

- erb-prefer-direct-output — has autofix — corrupts source via autofix. With em-dash before offence: expected `<%= second %>`, actual `<%= cond} %>`.
- erb-no-debug-output — no autofix — wrong text in error message. Expected `pp @foo`, actual ` @foo %`.
- erb-no-unused-expressions — no autofix — wrong text in error message. Expected `@foo.bar`, actual `oo.bar %`. 
- erb-no-unused-literals — no autofix — wrong text in error message. Expected `"hello"`, actual `ello" %`.
  
## Proposed fix

Convert byte offsets to code-unit offsets at the slice boundary, eg a `sliceByteRange(source, start, length)` shared helper so every rule that slices a JS string with prism byte offsets goes through one safe path.

### LINKED ISSUE #1855: Linter: `erb-prefer-direct-output` autocorrect corrupts expressions when a multibyte character precedes the node
**Package:** `@herb-tools/linter` 0.10.2 (bug lives in `@herb-tools/rewriter`)
**Command:** `herb-lint --fix`

## Reproduction

`app/views/example.html.erb`:

```erb
<p>é</p>
<p><%= "#{@organisation.slug}.#{AppHost.base_host}" %></p>
```

```bash
herb-lint --fix app/
```

**Actual** (garbled — leading characters cut off, trailing `}` pulled in):

```erb
<p>é</p>
<p><%= organisation.slug} %>.<%= ppHost.base_host} %></p>
```

**Expected:**

```erb
<p>é</p>
<p><%= @organisation.slug %>.<%= AppHost.base_host %></p>
```

Remove the `é` and the autocorrect is perfect — any multibyte character *anywhere before the node* breaks it. Each extra UTF-8 byte shifts the extracted expressions by one character: `é` (2 bytes) shifts by 1, an em dash `—` (3 bytes) by 2. We hit this in a real template where an `—` in an earlier comment turned

```erb
<strong><%= "#{@organisation.slug}.#{AppHost.base_host}" %></strong>
```

into

```erb
<strong><%= rganisation.slug}. %>.<%= pHost.base_host}" %></strong>
```

## Root cause

`ERBStringToDirectOutputRewriter.extractExpressionSource` slices the Ruby source with Prism location offsets applied as JavaScript string indices:

```js
const expressionStart = openingLocation.startOffset + openingLocation.length;
const expressionEnd = closingLocation.startOffset;
return source.substring(expressionStart, expressionEnd);
```

Prism's `startOffset` counts **UTF-8 bytes**, but `String.prototype.substring` indexes **UTF-16 code units** — the two diverge by one for every 2-byte character (and more for wider ones) that appears earlier in the string. String parts usually survive because `extractStringContent` prefers `stringNode.unescaped.value` and only falls back to offset slicing; expression parts always slice, so they always corrupt.

Likely fix: convert the source to a byte view before slicing (`Buffer.from(source, "utf8").subarray(start, end).toString("utf8")` / `TextEncoder`), or translate Prism byte offsets to character offsets once per document. Other call sites that mix Prism offsets with JS string APIs probably deserve the same audit.

<details>
<summary>Runnable reproduction script (exits 1 while the bug is present)</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail
cd "$(mktemp -d)"

mkdir -p app/views
cat > app/views/example.html.erb <<'EOF'
<p>é</p>
<p><%= "#{@organisation.slug}.#{AppHost.base_host}" %></p>
EOF
cat > .herb.yml <<'YAML'
version: 0.10.2
framework: actionview
linter:
  enabled: true
YAML

bunx --package @herb-tools/linter@0.10.2 herb-lint --fix app/ >/dev/null 2>&1 || true

echo "--- after herb-lint --fix ---"
cat app/views/example.html.erb

if grep -q "@organisation.slug" app/views/example.html.erb; then
  echo "OK: expressions extracted intact (bug not present)"
else
  echo "BUG: expressions corrupted — byte offsets applied as character indices"
  exit 1
fi
```

</details>

drjayvee:
Heads up: this commit and PR are LLM-generated and only superficially reviewed by me. The bug is real though! End of human output :D

## Summary

Prism reports node locations as UTF-8 **byte** offsets, but several linter rules and the `erb-prefer-direct-output` autofix used those offsets directly as UTF-16 **string-index** offsets when calling `source.substring()`. This is only correct when the file is pure ASCII — any multi-byte character earlier in the file (even inside an unrelated tag or comment) shifts every subsequent byte offset, corrupting the result.

This is what caused reports like:

```diff
-            <%= "#{card_event.event_type}(#{data})" %>
+            <%= ard_event.event_type} %>(<%= ata} %>)
```

when running `herb-lint --fix` on a file that has any non-ASCII character (accented letters, emoji, etc.) earlier in the source.

- Adds `stringIndexFromByteOffset` to `@herb-tools/core` to correctly convert a Prism UTF-8 byte offset to a UTF-16 string index.
- Uses it in the `erb-prefer-direct-output` autofix (`ERBStringToDirectOutputRewriter.extractStringContent` / `extractExpressionSource`), which was corrupting the rewritten source.
- Uses it in `locationFromOffset` (shared by `erb-no-instance-variables-in-partials`, `erb-no-debug-output`, `erb-no-unused-literals`, `erb-no-unused-expressions`, `erb-prefer-direct-output`), which was reporting the wrong offense line/column whenever non-ASCII characters preceded the offense.
- Adds a `substringFromByteOffset` helper and uses it in the 3 rules that extracted a source snippet for their message text (`erb-no-debug-output`, `erb-no-unused-literals`, `erb-no-unused-expressions`), which were showing a corrupted snippet in the same scenario.

## Test plan

- [x] New autofix regression tests in `erb-prefer-direct-output.autofix.test.ts` reproducing the reported bug (preceded by an accented character / emoji)
- [x] New unit tests for `stringIndexFromByteOffset` in `core/test/prism/byte-offset.test.ts`
- [x] New regression tests for the location/message corruption in `erb-no-debug-output.test.ts`, `erb-no-unused-literals.test.ts`, `erb-no-unused-expressions.test.ts`, `erb-no-instance-variables-in-partials.test.ts`
- [x] Verified end-to-end against real Prism-parsed byte offsets (via the published `0.10.2` packages, patched with this fix) that all of the above scenarios produce corrupted output before the fix and correct output after

Resolves https://github.com/marcoroth/herb/issues/1761
Resolves https://github.com/marcoroth/herb/issues/1855


marcoroth:
Hey @drjayvee, thanks a lot for digging into this and for the write-up!

The diagnosis is spot on. Prism reports everything in UTF-8 bytes and we were feeding those straight into `String#substring`, so any non-ASCII character earlier in the file shifted everything after it. Nasty one to track down, especially since the autofix silently wrote the corrupted source back to disk.

I pushed a few small things on top:

- Moved `substringFromByteOffset` into `@herb-tools/core` next to `stringIndexFromByteOffset`. `ERBStringToDirectOutputRewriter#extractStringContent` was hand-rolling the exact same thing because it couldn't import from the linter, so it's a one-liner now.
- Moved `positionFromOffset` and `locationFromOffset` into core as well, and renamed the latter to `locationFromByteOffset`. The old name didn't say anything about taking byte offsets, which is part of what made this easy to get wrong in the first place.
- Added two tests for cases we weren't covering yet: multi-byte characters *surrounding* an interpolation (which goes through the `openingLoc`/`closingLoc` arithmetic rather than the leading offset), and a multi-byte character on the *same line* as the offense, so the column walk itself has to step over it.

Everything else is exactly as you had it. Thank you! 🙌

drjayvee:
Servus Marco,

My pleasure. Thank you for picking this up!

Tschüss.

PR #1868, opened 2026-07-22
Linter: Implement `actionview-prefer-pluralize-helper` rule

### LINKED ISSUE #1723: Linter Rule: Prefer `ActionView::Helpers::TextHelper#pluralize` over `String#pluralize` for counts
### ✅ Good

```erb
<%= pluralize("Known Alias", aliases.size) %>
Known <%= pluralize("Alias", aliases.size) %>
```

### 🚫 Bad

```erb
<%= aliases.size %><%= "Known Alias".pluralize(aliases.size) %>
<%= aliases.size %> Known <%= "Alias".pluralize(aliases.size) %>
```

## References

- https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize
- https://apidock.com/rails/String/pluralize

joaoGabriel55:
## Overview

Adds a new linter rule, `actionview-prefer-pluralize-helper`, that flags a count rendered next to a `String#pluralize` call in ERB templates and recommends the `ActionView::Helpers::TextHelper#pluralize` helper instead.

The helper form (`pluralize(count, "Alias")`) prepends the count to the output and reads naturally in a template, whereas `"Alias".pluralize(count)` typically forces the count to be rendered separately — leading to duplicated output like `<%= aliases.size %> <%= "Alias".pluralize(aliases.size) %>`.

```erb
<!-- 🚫 Bad -->
<%= aliases.size %> Known <%= "Alias".pluralize(aliases.size) %>

<!-- ✅ Good -->
<%= pluralize(aliases.size, "Known Alias") %>
```

The offense message includes a concrete, source-derived suggestion, e.g.:

> Prefer the `pluralize` helper over separate count and `String#pluralize` output. Use `<%= pluralize(aliases.size, "Known Alias") %>` instead.

## Related Issue

Closes #1723

## Affected Component(s)

- [x] 🔎 Linter
- [x] 📚 Documentation

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Main Technical Changes

- New rule `actionview-prefer-pluralize-helper` (`javascript/packages/linter/src/rules/actionview-prefer-pluralize-helper.ts`), built on the Prism nodes the parser already exposes:
  - Uses `parserOptions: { prism_nodes: true, prism_program: true }` and walks the sibling nodes of every node that has children, so the pair is also found inside `if`, `unless`, `else` and iteration blocks.
  - Flags a pair when a count output tag (`length` / `size` / `count`) is followed by an output tag whose method name is `pluralize`, whose **receiver is a string literal** (`StringNode` / `InterpolatedStringNode`), and whose **single argument** is syntactically identical to the count. This scoping avoids false positives on non-string receivers (e.g. `model.pluralize(count)`), on the argument-less `"Alias".pluralize`, and on a `String#pluralize` call that stands on its own, matching the "for counts" intent of the issue.
  - Default severity: `warning`; `introducedIn: "unreleased"`.
  - The suggested fix is reconstructed from the original source spans of the receiver and arguments via a small `slice()` helper. Any literal text between the two tags is folded into the singular, keeping the receiver's original quoting and escape sequences.
- Registered the rule in `src/rules.ts` and re-exported it from `src/rules/index.ts`.
- Added docs page `docs/rules/actionview-prefer-pluralize-helper.md` and linked it from `docs/rules/README.md` (required by the `rule-docs` completeness test).

## How to Test

```bash
cd javascript/packages/linter
npx vitest run test/rules/actionview-prefer-pluralize-helper.test.ts   # new rule tests (30)
npx vitest run test/rule-docs.test.ts                           # docs completeness
npx tsc -b                                                      # typecheck
```

The rule test covers valid cases (helper form, argument-less `String#pluralize`, non-string receivers, an isolated `String#pluralize` call, mismatched counts, an additional locale argument) and invalid cases (a count output tag followed by a string-literal `pluralize`, with and without literal text between them, and inside `if`, `unless`, `else` and iteration blocks).

## Checklist

- [x] 🧪 Tests added or updated
- [ ] 🔨 Ran `bin/integration` locally (ran the linter package test suite: 4080 passed)
- [x] 📚 Documentation updated (if applicable)
- [x] 🧭 Commit messages follow the `Component: Description` convention (`Linter: Add \`erb-prefer-pluralize-helper\` rule`)


javascript/packages/linter/docs/rules/erb-prefer-pluralize-helper.md
```diff
@@ -0,0 +1,40 @@
+# Linter Rule: Prefer the `pluralize` helper over `String#pluralize` for counts
+
+**Rule:** `erb-prefer-pluralize-helper`
```
marcoroth: Hey @joaoGabriel55, thank you for the PR and the rule idea, I like this!

Would you mind renaming this to `actionview-*` instead?

```suggestion
**Rule:** `actionview-prefer-pluralize-helper`
```


Thank you!

marcoroth (commented):
Hey @joaoGabriel55, thanks for working on this!

I'm aware that I opened the issue #1723 and I just realized that it might not be as trivial as it seems on first sight. I also got the argument order wrong in the issue, it's `pluralize(count, singular, plural = nil)`, so the count comes first and my examples there had it the other way around.

The problem is that `ActionView::Helpers::TextHelper#pluralize` and `String#pluralize` serve two different use-cases, as they have different behavior.

The ActionView one outputs the count with it, while the String one doesn't. So, I feel like we'd need to also detect that and not flag the individual case like this, so it might still be valid to call `<%= "Known Alias".pluralize(aliases.size) %>` without wanting to also output the count.

So I think for this rule to be useful we need to do a few more things.

Given this example:

```html+erb
<%= aliases.size %> <%= "Known Alias".pluralize(aliases.size) %>
```

* Find two consecutive ERB tags (might include whitespace/other text between them)
* Make sure the first ERB tag calls `length`, `size` or `count` on the receiver, and if thats the case, remember the receiver
* Make sure the second ERB tag calls `pluralize` on a string literal receiver and then make sure the receiver from the first ERB tag matches the first argument of the pluralize `CallNode` including the method call.

So, these should remain valid with this rule:

```html+erb
<%= pluralize(aliases.size, "Known Alias") %>
Known <%= pluralize(aliases.size, "Alias") %>
<%= aliases.size %> <%= "Known Alias".pluralize(other.size) %>
<%= aliases.count %> <%= "Known Alias".pluralize(aliases.size) %> <%# we cannot necessarily assume size == count %>
<%= aliases.size %> <%= variable.pluralize(aliases.size) %> <%# we cannot ensure this is String#pluralize %>
```

While these should get flagged:

```html+erb
<%= aliases.size %><%= "Known Alias".pluralize(aliases.size) %>
<%= aliases.size %> <%= "Known Alias".pluralize(aliases.size) %>
<%= aliases.size %> Known <%= "Alias".pluralize(aliases.size) %>
```

and the suggestion should say, use this instead:

```html+erb
<%= pluralize(aliases.size, "Known Alias") %>
```

joaoGabriel55:
## QA - 08/04

image

image

joaoGabriel55:
@marcoroth Sorry about those labels again 😞 .

marcoroth (approved):
Thanks @joaoGabriel55! 🙌🏼 

I did a final pass on it and touched up some nits, thank you! 🙏🏼

PR #1872, opened 2026-07-24
Engine: Reduce template compilation overhead

joelhawksley:
## Summary

A set of allocation- and CPU-focused optimizations for the parse + compile
path, found by profiling `Herb::Engine` compilation of a large real-world
template corpus (~700 `.html.erb` files). Each optimization is in its own commit.

Measured on that corpus (Ruby 3.4.9 +PRISM, compiling with validation
disabled), versus v0.10.2:

- **Compile time: −36%** (≈281 µs → ≈181 µs per template)
- **Parse allocations: −72%**; compile-path allocations **−67%**
- **Compiled output is byte-identical** across the whole corpus and a set of
  adversarial whitespace/trim cases
- Hot-path (post-compile) rendering behavior is unchanged

## Optimizations (one per commit)

1. **`track_locations` parse option** — `Herb.parse` builds a `Location` (two
   `Position`s) for every node and token. Callers that never read source
   locations (e.g. rendering with validation off) can pass
   `track_locations: false` to leave them nil, skipping roughly half of the
   parse's Ruby allocations. Defaults to `true`, so existing behavior is
   unchanged.

2. **Intern node/token type strings and short token values** — type strings and
   short token values are drawn from tiny fixed vocabularies (~96% of token
   values are duplicates like `"\n"`, `"%>"`, `">"`, tag names). Interning
   shares one frozen `String` per distinct value instead of allocating a fresh
   copy each time. Longer token values (arbitrary text) are left as-is.

3. **Skip the recursive error walk for cleanly-parsed templates** —
   `ParseResult#errors` walked the entire AST (`value.recursive_errors`) on
   every call. Count errors as they are materialized and record the total on the
   result; when it is zero (the common case), return the top-level errors and
   skip the walk entirely.

4. **Accumulator-based `Node#recursive_errors`** — replace
   `errors + compact_child_nodes.flat_map(&:recursive_errors)` (which allocates
   an intermediate array at every node) with a single shared accumulator walked
   iteratively.

5. **Lazy `Engine#relative_file_path`** — was computed eagerly with
   `Pathname#relative_path_from` + `#to_s` on every compile, but is only used
   when emitting errors, overlays, or debug output. Derive it on demand instead.

6. **Fuse whitespace compaction into `optimize_tokens`; use `String#match?`** —
   fold the separate `compact_whitespace_tokens` pass (which built an
   intermediate array via `map.with_index` + `compact`) into `optimize_tokens`'
   single pass, and switch the boolean-context regexp guards in the whitespace
   helpers from `=~` to `String#match?` (no `MatchData` allocation, no `$~`).

## Notes

- The only public API change is the additive `track_locations:` option on
  `Herb.parse`.
- Generated files (`ext/herb/nodes.c`, `ext/herb/error_helpers.c`) are not
  committed; the changes live in their `.erb` templates.
- Verification: compiled `Herb::Engine#src` output is byte-for-byte identical to
  v0.10.2 across all ~700 corpus templates and a set of hand-crafted
  whitespace/trim edge cases; the downstream ReActionView test suite passes
  against this build.

---

_This PR was written with Claude Opus 4.8._

### DIFF
--- a/ext/herb/extension.c
+++ b/ext/herb/extension.c
@@ -39,7 +39,15 @@ typedef struct {
 static VALUE parse_convert_body(VALUE arg) {
   parse_args_T* args = (parse_args_T*) arg;
 
-  return create_parse_result(args->root, args->source, args->parser_options);
+  // Reset the per-parse error counter, then record the total on the result so
+  // Ruby can skip the recursive error walk when the template parsed cleanly.
+  herb_ext_error_count = 0;
+
+  VALUE result = create_parse_result(args->root, args->source, args->parser_options);
+
+  rb_ivar_set(result, rb_intern("@total_error_count"), UINT2NUM(herb_ext_error_count));
+
+  return result;
 }
 
 static VALUE parse_cleanup(VALUE arg) {
--- a/ext/herb/extension_helpers.c
+++ b/ext/herb/extension_helpers.c
@@ -2,6 +2,8 @@
 
 #include <stdbool.h>
 
+#include <ruby/encoding.h>
+
 #include "extension.h"
 #include "extension_helpers.h"
 #include "nodes.h"
@@ -22,6 +24,11 @@ const char* check_string(VALUE value) {
 
   return RSTRING_PTR(value);
 }
+
+// Maximum byte length of a token value we will intern. Structural tokens and
+// identifiers (delimiters, tag/attribute names) are short and highly repeated;
+// longer values are treated as unique text content and not interned.
+#define HERB_INTERN_VALUE_MAX_LENGTH 16
 
 static ID id_line, id_column, id_start, id_end, id_from, id_to, id_value, id_range, id_location, id_type;
 static bool ast_value_ivar_ids_initialized = false;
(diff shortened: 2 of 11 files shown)


joelhawksley:
@marcoroth I just took a first pass at trying to improve compilation performance with a lot of help from Claude. What do you think of these changes? I'm also happy to pursue any other approaches you have in mind.

marcoroth (commented):
Hey @joelhawksley, thanks for these ideas!

I think all 5 ideas are all valuable and definitely worth a look. Though, I think I would prefer these as individual pull requests so they are more easily identifiable. 

Or, at least, the `track_locations `, `ParseResult#errors` and `Node#recursive_errors` ideas feel like easier and simple to land as separate PRs.

Would you mind splitting them up? Thank you!

marcoroth:
Just for context, the work from this pull request landed in https://github.com/marcoroth/herb/pull/2199, https://github.com/marcoroth/herb/pull/2217, https://github.com/marcoroth/herb/pull/2232, https://github.com/marcoroth/herb/pull/2235, https://github.com/marcoroth/herb/pull/2234, and https://github.com/marcoroth/herb/pull/2262.

PR #1880, opened 2026-07-27
Engine: Add `title` to excluded tags

kcdragon:
This isn't a problem in most Rails apps since most apps have the `head` and `title` tag in the same view file (the layout file) but I have a project that breaks out the `title` into a separate partial so it can be included in several layouts. When that happens, a literal span gets included in the browser title.

**shared/_head.html.erb**

```
<%# locals: () %>
<title>
  <% if content_for?(:title) %>
    <%= content_for(:title) %> |
  <% end %>
  <%= Jumpstart.config.application_name %>
</title>
```

**Before**

Screenshot 2026-07-27 at 7 04 24 PM

**After**

Screenshot 2026-07-27 at 7 05 50 PM

### DIFF
--- a/lib/herb/engine/debug_visitor.rb
+++ b/lib/herb/engine/debug_visitor.rb
@@ -336,7 +336,7 @@ module Herb
       end
 
       def in_excluded_context?
-        excluded_tags = ["script", "style", "head", "textarea", "pre", "svg", "math"]
+        excluded_tags = ["script", "style", "head", "title", "textarea", "pre", "svg", "math"]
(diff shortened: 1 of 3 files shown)


marcoroth (commented):
Great catch, thanks for this pull request @kcdragon! 🙏🏼 

It looks like this branch is missing the snapshot for the `<title>` test. Would you mind committing the snapshot file as well? Thank you 🙌🏼

marcoroth (approved):
Awesome, thank you @kcdragon! 🙏🏼

PR #1953, opened 2026-08-01
Engine: Never wrap head-emitting helper output in debug spans

kcdragon:
# Problem

While using Herb in development, I'm seeing an error from Turbo saying that I'm loading Turbo from a `<script>` inside `<body>` despite it being inside `<head>` in my source code.

**app/views/shared/_head.html.erb**
```
<%# locals: () %>
<%= csrf_meta_tags %>
<%= csp_meta_tag %>
<%= viewport_meta_tag %>

<%= render "shared/seo" %>

<%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
<%= stylesheet_link_tag "application", media: "all", "data-turbo-track": "reload" %>

<%# Be sure to add your own custom favicons %>
<%= render "shared/favicons" %>
<%= render "shared/payments_dependencies" unless Rails.env.test? %>

<%= render "shared/google_analytics" %>
<%= render "shared/ahrefs_analytics" %>

<%= content_for(:head) %>

```

Screenshot 2026-08-01 at 5 16 29 PM

Screenshot 2026-08-01 at 5 25 53 PM

# Solution

A partial rendered into a layout's <head> has no <head> node in its own AST, so the element-stack exclusion in `in_excluded_context?` never fires and helpers like `csrf_meta_tags` or `javascript_include_tag` got wrapped in a `<span style="display: contents;">`. Since <span> is invalid head content, browsers implicitly close </head> there and push the remaining scripts/stylesheets into <body>, breaking Turbo apps.

Add a name-based `head_content_helper?` skip (alongside the existing `complex_rails_helper?` guard) covering the standard head helpers and `tag.meta`/`tag.link`/`tag.title`/`tag.base`, and add `meta`, `link`, and `base` to the excluded tags for the inline-in-layout case.

Screenshot 2026-08-01 at 5 17 24 PM

# _**Is there a better fix than this?**_ 

It seems like it would be impossible to catch all the potential helper methods that belong in `<head>` which is the direction this PR is going in. Maybe its good enough for now? It seemed like a fine approach since its similar to the existing check done by `complex_rails_helper?(code)`. I'm curious if you know of a better approach.

Another option is a herb linter that disscourages `<head>` partials but I think those a fairly common especially in apps with multiple layouts.


lib/herb/engine/debug_visitor.rb
```diff
@@ -399,6 +393,30 @@ def javascript_tag?(code)
         false
       end
 
+      # TODO: Rewrite using Prism Nodes once available
```
kcdragon, line 394: I added this TODO because related methods have this same TODO

lib/herb/engine/debug_visitor.rb
```diff
@@ -399,6 +393,30 @@ def javascript_tag?(code)
...
+          "csp_meta_tag",
+          "viewport_meta_tag",
+          "javascript_include_tag",
+          "javascript_importmap_tags",
+          "stylesheet_link_tag",
+          "stylesheet_import_tag",
+          "favicon_link_tag",
+          "auto_discovery_link_tag",
+          "preload_link_tag"
+        ]
```
marcoroth: I wonder if we should add this metadata into our Action View Helper registry in `config/action_view_helpers/*` so we could dynamically query them.

I added the missing `importmap-rails` helpers in https://github.com/marcoroth/herb/pull/2005

config/action_view_helpers/actioncable/action_cable_helper/action_cable_meta_tag.yml
```diff
@@ -4,6 +4,7 @@ source: "ActionCable::Helpers::ActionCableHelper#action_cable_meta_tag"
 gem: "actioncable"
 output: "html"
 visibility: "public"
+head_content: true
```
marcoroth: @kcdragon I think this is the best thing to do long-term!

Maybe we could reverse it to something like:
```yaml
context: "head"
```

So other helpers could define other contexts, like:
```yaml
context: "body"
context: "html"
context: "table"
```

There is also this other pull request which adds more schema-like metadata that could be also useful to have: https://github.com/marcoroth/herb/pull/1030

kcdragon: @marcoroth Cool. I'll change this PR to use `context: "head"`. The other PR is interesting. Is the idea that if we knew which element these helpers outputted, we could automatically determine if it could only appear in `<head>` without needing something like `context: "head"`?

marcoroth: Yeah, at least as long as the helper actually maps to an HTML tag.

test/snapshots/engine/debug_mode_test/test_0074_standalone_meta_link_and_base_content_erb_expressions_do_NOT_get_debug_spans_89fb0c6b52349ca087eecfe6f2f5b627.txt
```diff
@@ -0,0 +1,9 @@
+---
+source: "Engine::DebugModeTest#test_0074_standalone meta link and base content erb expressions do NOT get debug spans"
+input: "{source: \"<meta name=\\\"description\\\" content=\\\"<%= @description %>\\\">\\n<link rel=\\\"canonical\\\" href=\\\"<%= canonical_url %>\\\">\\n<base href=\\\"<%= base_href %>\\\">\\n\", options: {debug: true}}"
+---
+_buf = ::String.new; _buf << '<meta name="description" content="'.freeze; _buf << ::Herb::Engine.attr((@description)); _buf << '" data-herb-debug-outline-type="view" data-herb-debug-file-name="unknown" data-herb-debug-file-relative-path="unknown" data-herb-debug-file-full-path="unknown" data-herb-debug-attach-to-parent="true">
```
kcdragon, line 5: @marcoroth does this snapshot seem right to you? I'm surprised it's got the herb data attributes. Though the attributes appear to be on `meta` instead of a `span` that's wrapping it.

marcoroth: @kcdragon I think the problem is its not directly enclosed in a `<head>` context, so it doesn't apply the skip.

PR #1954, opened 2026-08-01
Config: Add `all` pseudo rule to set the default for every linter rule

### LINKED ISSUE #1643: Add possibility to disable all rules via config
## Context

Right now, if I want to verify a set of rules on a codebase, I have to disable all rules manually in the `.herb.yml` config file (for which I first have to know which are even enabled by default), and then I can enable the set of rules that I wanted to add.

Example
```
linter:
  enabled: true

  rules:
    # Disable rules that are enabled by default
    parser-no-errors:
      enabled: false
    html-no-unescaped-entities:
      enabled: false
    erb-no-unused-expressions:
      enabled: false
    # ...

    # Enable rules we are testing / adding
    html-no-event-handlers:
      enabled: true
```

This setup is quite tedious and also makes it more difficult for people to start with their own _fully configured_ set of rules.

## Proposal

Add a configuration option to `.herb.yml` that enables this workflow

### Examples

Rubocop enables this via the `DisabledByDefault` option. Example:

```
AllCops:
  DisabledByDefault: true
```

## Result

Ultimately we would end up with something like

```
linter:
  enabled: true

  rules:
    all:
      enabled: false
    my-rule:
      enabled: true
```

marcoroth:
This pull request adds an `all` pseudo rule to `linter.rules` in `.herb.yml` that sets the default `enabled` state for every rule you don't list explicitly.

Verifying a small set of rules against a codebase currently means disabling everything else by hand, which requires knowing which rules are on by default in the first place, and then keeping that list up to date as new rules land. With `all` you can start from an empty set and opt back in:

```yaml
linter:
  enabled: true

  rules:
    all:
      enabled: false

    html-no-event-handlers:
      enabled: true
```

Setting `all: enabled: true` does the opposite and turns on every rule, including the ones that are off by default:

```yaml
linter:
  rules:
    all:
      enabled: true

    html-no-title-attribute:
      enabled: false
```

Resolves https://github.com/marcoroth/herb/issues/1643

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -185,6 +185,52 @@ Each rule can be configured with the following options:
 - **`only`**: Array of glob patterns - Restrict rule to ONLY these files (can override parent excludes, overrides `include`)
 - **`exclude`**: Array of glob patterns - Exclude files from this rule (always applied)
 
+### Setting the Default for All Rules <Badge type="tip" text="^0.11.0" />
+
+The `all` pseudo rule sets the default `enabled` state for every rule you don't list explicitly. It's the way to opt into a fully explicit rule set without having to know (and repeat) which rules are on by default:
+
+```yaml [.herb.yml]
+linter:
+  enabled: true
+
+  rules:
+    # Turn every rule off ...
+    all:
+      enabled: false
+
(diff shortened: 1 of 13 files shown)


markokajzer:
legend ❤️

PR #1961, opened 2026-08-02
Linter: Implement `html-no-nested-forms` rule

### LINKED ISSUE #166: Linter Rule: Disallow nested forms
##### Description

Disallow placing one `<form>` element inside another `<form>` element. HTML does not support nested forms. Doing so results in invalid markup and unpredictable behavior.

##### Rationale

Nesting forms is invalid according to the HTML specification. Browsers will automatically close any open `<form>` tag when encountering a new `<form>` start tag, often leading to:

- broken form submissions,
- incomplete or missing form fields,
- confusing DOM structure,
- inconsistent behavior across browsers.

Even if some browsers attempt to handle this situation, the resulting form behavior is unreliable and prone to subtle bugs.

This rule ensures that each form is properly isolated.

##### ✅ Good

```html+erb
<form>
  <input type="text" name="name">
</form>

<form>
  <input type="text" name="email">
</form>
```

##### 🚫 Bad

```html+erb
<form>
  <input type="text" name="name">

  <form>
    <input type="text" name="nested">
  </form>
</form>
```

##### 🚫 Bad

```html+erb
<%= form_with model: @user do |user_form| %>
  <%= form_with model: @address do |address_form| %>
    <%= address_form.text_field :street %>
  <% end %>
<% end %>
```

#### References

* [HTML Living Standard - Forms](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) 
* Inspired by [@pushcx](https://bsky.app/profile/push.cx/post/3lsfddauapk2o)

irinanazarova:
This pull request implements a new `html-no-nested-forms` rule that disallows placing one `<form>` element inside another `<form>` element, including forms rendered by Action View helpers.

```erb
<%= form_with model: @mission do |form| %>
  <%= form.submit "Update" %>
  <%= button_to "Delete", mission_path(@mission), method: :delete %>
<% end %>
```

```
`button_to` renders its own `<form>` element and cannot be nested inside another form. Move it outside of the enclosing form.
```

HTML does not support nested forms: browsers silently drop or re-parent the inner form, leading to broken submissions and buttons that submit the wrong form. The helper case is particularly easy to miss because the nested `<form>` never appears in the template. In the example above, the form rendered by `button_to` is dropped during parsing and clicking "Delete" submits the outer form instead, performing an update.

The rule detects literal `<form>` elements with a depth counter (same approach as `html-no-nested-links`) and form-rendering helpers via Prism (same approach as `actionview-no-silent-helper`). The helper list comes from the Action View helper registry via `getHelpersForTag("form")`, filtered to helpers the parser does not transform yet: once a form helper becomes `supported`, the element branch catches its synthesized `<form>` element and the Prism branch stops covering it automatically. `button_to` is included per the discussion in #166, since it is effectively a hidden form. The offense is reported on the inner form's tag name, or on the helper call for helper-rendered forms.

Nesting composed across partials (an outer form in one template, an inner form in a rendered partial) is not detectable with per-file linting. `tag.form` and `content_tag(:form)` are not detected yet; the registry's `detectStyle` metadata encodes how to find these. This could be a follow-up.

Resolves https://github.com/marcoroth/herb/issues/166


javascript/packages/linter/src/rules/html-no-nested-forms.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  visitHTMLElementNode(node: HTMLElementNode): void {
+    if (getTagLocalName(node) !== "form") {
+      super.visitHTMLElementNode(node)
+      return
+    }
+
+    if (this.formDepth > 0) {
+      this.addOffense(
+        "Nested `<form>` elements are not allowed. Move the inner `<form>` outside of the enclosing form, or associate its controls using the `form` attribute.",
```
marcoroth:
```suggestion
        "Nested `<form>` elements are not allowed. Move the inner `<form>` outside of the enclosing `<form>`, or associate its controls using the `form` attribute.",
```

javascript/packages/linter/src/rules/html-no-nested-forms.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    super.visitERBContentNode(node)
+  }
+
+  private checkNestedHelper(helperName: string, location: Location): void {
+    if (this.formDepth === 0) return
+
+    this.addOffense(
+      `\`${helperName}\` renders its own \`<form>\` element and cannot be nested inside another form. Move it outside of the enclosing form.`,
```
marcoroth:
```suggestion
      `\`${helperName}\` renders its own \`<form>\` element and cannot be nested inside another \`<form>\`. Move it outside of the enclosing \`<form>\`.`,
```

javascript/packages/linter/src/rules/html-no-nested-forms.ts
```diff
@@ -0,0 +1,132 @@
...
+        this.elementLocation(node),
+      )
+    }
+
+    this.formDepth++
+    super.visitHTMLElementNode(node)
+    this.formDepth--
+  }
+
+  visitERBBlockNode(node: ERBBlockNode): void {
```
marcoroth, line 47:
```suggestion
  // TODO: remove once we support transforming all the Action View form helpers
  visitERBBlockNode(node: ERBBlockNode): void {
```

javascript/packages/linter/src/rules/html-no-nested-forms.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    this.checkNestedHelper(helperName, node.location)
+
+    this.formDepth++
+    super.visitERBBlockNode(node)
+    this.formDepth--
+  }
+
+  visitERBContentNode(node: ERBContentNode): void {
```
marcoroth, line 63:
```suggestion
  // TODO: remove once we support transforming all the Action View form helpers
  visitERBContentNode(node: ERBContentNode): void {
```

javascript/packages/linter/test/rules/html-no-nested-forms.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    `)
+  })
+
+  test("reports the nested form location", () => {
+    expectError("Nested `<form>` elements are not allowed. Move the inner `<form>` outside of the enclosing form, or associate its controls using the `form` attribute.", [1, 12])
+
+    assertOffenses(dedent`
+      <form><div><form></form></div></form>
+    `)
+  })
```
marcoroth, line 238: lets also add tests for `form_tag`

marcoroth (commented):
Thanks @irinanazarova, this looks great!

Just a few nits and this is good to go, thank you!

marcoroth (approved):
Thank you @irinanazarova! 🙏🏼

PR #1990, opened 2026-08-04
RBS: Update signatures of `lex_file` and `parse_file`

### LINKED ISSUE #1989: RBS: Error using tapioca gem on herb because of malformed RBS on lex_file and parse_file
When using `bundle exec tapioca dsl` in our project with the latest tapioca version, the .rbi autogenerated for Herb is invalid. This happens because there is an invalid RBS signature for the API of `lex_file` and `parse_file` that are not declaring the `**` argument.

I will open a PR fixing the comment, but leave the issue for now to track it.

Image

domingo2000:
Fixes #1989 

With his we prevent having issues with tapioca
rbi generation because of the invalid
RBS for this method omitting the ** argument

### DIFF
--- a/lib/herb.rb
+++ b/lib/herb.rb
@@ -80,12 +80,12 @@ end
 
 module Herb
   class << self
-    #: (String path, ?arena_stats: bool) -> LexResult
+    #: (String path, ?arena_stats: bool, **untyped) -> LexResult
     def lex_file(path, **)
       lex(File.read(path), **)
     end
 
-    #: (String path, ?track_whitespace: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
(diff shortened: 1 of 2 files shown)


domingo2000:
@marcoroth feel free to edit the PR in any way you need. Just left here the fix that will fix our RBI tapioca generation that I tested in my local branch. 

Here is an image before and after applying the fix locally:

image

lib/herb.rb
```diff
@@ -80,12 +80,12 @@ module Herb
 
 module Herb
   class << self
-    #: (String path, ?arena_stats: bool) -> LexResult
+    #: (String path, ?arena_stats: bool, **untyped) -> LexResult
```
marcoroth, line 83: As mentioned in https://github.com/marcoroth/herb/issues/1989#issuecomment-5179274051, we probably want to remove the `**untyped` and enumerate all kwargs again in the actual method declaration:
```ruby
#: (String path, ?arena_stats: bool) -> LexResult
def lex_file(path, arena_stats: nil)
  lex(File.read(path), arena_stats:)
end
```

I wish there was a way around this.

domingo2000:
Closing because new tapioca version currently unblocks the issue.

EDIT: The issue is still a blocker in 0.10.3, but non blocker in 0.10.2

Still closed because i belive the aproach in elizamarcum PR is better.

PR #2017, opened 2026-08-05
Linter: Implement `actionview-no-implicit-polymorphic-url` rule

### LINKED ISSUE #1345: Linter Rule: Prefer explicit route helpers over implicit polymorphic URLs
# Linter Rule: Prefer explicit route helpers over implicit polymorphic URLs

**Rule:** `erb-no-implicit-polymorphic-url`

## Description

Prefer using explicit Rails route helpers (e.g., `profile_path(@profile)`) over passing model objects directly to `link_to`, which relies on implicit polymorphic routing to generate the URL.

## Rationale

When you pass a model object directly to `link_to` (e.g., `<%= link_to "Profile", @profile %>`), Rails uses `polymorphic_path` behind the scenes to infer the URL. While convenient, this implicit behavior makes it unclear what URL will be generated, a reader must know the model's class and routing conventions to determine the target path.

Implicit routing also depends on the model's class name matching a route, so renaming a model, using STI, or introducing namespaced routes can silently break the generated URL. Explicit route helpers like `profile_path` are easy to grep for across the codebase, making it straightforward to find all links to a given resource. They make the developer's intent unambiguous, reducing the chance of subtle routing bugs.

Using explicit route helpers also enables better static analysis. Tools like Herb can determine the target path from the helper name alone without needing to resolve the model's class at runtime, making it possible to detect broken links, analyze navigation structure, and provide more accurate diagnostics.

## Notes

This rule only applies when a model object (e.g., `@profile`, `@user`) is passed as the URL argument to `link_to`. Using `link_to` with an explicit named route helper like `articles_path` or a hardcoded string path like `"/articles"` is perfectly fine and will not trigger this rule.

### ✅ Good

```erb
<%= link_to "Profile", profile_path(@profile) %>
```

```erb
<%= link_to "Edit Profile", edit_profile_path(@profile) %>
```

```erb
<%= link_to @user.name, user_path(@user) %>
```

```erb
<%= link_to "View Post", post_path(@post), class: "btn" %>
```

```erb
<%= link_to "Articles", articles_path %>
```

```erb
<%= link_to "Articles", "/articles" %>
```

### 🚫 Bad

```erb
<%= link_to @profile %>
```

```erb
<%= link_to "Profile", @profile %>
```

```erb
<%= link_to @user.name, @user %>
```

```erb
<%= link_to "View Post", @post, class: "btn" %>
```

## References

* [Rails `link_to` helper documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to)
* [Rails `polymorphic_path` documentation](https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html)

marcoroth:
This pull request implements a new `actionview-no-implicit-polymorphic-url` rule that flags model objects passed directly to `link_to`, where Rails resolves the URL through `polymorphic_path` behind the scenes.

```html+erb
<%= link_to "Profile", @profile %>
```

```
Avoid passing `@profile` directly to `link_to`. The URL is resolved implicitly through polymorphic routing, so what actually gets rendered can't be read off the template or traced by tooling. Use an explicit route helper like `profile_path(@profile)`, or `polymorphic_path(@profile)` when the route has to be resolved from the model.
```

The template stops describing what it renders. `<%= link_to "Profile", @profile %>` and `<%= link_to "Profile", profile_path(@profile) %>` produce the same `<a href>`, but only the second one says so. With the first, answering "where does this link go?" means leaving the template and reconstructing the model's class and the routes it maps to. That also puts it out of reach of static analysis: the helper name is a fact about the template, while the model's class is a fact about the request that produced it.

Explicit `polymorphic_path(@profile)` and `polymorphic_url(@profile)` are not reported. The rule is about the routing being implicit, not about polymorphic routing itself, and resolving the route from the model is sometimes exactly right, such as in a partial rendered for several model types.

Resolves #1345


natematykiewicz:
Will this also catch things like `<%= link_to 'Foo', [:admin, @foo] %>`?

Some of the older code in our app did this pattern and as we refactor / delete pages, it makes it really hard to find dead links. So many times I think I've found them all and this array syntax burns me with a 500 error somewhere.

marcoroth:
@natematykiewicz it doesn't, TIL! Wow.

I think this should also be caught by this rule, would you mind opening a short issue for this and reference back to this PR? Thank you!

natematykiewicz:
Done! https://github.com/marcoroth/herb/issues/2030

Issue #1878, opened 2026-07-27
Linter: `erb-no-duplicate-branch-elements` false positive


natematykiewicz:
The `erb-no-duplicate-branch-elements` rule is great for something like:

```erb
<% if something %>
  <span>Yes</span>
<% else %>
  <span>No</span>
<% end %>
```

But when the content is more complex (many underlying nodes), you start to get false positives. For example, it's suggesting we need to refactor to deduplicate the nodes, but I cannot think of a way to actually do that. `--fix` doesn't do anything either.

```erb
<div class="space-y-4 text-center">
  <% if @mode == :upgrade %>
    <p>
      Your account will be upgraded to the higher user tier immediately after confirming 🙌.
    </p>

    <p>
      Your next invoice on
      <strong><%= @group.group_subscription.expires_at.strftime('%B %-d, %Y') %></strong>
      will include a prorated difference as well as your new plan rate for a total one-time payment of
      <strong>"$<%= item_price(@subtotal) %>"</strong>,
      plus applicable tax.
    </p>

    <p>
      After the first month, your invoice will be
      <strong>$<%= item_price(@new_plan.price) %>/<%= @new_plan.short_period %></strong>, plus applicable tax.
    </p>
  <% else %>
    <p>
      Your account will be downgraded to the new user tier immediately after confirming.
    </p>

    <p>
      Your next invoice will apply any prorated credit for the unused time on your current plan
      (usually, this means you won't pay full price for your next month after downgrading).
    </p>

    <p>
      After all remaining credit is used, your invoices will be
      <strong>$<%= item_price(@new_plan.price) %>/<%= @new_plan.short_period %></strong>, plus applicable tax.
    </p>
  <% end %>
</div>
```

<img width="1080" height="518" alt="Image" src="https://github.com/user-attachments/assets/f438143a-a623-4c96-8d58-9fdec644cd9e" />

It's noticed that both sides of the `if` have `<p>` tags (a few of them). But with all the underlying formatting and nested nodes, I don't think there's actually a way to simplify this. (The fact that both sides of the `if` have 3 `p` tags is a coincidence. I don't want to refactor in a way that expects that to always be the case.)

marcoroth:
Great catch, thanks for flagging this @natematykiewicz!

I think it would make sense if we could come up with some kind of algorithm to detect "complexity" and then abort the path if it appears to be too complex.

marcoroth:
Going to close this for now, as this seems to be resolved via #2033

[...]

PR #2033, opened 2026-08-05
Linter: Improve unextractable tags in `erb-no-duplicate-branch-elements`

### LINKED ISSUE #1570: Bug report from Herb Playground
## Bug Report

I think I’m seeing a false positive for the `erb-no-duplicate-branch-elements false positive` linter.

## Reproduction

You can reproduce this issue using the following playground link:

https://herb-tools.dev/playground/#DwUgBAlgZmUPZzCAfAKDGYAHMBjANgIYDOxAvAEQC2ALgLQDMYNApgB70BGhRAdriwpoMGAMJxerSekwB6LGhmgyYfBF4BrAPo1EFADLqNFADRg4NABYsATlqyErSNKDAt8xFs6U4CJctT0TKwcdNx8AkIyGADyVrZ4ElI0SvKKGMqqRjp6cdY2WZqm5vF2Dk4oqK4svAAm3kA

## Expected Behavior

While the `p` tags do have the same classes, their content is different and they each have unique sibling link elements, so the branches are not identical. I expect the linter to allow this.

## Actual Behavior

Linter is throwing a hint and also indicating it can autocorrect (but autocorrect doesn’t work).

marcoroth:
This pull request updates `erb-no-duplicate-branch-elements` to stop reporting shared tags that cannot actually be extracted, and makes the linter CLI stop advertising those offenses as autocorrectable.

```erb
<% if foo %>
  <p class="mt-3 text-balance">
    Content
  </p>

  <%= link_to "Link", other_path %>
<% else %>
  <p class="mt-3 text-balance">
    Other content
  </p>

  <%= link_to "Other link", other_path %>
<% end %>
```

which produced two hints plus `Fixable 2 offenses | 2 autocorrectable using --fix`, while `--fix` left the file untouched.

When the content of a shared tag differs between branches, the only way to extract it is to wrap the whole conditional in that tag. 

The rule reported the hint without checking that this rewrite was available. In the template above the `<p>` has a sibling `link_to` that has to stay inside the conditional, so wrapping would swallow the links too. The only remaining "fix" is to split the conditional into several, which duplicates the condition instead of the markup.

Resolves #1570

### DIFF
--- a/javascript/packages/linter/docs/rules/erb-no-duplicate-branch-elements.md
+++ b/javascript/packages/linter/docs/rules/erb-no-duplicate-branch-elements.md
@@ -6,6 +6,8 @@
 
 Disallow the same HTML elements wrapping content in every branch of an ERB conditional (`if/elsif/else`, `unless/else`, `case/when/else`). When all branches share identical wrapper elements, those elements should be hoisted outside the conditional. Only flags when all branches are covered (i.e., an `else` clause is present).
 
+Elements that are byte-for-byte identical in every branch can always be moved out, and are reported as warnings. A shared tag whose *content* differs between branches can only be extracted by wrapping the entire conditional in it, so it is reported as a hint, and only when that rewrite is available: every branch must consist of nothing but shared elements, and only one of them may differ in content. Anything else would require splitting the conditional into several, which duplicates the condition instead of the markup, so it is not reported.
+
 ## Rationale
 
 Duplicated wrapper elements across all branches of a conditional are unnecessary repetition. Moving them outside the conditional reduces template size, makes the structure clearer, and avoids the risk of branches getting out of sync when one is updated but others are forgotten.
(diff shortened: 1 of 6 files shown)


natematykiewicz:
Does this also resolve https://github.com/marcoroth/herb/issues/1878?

marcoroth:
@natematykiewicz looks like it does. On current main they are not reported anymore: [Playground](https://herb-tools.dev/playground/?tab=diagnostics#DwEwlgbgBAxgNgQwM5ILwCIkAcEwKYC0AngQCxQAueAHhQfgHZUBO6AfAFBRTACkUYAGZQAAgFsA9iDxRUqKAC4ArlgDmzBNKi9O3bsCy693AJoSlzKLhjmmUAO5g4cKACMZK9ZrwhKEygAWMgFgqkGWSkh4lhRg0QJiYj5gCFRwRFaCLLASDIJgzGJgDKpQgLwbgDJ7AHRcesAA9IYctfpNxqbmlgw0FAIMEBJg+FC5LXVIFMy5qmx88iLq5lhViyoA+khKrkgwzGBYsblVNFgFeEhrqVUTzFlgSQAUAOS8AELaBCAANNomTwCU2lm9Ru0yMxkczj68CUWgQUCwUw0VF84EEgmieAYw2QDjwUNxRE6UG69gRiAYUGRMkEEks8IoEgoCBcuUIsSSCIQRCSdgkgjG+lBJTY6AAJHMBFQxGtEUM8A8RJtXIzmXBATp0A1hTMvoLyZErFgsHAhghXHAZMzqDVjA0mmMDOC9ABBLLxChBKD5ZgTKCSJgBH5Eix9AbyhxOFzufXAHVsCW8eRgaWyvb4RWk2UUqpy-Aatj1SUiLMmhAMa4BOkUWXRMBSIHayZgn4mw0IY2mmDmy2UBA2x2NIx8KD4qJAx1tdpmUPWWy9SHRmQgCT2EoaaS+RmBGSkqCRD1xSz3JLgVL4jIId2WGx5ApFEq2upD5p2qfGGddHphwbDRdGk1LwYDJETpc9fF2ZJelpGIvSUBgDy3e4ZFyKAQxvCxmCxXoywYfUHkiJQWXSH5PTAJB-TwcsKJDBxcieHDuW9JQoTzGk6TQ4lulof1ck9TJshXNcvHAEp-iffQX0nZ1uDdbJiKpPAxAQYpilKSDwF6cj9yib5ONDYpw3wCj-xjdoeHjRNk1TNjMzwexs3LXN0zwAsiyTURSxzJAq2YGssDrBsdCbKYRVbOB207M0LStfsJJ4F99H4LFfB0DgGnACA2CAA)

Issue #1862, opened 2026-07-20
Problem running `bundle console`


joaoGabriel55:
During the setup of the project, I faced this error when running the `bundle console`:
```console
There was an error while trying to load the gem 'reactionview'.

Gem Load Error is: uninitialized constant ReActionView::Template::Handlers::ActionView

Backtrace for gem load error is:

/Users/quaresma/opensource/herb/vendor/bundle/ruby/4.0.0/gems/reactionview-0.3.0/lib/reactionview/template/handlers/erb.rb:6:in '<module:Handlers>'
...
```

**Solution**

Applied `require: false` for `reactionview` and `turbo-rails`

```rb
gem "reactionview", "~> 0.3.0", require: false
gem "reline", "~> 0.6"
gem "rubocop", "~> 1.71"
gem "sorbet"
gem "steep", "~> 2.0"
gem "turbo-rails", "~> 2.0", require: false
```

marcoroth:
Great catch, thanks @joaoGabriel55! 

Do you want to open a pull request for this? Thank you!

joaoGabriel55:
Sure! I will open soon

marcoroth:
Resolved via https://github.com/marcoroth/herb/pull/2040

PR #2040, opened 2026-08-06
Ruby: Problem running `bundle console`

joaoGabriel55:
## What does this PR do?

Resoles #1862 

While setting up the project, running `bundle console` failed with:

```text
uninitialized constant ReActionView::Template::Handlers::ActionView
```

This happened because the `actionview` gem was installed, but Bundler tried to require it using the gem name instead of its actual entrypoint, `action_view`. As a result, ReActionView was loaded before the `ActionView` constant was available.

`turbo-rails` was also being loaded automatically, even though it expects a fully initialized Rails application.

This PR:

- Configures `actionview` to load through `action_view`.
- Keeps ReActionView loading after Action View.
- Prevents `turbo-rails` from automatically loading in this standalone gem project.

## Verification

After these changes, `bundle console` starts successfully and both `ActionView` and `ReActionView` load as expected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gemspec
 
 gem "prism", github: "ruby/prism", tag: "v1.9.0"
 
-gem "actionview", "~> 8.1"
+gem "actionview", "~> 8.1", require: "action_view"
 gem "benchmark"
 gem "charm"
 gem "cruise"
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Sweet, thank you @joaoGabriel55! 🙏🏼

PR #2059, opened 2026-08-07
Formatter: Preserve author-expanded multi-line ERB tags

### LINKED ISSUE #1835: Formatter: Formatting of multi-line methods calls
I've been checking out the formatter and it's really interesting - so far I've come across only two things that I'm on the fence about.

The first is with really long classes which I think is actually probably fine, I just wasn't expecting it, but the second is with code like this:

```erb
# views/kaminari/_last_page.html 
<%# locals: (url:, current_page:, total_pages:, per_page:, remote:) %>

<li class="<%= class_names("page-item user-select-none", disabled: current_page.first?) %>">
  <%=
    link_to(
      sanitize(t("views.pagination.first")),
      url,
      remote:,
      class: "page-link",
      tabindex: current_page.first? ? -1 : nil
    )
  %>
</li>

```

This is how I have laid the template out, which RubyMine and `erb_lint` both are happy with - but Herb wants to squash it:

```erb
<%# locals: (url:, current_page:, total_pages:, per_page:, remote:) %>

<li class="<%= class_names("page-item user-select-none", disabled: current_page.first?) %>">
  <%= link_to(
      sanitize(t("views.pagination.first")),
      url,
      remote:,
      class: "page-link",
      tabindex: current_page.first? ? -1 : nil
    ) %>
</li>
```

`erb_lint` does not care about this, but RubyMine then wants to indent everything:

```erb
<%# locals: (url:, current_page:, total_pages:, per_page:, remote:) %>

<li class="<%= class_names("page-item user-select-none", disabled: current_page.first?) %>">
  <%= link_to(
        sanitize(t("views.pagination.first")),
        url,
        remote:,
        class: "page-link",
        tabindex: current_page.first? ? -1 : nil
      ) %>
</li>
```

after which Herb, RubyMine, and `erb_lint` all agree again.

I'm not super worried about all three tools disagreeing because that's the point of picking a formatter, but personally for this kind of code I think I prefer having more lines and balanced on both sides to make it clear where the "block of Ruby" begins and ends

Ultimately I'm wondering if this is something you'd be open to support either configuring or doing something like what Prettier does w/ objects whereby if Herb saw the newline after the opening `<%`, it'd keep it.

G-Rath:
Resolves #1835

### DIFF
--- a/javascript/packages/formatter/test/erb/multiline.test.ts
+++ b/javascript/packages/formatter/test/erb/multiline.test.ts
@@ -1,0 +1,231 @@
+import { describe, test, expect, beforeAll } from "vitest"
+import { Herb } from "@herb-tools/node-wasm"
+import { Formatter } from "../../src"
+import { createExpectFormattedToMatch } from "../helpers"
+
+import dedent from "dedent"
+
+let formatter: Formatter
+let expectFormattedToMatch: ReturnType<typeof createExpectFormattedToMatch>
+
+// https://github.com/marcoroth/herb/issues/1835
+//
+// When the source has a newline immediately after the opening tag (`<%=` or
+// `<%`), the formatter should keep the tag expanded with the opening and
+// closing delimiters on their own lines, rather than squashing the first line
+// of Ruby onto the opening delimiter. This mirrors how Prettier preserves
+// user-authored expansion of object literals, and how the formatter already
+// renders multi-line ERB comments.
+describe("@herb-tools/formatter", () => {
+  beforeAll(async () => {
+    await Herb.load()
+
+    formatter = new Formatter(Herb, {
+      indentWidth: 2,
+      maxLineLength: 80,
+    })
+
+    expectFormattedToMatch = createExpectFormattedToMatch(formatter)
+  })
+
+  describe("multi-line ERB output tags (issue 1835)", () => {
+    test("preserves newline after opening tag at top-level", () => {
+      expectFormattedToMatch(dedent`
+        <%=
+          link_to(
+            sanitize(t("views.pagination.first")),
+            url,
+            remote:,
+            class: "page-link",
+            tabindex: current_page.first? ? -1 : nil
+          )
+        %>
+      `)
+    })
+
+    test("preserves newline after opening tag inside an element", () => {
+      expectFormattedToMatch(dedent`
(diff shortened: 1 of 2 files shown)


G-Rath:
@marcoroth let me know what you think of this! I have done this with some help from AI, and ended up having to use an EC2 instance to sort all the dependencies for building, but I'm really happy with how small a patch this ended up requiring.

I've run it over a few of our codebases internally and it's working a treat: adding a newline before the tag in e.g. `<%= link_to(` has the formatter mirror that on the closing side, and vice versa if you remove the newline.

javascript/packages/formatter/test/erb/multiline.test.ts
```diff
@@ -0,0 +1,231 @@
...
+        %>">First</li>
+      `
+      const result = formatter.format(source)
+
+      expect(result).toEqual(dedent`
+        <li class="<%= class_names( "page-item", disabled: current_page.first? ) %>">
+          First
+        </li>
+      `)
+    })
```
marcoroth, line 202: I wonder if this one could possibly corrupt some Ruby 🤔

marcoroth (commented):
Thanks @G-Rath! 🙌🏼 

This indeed looks reasonable, thank you!

I only got this one comment about the squashing in the attribute value case, but otherwise I think this looks good.

javascript/packages/formatter/test/erb/multiline.test.ts
G-Rath: I've not seen any corruption from my local runs, but also [this is an existing behaviour](https://herb-tools.dev/playground/?tab=format#AQ0g8A2CWwMaQIYGdkF4BE4CkaCwAUGMfEqgPoB2iAtgKbIAUhJJGADogOZ0C00AFzo0MAGhaswAE2jJEAI0h0pALngBXAE6a6lAeU48AdADNom5AID8EyQEpbxbAD4MzgGLnL4APQxnQA) - the test here is to show that this patch does not change that.

Happy to put a todo comment in if you'd like

marcoroth: Great catch, thanks!

Lets keep it simple then and leave it as-is. 🙌🏼

marcoroth (approved):
Thank you @G-Rath! 🙏🏼

PR #2084, opened 2026-08-08
JavaScript: Migrate the build from `rollup` to `rolldown`

marcoroth:
This pull request moves every package's bundling from `rollup` to [`rolldown`](https://rolldown.rs), which is the Rust-based bundler Vite is migrating to. Bundling is by far the slowest part of a package build, so the change is worth quite a lot  in terms of bundling time.


javascript/packages/node-wasm/libherb-plugin.mjs
```diff
@@ -0,0 +1,101 @@
+const RENAMED = "__herbNodeRequire"
```
marcoroth, line 1: See: https://github.com/rolldown/rolldown/pull/10655

PR #2101, opened 2026-08-08
Tailwind Class Sorter: Make `tailwindcss` peer dependency optional

### LINKED ISSUE #1722: `@herb-tools/rewriter@0.10.1` missing `tailwindcss` peer for `@herb-tools/tailwind-class-sorter` under Yarn 4
Hi, thanks for the Herb tools package set.

I’m seeing a Yarn peer dependency warning with the current `0.10.1` releases:

- `@herb-tools/formatter`
- `@herb-tools/linter`

When running `yarn install` or `yarn up` with Yarn `4.13.0`, Yarn reports that `@herb-tools/rewriter@0.10.1` does not provide `tailwindcss` to `@herb-tools/tailwind-class-sorter@0.10.1`.

`yarn explain peer-requirements` shows:

```text
p962f40 → ✘ @herb-tools/rewriter@npm:0.10.1 doesn't provide tailwindcss to @herb-tools/tailwind-class-sorter@npm:0.10.1 [ccfb1]
```

My project already has `tailwindcss` installed directly:

```json
"dependencies": {
  "tailwindcss": "^3.4.19"
}
```

So this does not appear to break install or runtime behavior, but it does produce a persistent `YN0086` warning during normal dependency updates.

**Minimal package setup**

```json
{
  "packageManager": "yarn@4.13.0",
  "devDependencies": {
    "@herb-tools/formatter": "^0.10.1",
    "@herb-tools/linter": "^0.10.1"
  },
  "dependencies": {
    "tailwindcss": "^3.4.19"
  }
}
```

**Question**  
Would it make sense for `@herb-tools/rewriter` to declare `tailwindcss` as a dependency or peer dependency so Yarn no longer reports this warning?

Happy to test a fix if helpful.

G-Rath:
This means those of us that are not using Tailwind won't have to have `tailwindcss` installed, while those that are using the framework should not have to change anything as they'll very likely already have `tailwindcss` installed for their own use

Ideally I would prefer not having `@herb-tools/tailwind-class-sorter` installed either, but that's a larger change since people would have to explicitly start installing that.

Resolves #1722 (though technically that is already resolved 😄)

### DIFF
--- a/javascript/packages/tailwind-class-sorter/package.json
+++ b/javascript/packages/tailwind-class-sorter/package.json
@@ -52,6 +52,11 @@
   "peerDependencies": {
     "tailwindcss": "^3.0 || ^4.0"
   },
+  "peerDependenciesMeta": {
+    "tailwindcss": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">=14.21.3"
   }


marcoroth (approved):
Thanks @G-Rath!

I'd love to somehow make it completely optional. But for that we need a better story for making rewriters (and also linter rules) discoverable from NPM packages.

In an ideal world, `@herb-tools/rewriter` drops the whole `tailwind-class-sorter` rewriter and the dependency on the `@herb-tools/tailwind-class-sorter` package, and a new package provides the rewriter by wrapping the `@herb-tools/tailwind-class-sorter` package, or ship it as part of that package.

javascript/packages/tailwind-class-sorter/package.json
```diff
@@ -52,6 +52,11 @@
   "peerDependencies": {
     "tailwindcss": "^3.0 || ^4.0"
   },
+  "peerDependenciesMeta": {
```
marcoroth, line 55: Wait, hold on, shouldn't this be on the `@herb-tools/rewriter` package instead?

G-Rath: No, because it's `@herb-tools/tailwind-class-sorter` that declares the peer dependency dependency on `tailwindcss`

marcoroth: Right. I'm surprised the existing `peerDependencies` declaration wasn't enough 🤔 

Anyway, thank you @G-Rath! 🙏🏼

G-Rath: It's likely because for a while `peerDependencies` were "optional" - they were always intended to be a "required" dependency that package managers installed like `dependencies`, but around the NPM 3ish days there were a bunch of issues so they weren't automatically installed.

At the time we didn't have `peerDependenciesMeta`, so they became the way to express "we can do something with this package if its present, but we don't require it so it doesn't have to be installed always".

Later `peerDependenciesMeta` was introduced so you could proper express that, since the intention was always to make `peerDependencies` automatically installed again, which finally happened in NPM v7.

Yarn has taken the stance that this is a mistake and declared they will never automatically install peer dependencies, while PNPM I believe have made it an option and that became the default around v10 or v11?

Technically for this package this isn't an optional dependency, but this is a safe compromise for until we have that better story re discoverability, since anyone using Tailwind will have `tailwindcss` installed already either directly or via another peer dependency - the latter isn't something folks should be relying on, but it can happen; however I would be extremely surprised if anyone was depending on _only_ the `@herb-tools` packages to bring in their `tailwindcss` 😅

PR #2114, opened 2026-08-09
Formatter: Add `indentStyle` to allow formatting using tabs

### LINKED ISSUE #159: Linter Rule: Enforce consistent indentation
##### Description

Enforce consistent indentation throughout HTML+ERB templates. Maybe even with configurable indentation styles, like:

- Spaces (default: 2 spaces)
- Tabs

##### Rationale

Consistent indentation improves code readability, reduces cognitive load, and minimizes unnecessary diffs. This is especially important in mixed HTML+ERB templates, where visual structure communicates both HTML nesting and Ruby control flow.

Allowing the indentation style to be configured ensures that teams can apply their preferred style consistently across projects.

##### ✅ Good

```html+erb
<div class="container">
  <% if user %>
    <p>Welcome, <%= user.name %></p>
  <% else %>
    <p>Please sign in.</p>
  <% end %>
</div>
```

##### 🚫 Bad

```html+erb
<div class="container">
  <% if user %>
        <p>Welcome, <%= user.name %></p>
<% else %>
  <p>Please sign in.</p>
  <% end %>
</div>
```

#### References

\-

brandonweiss:
Added a `formatter.indentStyle` config option (`"space"`, the default, or `"tab"`) that the `source-indentation` rule uses to decide which character to flag and which one autofix writes.

Resolves #159

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -381,6 +381,7 @@ formatter:
 
 - **`enabled`**: `true` or `false` - Enable or disable the formatter
 - **`indentWidth`**: Number (default: `2`) - Spaces per indent level
+- **`indentType`**: `"spaces"` or `"tabs"` (default: `"spaces"`) - Indentation character expected by the `source-indentation` linter rule
 - **`maxLineLength`**: Number (default: `80`) - Maximum line length
 - **`include`**: Array of glob patterns - Additional patterns to format (additive to defaults)
 - **`exclude`**: Array of glob patterns - Additional patterns to exclude from formatting (additive to defaults)
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -363,10 +363,10 @@ Configure the formatter behavior:
 
 ```yaml [.herb.yml]
 formatter:
-  enabled: false     # Disabled by default (experimental)
-  indentWidth: 2     # Number of spaces for indentation
-  indentType: spaces # "spaces" or "tabs"
-  maxLineLength: 80  # Maximum line length before wrapping
+  enabled: false       # Disabled by default (experimental)
+  indentWidth: 2       # Number of spaces for indentation
+  indentStyle: spaces  # "spaces" or "tabs"
+  maxLineLength: 80    # Maximum line length before wrapping
 
   # Additional glob patterns to include (additive to defaults)
   include:
@@ -382,7 +382,7 @@ formatter:
 
 - **`enabled`**: `true` or `false` - Enable or disable the formatter
 - **`indentWidth`**: Number (default: `2`) - Spaces per indent level
-- **`indentType`**: `"spaces"` or `"tabs"` (default: `"spaces"`) <Badge type="tip" text="^0.11.0" /> - Character used for indentation
+- **`indentStyle`**: `"spaces"` or `"tabs"` (default: `"spaces"`) <Badge type="tip" text="^0.11.0" /> - Character used for indentation
 - **`maxLineLength`**: Number (default: `80`) - Maximum line length
 - **`include`**: Array of glob patterns - Additional patterns to format (additive to defaults)
 - **`exclude`**: Array of glob patterns - Additional patterns to exclude from formatting (additive to defaults)
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -365,7 +365,7 @@ Configure the formatter behavior:
 formatter:
   enabled: false       # Disabled by default (experimental)
   indentWidth: 2       # Number of spaces for indentation
-  indentStyle: spaces  # "spaces" or "tabs"
+  indentStyle: space   # "space" or "tab"
   maxLineLength: 80    # Maximum line length before wrapping
 
   # Additional glob patterns to include (additive to defaults)
@@ -382,7 +382,7 @@ formatter:
 
 - **`enabled`**: `true` or `false` - Enable or disable the formatter
 - **`indentWidth`**: Number (default: `2`) - Spaces per indent level
-- **`indentStyle`**: `"spaces"` or `"tabs"` (default: `"spaces"`) <Badge type="tip" text="^0.11.0" /> - Character used for indentation
+- **`indentStyle`**: `"space"` or `"tab"` (default: `"space"`) <Badge type="tip" text="^0.11.0" /> - Character used for indentation
 - **`maxLineLength`**: Number (default: `80`) - Maximum line length
 - **`include`**: Array of glob patterns - Additional patterns to format (additive to defaults)
 - **`exclude`**: Array of glob patterns - Additional patterns to exclude from formatting (additive to defaults)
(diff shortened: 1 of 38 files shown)


brandonweiss:
@marcoroth Let me know if there’s anything you’d like me to add or change!

docs/docs/configuration.md
```diff
@@ -381,6 +381,7 @@ formatter:
 
 - **`enabled`**: `true` or `false` - Enable or disable the formatter
 - **`indentWidth`**: Number (default: `2`) - Spaces per indent level
+- **`indentType`**: `"spaces"` or `"tabs"` (default: `"spaces"`) - Indentation character expected by the `source-indentation` linter rule
```
marcoroth:
```suggestion
- **`indentType`**: `"spaces"` or `"tabs"` (default: `"spaces"`) <Badge type="tip" text="^0.11.0" /> - Indentation character expected by the `source-indentation` linter rule
```

javascript/packages/config/src/config-template.yml
```diff
@@ -89,6 +89,7 @@ linter:
 formatter:
   enabled: false
   indentWidth: 2
+  # indentType: spaces  # 'spaces' or 'tabs' (used by the source-indentation linter rule)
```
marcoroth:
```suggestion
  # indentType: spaces  # "spaces" or "tabs"
```

marcoroth (commented):
Thanks, @brandonweiss!

Would you mind also following the edits in `lib/herb/configuration.rb` (for Ruby) and `rust/herb-config/` (for Rust) so we have them matching across all configuration implementations, thank you!

marcoroth:
Ideally, we could also update it in `javascript/packages/language-server/src/settings.ts` and the VS Code extension

brandonweiss:
Thanks for the review! I've mirrored the `formatter.indentType` option into the Ruby (`lib/herb/defaults.yml`) and Rust (`rust/herb-config`) config implementations so parsing/defaults stay consistent everywhere.

Before touching `language-server/src/settings.ts` and the VS Code extension, I want to check scope, since there are two different things that could mean:

1. **Config-shape parity only** — add `indentType` to the settings interfaces/VS Code options for consistency, even though it wouldn't change any behavior yet, since...
2. **Actual formatter tab output** — the `formatter.indentWidth`/`maxLineLength` settings in those two files drive the real "Format Document" command (line wrapping via the printer), not the linter. Right now the formatter only ever writes spaces; teaching it to write tabs is a separate, larger change across `format-printer.ts`, `attribute-renderer.ts`, `comment-helpers.ts`, and `indent-printer.ts` in the printer/formatter packages.

The linter's `source-indentation` rule (including fix-on-save) already picks up `indentType` correctly today via the project config, independent of these two files.

Which of these did you have in mind? Happy to take on full formatter tab support as a follow-up if that's the goal, just didn't want to guess and build the wrong thing.

marcoroth:
@brandonweiss, I feel like if we do this, we should have also have it be a proper option in the VS Code extension and probably have the formatter deal with it. 

I'd also then repurpose this pull request more in the sense of "Formatter: Add `indentType` to allow formatting using tabs", which just makes the Linter and VS Code changes an after-effect. Thank you!

docs/docs/configuration.md
```diff
@@ -381,6 +381,7 @@ formatter:
 
 - **`enabled`**: `true` or `false` - Enable or disable the formatter
 - **`indentWidth`**: Number (default: `2`) - Spaces per indent level
+- **`indentType`**: `"spaces"` or `"tabs"` (default: `"spaces"`) <Badge type="tip" text="^0.11.0" /> - Indentation character expected by the `source-indentation` linter rule
```
marcoroth: I feel like we shouldn't point out `source-indentation` here

javascript/packages/config/src/config-template.yml
```diff
@@ -89,6 +89,7 @@ linter:
 formatter:
   enabled: false
   indentWidth: 2
+  # indentType: spaces  # "spaces" or "tabs"
```
marcoroth: Since we also emit the other defaults, lets do:
```suggestion
  indentType: spaces # "spaces" or "tabs"
```

brandonweiss:
Sounds good — implemented full formatter tab support and reframed the PR title/scope accordingly. Also fixed the two open review comments (`config-template.yml` now emits `indentType` uncommented like the other defaults, and the docs no longer single out `source-indentation` since the option now governs the formatter too).

Summary of what changed:

- **`@herb-tools/printer`**: new `convertIndentation(source, indentWidth, indentType)` utility, shared by the linter's autofix and the formatter. The printer's internal line-wrapping math stays entirely in spaces (it's baked into `.length` checks across `format-printer.ts`/`attribute-renderer.ts`), so rather than rewriting that, tab output is a final pass: convert every full `indentWidth`-space run at the start of each line into a tab. Any remainder (e.g. attribute-wrapping alignment padding that isn't a whole indent level) stays as spaces, which also happens to match the common "tabs for indent, spaces for alignment" convention.
- **`@herb-tools/formatter`**: `FormatOptions.indentType`, applied as that final pass in `formatWithResult`. New `--indent-type <spaces|tabs>` CLI flag.
- **Language server**: `indentType` threaded through settings and `formatting_service.ts`, including the range-formatting de-indent/re-indent logic (which previously assumed spaces when re-applying the caller's original indentation).
- **VS Code extension**: new `languageServerHerb.formatter.indentType` setting, a `Herb: Set Indent Type` command (quick-pick, mirroring `Set Indent Width`), and the status bar/sidebar detail text now says "tabs" instead of "X spaces" when applicable.
- **Ruby/Rust**: already covered in the prior commit, no further changes needed there.

One thing worth flagging: I couldn't build the wasm toolchain (Emscripten) in my sandbox, so I wasn't able to run the actual test suite end-to-end here — I verified via `tsc -b` across the full dependency chain (core → config → printer → formatter → linter → language-server, all clean) and hand-verified the `convertIndentation` logic against the added test cases. Would appreciate CI/a local run confirming before merge.

javascript/packages/linter/src/rules/source-indentation.ts
```diff
@@ -2,28 +2,31 @@ import { Location } from "@herb-tools/core"
 
 import { BaseSourceRuleVisitor } from "./rule-utils.js"
 import { positionFromOffset } from "@herb-tools/core"
+import { convertIndentation } from "@herb-tools/printer"
 import { SourceRule } from "../types.js"
 import type { UnboundLintOffense, LintOffense, LintContext, FullRuleConfig } from "../types.js"
 
-const START_BLANKS = /^[^\S\n]*\t[^\S\n]*/
+const LEADING_BLANKS = /^[^\S\n]*/
```
marcoroth: Lets import this from `@herb-tools/printer`

brandonweiss: Ah yeah, thanks for pointing that out 👍🏼

marcoroth (approved):
Thanks @brandonweiss, I took the chance and renamed it to `identStyle` to match the option name in most other tools and aligned the values to go with it.

Thank you! 🙏🏼

brandonweiss:
@marcoroth My pleasure! Thanks for working with me on it 😄 And yeah, `indentStyle` is a better name—good call 👍🏼

PR #2170, opened 2026-08-11
Linter: Implement `erb-no-module-definitions` rule

### LINKED ISSUE #197: Linter Rule: Disallow module definitions in ERB templates
##### Description

Disallow defining Ruby `module` structures inside ERB templates. Like class definitions, modules introduce architectural logic that belongs in helper files, libraries, or initializers, not in view templates.

##### Rationale

Defining a module in an ERB file:

- Violates separation of concerns
- Pollutes the global or view context
- Cannot be autoloaded, extended, or reused properly
- Makes templates harder to read and debug
- Can introduce unpredictable state when templates are re-rendered

Move modules to their appropriate locations in `app/helpers/`, `app/`, `lib/`, or other structural Ruby files.

##### ✅ Good

```ruby
# app/helpers/display_helpers.rb

module DisplayHelpers
  def highlight(text)
    "<mark>#{text}</mark>".html_safe
  end
end
```

```html+erb
<%= highlight("Important") %>
```

##### 🚫 Bad

```html+erb
<% 
  module DisplayHelpers
    def highlight(text)
      "<mark>#{text}</mark>".html_safe
    end
  end 
%>
```

#### References

\-

markokajzer:
closes #197

**Rule:** `erb-no-module-definitions`

## Description

Disallow Ruby `module ... end` definitions inside ERB templates.

## Rationale

Modules introduce reusable application structure, which belongs in a helper, library, or initializer rather than a view. Defining one while rendering a template can pollute the view context, prevents normal autoloading, and makes the template harder to understand and reuse.

### ✅ Good

```ruby
# app/helpers/display_helpers.rb
module DisplayHelpers
  def highlight(text)
    "<mark>#{text}</mark>".html_safe
  end
end
```

```erb
<%= highlight("Important") %>
```

### 🚫 Bad

```erb
<%
  module DisplayHelpers
    def highlight(text)
      "<mark>#{text}</mark>".html_safe
    end
  end
%>

<%= highlight("Important") %>
```

```erb
<%
  decorator = Module.new do
    def highlight
      "<mark>#{text}</mark>".html_safe
    end
  end
%>
```


javascript/packages/linter/src/rules/erb-no-module-definitions.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ParserRule } from "../types.js"
+import type { ParseResult, ParserOptions } from "@herb-tools/core"
+import type {
+  UnboundLintOffense,
+  LintContext,
+  FullRuleConfig,
+} from "../types.js"
+
+const MESSAGE =
+  "Avoid defining modules in ERB templates. Move modules to their appropriate locations in `app/helpers/`, `app/`, `lib/`, or other structural Ruby files."
```
markokajzer: maybe this error message is too rails specific? open to a more general error message, e.g.


```suggestion
  "Avoid defining modules in ERB templates. Move the module to an appropriate Ruby file."
```

marcoroth: Yeah might be better, though, its less actionable imho.

javascript/packages/linter/docs/rules/erb-no-module-definitions.md
```diff
@@ -0,0 +1,59 @@
...
+### ✅ Good
+
+```ruby
+# app/helpers/display_helpers.rb
+
+module DisplayHelpers
+  def highlight(text)
+    "<mark>#{text}</mark>".html_safe
+  end
+end
```
marcoroth, line 32: lets use a different example here. Promoting `.html_safe` is not great, especially since this could become a problem if `highlight` gets called with user-controlled data.

Maybe an example with  like `pluralize` could work?

markokajzer: switched to pluralize for example and specs, and used a more generic error message 🙏 

let me know if any other changes are necessary!

i'm currently running against a few repos of https://github.com/steveclarke/real-world-rails, and i'm slowly adding more of them to have a wider base of examples. if you have a good one, feel free to suggest 👀

javascript/packages/linter/docs/rules/erb-no-module-definitions.md
```diff
@@ -0,0 +1,59 @@
+# Rule:`erb-no-module-definitions`
```
marcoroth: I think this is missing the rule "description" as the H1 here

javascript/packages/linter/docs/rules/erb-no-module-definitions.md
marcoroth: @markokajzer directly related: https://github.com/marcoroth/herb-corpus

We are running against this corpus on every commit, at least for the parse + printer, but not the linter yet.

javascript/packages/linter/docs/rules/erb-no-module-definitions.md
markokajzer: woops 😭 

added!

marcoroth (approved):
Sweet, thank you @markokajzer! 🙏🏼

Issue #990, opened 2025-12-18
Parser: Multi-ERB tag method definitions are not correctly parsed


grk:
The following erb template:
```erb
<% def greet(n) %>
Hello, <%= n %>!
<% end %>

<% greet(name) %>
```

shows this error on line 3:
```
`<% end %>` appears outside its control flow block. Keep ERB control flow statements together within the same HTML scope (tag, attribute, or content). (Herb Parser  ERB_CONTROL_FLOW_SCOPE_ERROR)
```

I guess this is not a common pattern, but works in Rails views. Should this be valid syntax in Herb?

marcoroth:
Hey @grk, thanks for reporting!

I sadly don't think that we want to support defining methods that way, as this is really confusing. If you absolutely need to, I think the best way is to define the method in a single ERB content tag instead:

```erb
<% def greet(n)
  "Hello, #{n}!"
end %>

<%= greet(name) %>
```

Alternatively:

```erb
<% def greet(n) = "Hello, #{n}!" %>

<%= greet(name) %>
```

grk:
Understandable, found this when running Herb on an existing codebase. I wonder if there would be any way to ignore those as "known errors".

marcoroth:
I'm going to close this, as I don't think we should be supporting to parse this pattern, But, now with https://github.com/marcoroth/herb/pull/2172 we can at least report and guide against this pattern. Thank you!

PR #2172, opened 2026-08-11
Linter: Implement `erb-no-method-definitions` rule

### LINKED ISSUE #195: Linter Rule: Disallow method definitions inside ERB templates
##### Description

Disallow defining Ruby methods inside ERB files using `def ... end`. View templates should only contain logic necessary for rendering the page. Defining methods in ERB mixes concerns and breaks MVC conventions.

##### Rationale

Defining methods in a view template:

- Pollutes the view context
- Makes templates harder to read and debug
- Introduces global-like behavior into views
- Encourages non-reusable, non-testable logic
- Violates Rails' separation of concerns

Method definitions belong in helpers, presenters, or view components, not in view files.

##### ✅ Good

```ruby
# app/helpers/application_helper.rb
def format_date(date)
  date.strftime("%B %d")
end

def admin?
  current_user.admin?
end
```

##### 🚫 Bad

```html+erb
<% def format_date(date) %>
  <%= date.strftime("%B %d") %>
<% end %>

<p><%= format_date(Date.today) %></p>

<% 
  def admin? 
     current_user.admin?
   end 
%>
```

#### References

\-

markokajzer:
closes #195

**Rule:** `erb-no-method-definitions`

## Description

Disallow defining Ruby methods inside ERB files using `def ... end` or `define_method`. View templates should only contain logic necessary for rendering the page. Defining methods in ERB mixes concerns and breaks MVC conventions.

## Rationale

Defining methods in a view template:

- Pollutes the view context
- Makes templates harder to read and debug
- Introduces global-like behavior into views
- Encourages non-reusable, non-testable logic
- Violates Rails' separation of concerns

Method definitions belong in helpers, presenters, or view components, not in view files.

### ✅ Good

```ruby
# app/helpers/application_helper.rb

module ApplicationHelper
  def format_date(date)
    date.strftime("%B %d")
  end

  def admin?
    current_user.admin?
  end
end
```

### 🚫 Bad

```erb
<% def format_date(date) %>
  <%= date.strftime("%B %d") %>
<% end %>

<p><%= format_date(Date.today) %></p>
```

```erb
<% def admin? = current_user.admin? %>
```

```erb
<% define_method(:admin?) { current_user.admin? } %>
```


javascript/packages/linter/test/rules/erb-no-method-definitions.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+describe("ERBNoMethodDefinitionsRule", () => {
+  test("passes without method definitions", () => {
+    expectNoOffenses(dedent`
+      <% formatted_date = date.strftime("%B %d") %>
+      <p><%= formatted_date %></p>
+    `)
+  })
+
+  test("ignores lambdas", () => {
```
markokajzer, line 23: not sure if we want to catch these

issue description was somewhat specific regarding `def ... end` blocks, but knowing ruby there are a bunch of other ways to define "methods"

marcoroth: I think it's fine to leave them for now. If'd want to catch them we could still introduce a new rule specifically for lambdas.

marcoroth (approved):
Thanks @markokajzer! 🙏🏼 

To only thing to note here is that we can't catch issues like https://github.com/marcoroth/herb/issues/990 because they are just straight up a parser error.

Though, we might be able to set `consumesParserErrors = true` on the rule itself, and then, since we can look at the extracted Ruby from that file, detect the method definition that way since then its all valid Ruby again 🤔

markokajzer:
i didnt even know that was possible haha

PR #2183, opened 2026-08-11
Linter: Implement `erb-no-return` rule

### LINKED ISSUE #1564: Linter Rule: Disallow `return` in ERB views
**Rule:** `erb-no-return`

## Description

Disallow the use of return statements inside ERB templates. Templates should not use return to control rendering flow. Use conditionals or early render guards in the controller/component instead.

## Rationale

Using return in ERB templates is a code smell that breaks the expected rendering model. Templates are meant to be declarative views, not procedural code with early exits. A return in a partial can silently abort rendering of the entire view, leading to subtle bugs and confusing behavior. It also makes templates harder to reason about since the reader must consider hidden exit points. Conditional rendering should be handled with if/unless blocks in the template or with logic in the controller/component.

### ✅ Good

```erb
<% if condition? %>
  ...
<% end %>
```

### 🚫 Bad

```erb
<% return "" unless condition? %>

...
```

markokajzer:
closes #1564

**Rule:** `erb-no-return`

## Description

Disallow `return` statements inside ERB templates. Templates should not use `return` to control rendering flow; use a conditional in the template or move the guard to the controller or component instead.

## Rationale

Using `return` in an ERB template can silently abort rendering of the surrounding view, particularly when it appears in a partial. This makes rendering flow harder to reason about and can produce subtle bugs. Templates are declarative views, so rendering decisions should use conditionals or live outside the template.

### ✅ Good

```erb
<% if condition? %>
  <p>Content</p>
<% end %>
```

### 🚫 Bad

```erb
<% return "" unless condition? %>

<p>Content</p>
```


marcoroth (approved):
Thank you @markokajzer! 🙏🏼

PR #2195, opened 2026-08-12
Language Server: Implement `textDocument/onTypeFormatting` for ERB block closers

### LINKED ISSUE #2167: Language Service: Completing a `<% ... %>` with a `do` in it, should insert an `<% end %>` automatically
The same should be true for `if`, `for`, `while`, `unless`, etc.

joaoGabriel55:
Adds LSP on-type formatting for ERB block openers. When the final `>` is typed, the language server automatically inserts a matching `<% end %>` on the next line.

This change:

- Adds an `OnTypeFormattingProvider` to the language service.
- Supports `do`, `if`, `unless`, `while`, and `for` blocks.
- Preserves the indentation of the opening ERB tag.
- Avoids inserting closers for output tags, comments, regular statements, and non-block constructs.
- Detects an existing matching `<% end %>`, including after nested blocks, to prevent duplicates.
- Advertises `documentOnTypeFormattingProvider` with `>` as its trigger character.
- Wires `textDocument/onTypeFormatting` requests through the language server.
- Adds language-service and language-server test coverage.

## How to test

Run the focused test suites:

```sh
yarn nx test @herb-tools/language-service
yarn nx test @herb-tools/language-server
```

To test manually in an editor using the Herb language server, open an `.html.erb` file and type:

```erb
<% if user.admin? %
```

After typing the final `>`, it should become:

```erb
<% if user.admin? %>
<% end %>
```

Confirm that:

- `do`, `if`, `unless`, `while`, and `for` blocks receive a closing tag.
- The closing tag matches the opener’s indentation.
- Existing matching `<% end %>` tags are not duplicated.
- Output tags such as `<%= user.name %>`, comments, and non-block statements do not receive a closing tag.

## Demo

https://github.com/user-attachments/assets/9dfa4647-a6f6-42ad-b531-cf91b96773d1

Resolves #2167


marcoroth (commented):
Thanks for taking this on @joaoGabriel55, and for the demo video!

There is one bug I would want fixed before this goes in, and I think fixing it makes the implementation even smaller and more logical to follow.

The closer is skipped when you type inside an existing block. `hasMatchingEnd` scans forward from the cursor and counts nesting, but it cannot tell whether an `end` it finds belongs to the block you just opened or to one that already encloses you. Running the logic from this PR against a few inputs:


Concretely:

```erb
<% items.each do |item| %>
  <% if item.ok? %>|
<% end %>
```

Typing that final `>` inserts nothing, because the `<% end %>` that closes `items.each` is read as already closing the new `if`. In real templates you are nearly always inside something.


The parser already knows and reports an unclosed block as `MISSING_ERB_END_TAG_ERROR`, and the node itself comes back with `end_node` unset, so a half typed template still gives a usable answer. 

That turns the whole decision into two conditions: the line you just typed closes an ERB tag, and the document is now short an `end`.

```ts
const missingEnds = (source: string) => parse(source).recursiveErrors().filter(error => error.type === "MISSING_ERB_END_TAG_ERROR").length
```

And also, `case` and `begin` would come for free, because the parser already knows they need an `end`. `isBlockOpener` and `hasMatchingEnd` both go, which are the two pieces carrying all the complexity.

Keep the indentation logic exactly as it is, since this part is already right.

Since the closer is inserted directly after the opener's line, the opener's own leading whitespace is the correct indentation by construction. `line.match(/^\s*/)` does that, and it holds up nested and inside HTML:

```erb
<% a.each do |x| %>
  <% b.each do |y| %>
    <% if y %>
    <% end %>      <- new, at the if's depth
  <% end %>
<% end %>
```

So the split is: the AST decides whether to insert, the typed line decides where. I hope this makes sense.

### Smaller things

`on_type_formatting.ts` builds a module level `const provider = new OnTypeFormattingProvider()` and exposes a free function, while every other provider is constructed in `Session` and reached through `this.session`. I think this is worth matching. The provider will still need a `HerbBackend` injected anyway once it parses, the same way `ParserService` is handed to the others.

Please let me know if you need any more guidance (and sorry for not providing more context in the first place, I pretty much just dumped that idea in an issue so I don't forget 🙈).

Either way, thank you for giving it a shot! 🙏🏼

joaoGabriel55:
Hello @marcoroth, thanks a lot for your feedback

joaoGabriel55:
Reporting `MISSING_ERB_END_TAG_ERROR` as usual:

image

marcoroth (commented):
Awesome, thank you @joaoGabriel55, this looks great!

I got one more small nit thats more behavior than code related. Right now if you complete the `>` in `<% if condition %` it inserts the `<% end %>` and puts the curser here:

```html+erb
<% if condition %>
<% end %>▮
```

I wonder if its more natural to put it like this instead:
```html+erb
<% if condition %>
  ▮
<% end %>
```
One extra newline and then indent one more level. What do you think?

joaoGabriel55:
Cool. I think that it is always good improving the DX

joaoGabriel55:
## QA - Place cursor inside completed ERB blocks

https://github.com/user-attachments/assets/59f1d42a-3f67-4050-945e-7a0e34f64822


1. Restore the required esbuild setting in `javascript/packages/vscode/esbuild.js`:

```js
platform: 'node',
mainFields: ['module', 'main'],
outfile: 'dist/extension.js',
```

2. Build the extension and updated language server:

```
yarn workspace herb-lsp build
```

3. Confirm the broken import is absent:

```
rg 'require2\("\./parser/htmlScanner"\)' \
javascript/packages/vscode/dist/extension.js
```

Expected: no output.

4. Open the extension package in VS Code:

```
code javascript/packages/vscode
```

5. Ensure the Ruby LSP extension is installed and enabled.
6. Open the Run and Debug panel, select Launch Client, and press F5.
7. In the Extension Development Host, open:

```
/Users/quaresma/opensource/herb/examples/erb.html.erb
```

Ensure the language mode is Herb (HTML+ERB).

8. Type this manually, entering the final > last:

```HTML+ERB
<% if condition %
```

After typing `>`, expect:

```HTML+ERB
<% if condition %>
  ▮
<% end %>
```

The cursor should be on the indented blank line. Also try changing VS Code indentation to four spaces or tabs and verify the inserted body follows that setting.

marcoroth (commented):
Hey @joaoGabriel55, thanks for getting this ready!

I have been testing it locally and somehow I couldn't get it to work. I didn't have time yet to investigate what could be the problem.

I wonder if it broke with the recent merges on main and that its somehow not compatible anymore now.

joaoGabriel55:
The latest commit is [Fix cursor placement for ERB block completion](https://github.com/marcoroth/herb/pull/2195/commits/cd9f951c96e9d785a383ce79aaa3786c6175852f).

### Manual QA steps

1. From the repository root, build the updated packages:

```bash
yarn workspace @herb-tools/language-service build
yarn workspace herb-lsp build
```

2. Confirm the broken bundled import is absent:

```bash
rg 'require2\("\./parser/htmlScanner"\)' \
  javascript/packages/vscode/dist/extension.js
```

Expected: no output.

3. Open the extension package:

```bash
code javascript/packages/vscode
```

4. Ensure the Shopify Ruby LSP extension is installed and enabled.

5. Open Run and Debug, select **Launch Client**, and press F5.

6. In the Extension Development Host, open:

```text
/Users/quaresma/opensource/herb/examples/erb.html.erb
```

7. Confirm the language mode in the status bar is **Herb (HTML+ERB)**.

8. Confirm **Editor: Format On Type** is enabled. The extension now enables it by default for Herb documents.

9. On a fresh line, type this manually, entering the final `>` last:

```erb
<% if condition %
```

After typing `>`, expect:

```erb
<% if condition %>
  ▮
<% end %>
```

The cursor must be on the indented blank body line.

### Verify indentation settings

Repeat the test on fresh lines after changing indentation through the VS Code status bar:

- **Spaces: 2** → body contains two spaces.
- **Spaces: 4** → body contains four spaces.
- **Indent Using Tabs** → body contains one tab.

For each attempt, undo or remove the previously generated block before typing another opener.

### Additional regression checks

Existing closer—no duplicate should be inserted:

```erb
<% if condition %
<% end %>
```

Type the final `>` on the first line. The existing `<% end %>` should remain the only closer.

Non-block ERB—no closer should be inserted:

```erb
<% user = current_user %
```

Type the final `>`. Only the ERB tag should be completed.

marcoroth (approved):
Thank you @joaoGabriel55! 🙏🏼

PR #2197, opened 2026-08-12
Linter: Improve `actionview-no-implicit-partial` offense message

### LINKED ISSUE #2196: Linter: Clarify `actionview-no-implicit-partial` offense message
Currently it reads:
```
Rails derives the partial from `to_partial_path` on `@posts` when the template renders, so the template this renders is not named in this `<%= render %>` call. Name it explicitly, with `object:` for a single record or `collection:` for many, and Herb can take you to it, check the locals you pass against its strict locals, and help you rename them.
```

for this template:
```html+erb
<%= render @posts %>
```

Image

But I think it should be more actionable and mention the expected format in the offense message:
```html+erb
<%= render partial: "...", collection: @posts %>
<%= render partial: "...", object: @post %>
```

luantaraschi:
## Summary

The `actionview-no-implicit-partial` offense now shows complete render examples for both a single object and a collection. This makes the suggested fix directly usable instead of mentioning only the `object:` and `collection:` keywords.

## Testing

- focused rule suite: 25 tests passed
- Oxlint passed for both changed files
- the full linter suite produced the same existing failures as the base commit: 23 test files, 368 tests, and 61 snapshots failed in both runs

Closes #2196

### DIFF
--- a/javascript/packages/linter/src/rules/actionview-no-implicit-partial.ts
+++ b/javascript/packages/linter/src/rules/actionview-no-implicit-partial.ts
@@ -33,7 +33,7 @@ class ActionViewNoImplicitPartialVisitor extends BaseRuleVisitor {
     const object = node.keywords?.object?.value
 
     this.addOffense(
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Thanks @luantaraschi! 🙏🏼

PR #2205, opened 2026-08-12
Linter: Update `erb-no-output-control-flow` message for end tags

### LINKED ISSUE #2203: Linter: Improve `erb-no-output-control-flow` message
File:
```html+erb
<%= tag.div do %>
  Hello
<%= end %>
```
Reports:
```
Control flow statements like `end` should not be used with output tags. Use `<% end %>` instead.Herb Linter (erb-no-output-control-flow)
```

`<% end %>` is not a "control flow statement", lets update the linter message to be specific for `<% end %>`

akira-in-tech:
## Summary

`end` isn't a control flow statement — it's just the closing keyword — so reusing the generic "Control flow statements like `end` should not be used with output tags" phrasing for it was misleading. This gives the `end` case its own message.

Before: `Control flow statements like \`end\` should not be used with output tags. Use \`<% end %>\` instead.`

After: `` `end` should not be used with an output tag. Use `<% end %>` instead. ``

Only the message for the end-tag case changes; `if`/`unless`/`elsif`/`else` keep the existing "Control flow statements like ..." wording.

Fixes #2203

## Testing

- `yarn vitest run` in `javascript/packages/linter` — 193 test files, 4295 passed (9 expected fail, 32 todo), no regressions


marcoroth (approved):
Thanks @akira-in-tech

PR #2233, opened 2026-08-14
Ruby: Reform RBS signatures of `lex_file` and  `parse_file`

### LINKED ISSUE #1989: RBS: Error using tapioca gem on herb because of malformed RBS on lex_file and parse_file

elizamarcum:
* This fixes marcoroth/herb#1989 by refining the approach taken in marcoroth/herb#1990
  * I think that this resolves the feedback you posted in marcoroth/herb#1990, but I'm happy to tweak this if it isn't quite what you were hoping for :)
* I also added two dependency lines to the `Brewfile` that were necessary for developing against this codebase
  * I'm happy to refine and/or swap those into their own PR, if you wish!

### DIFF
--- a/lib/herb.rb
+++ b/lib/herb.rb
@@ -78,6 +78,7 @@ module Herb
     end
 
     # rubocop:disable Metrics/ParameterLists
+
     #: (String path, ?track_locations: bool, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
     def parse_file(path, track_locations: ParserOptions::DEFAULT_TRACK_LOCATIONS, track_whitespace: ParserOptions::DEFAULT_TRACK_WHITESPACE, analyze: ParserOptions::DEFAULT_ANALYZE, strict: ParserOptions::DEFAULT_STRICT, action_view_helpers: ParserOptions::DEFAULT_ACTION_VIEW_HELPERS, transform_conditionals: ParserOptions::DEFAULT_TRANSFORM_CONDITIONALS, strict_locals: ParserOptions::DEFAULT_STRICT_LOCALS, prism_nodes: ParserOptions::DEFAULT_PRISM_NODES, prism_nodes_deep: ParserOptions::DEFAULT_PRISM_NODES_DEEP, prism_program: ParserOptions::DEFAULT_PRISM_PROGRAM, arena_stats: ParserOptions::DEFAULT_CAPTURE_ARENA_STATS)
       parse(File.read(path), track_locations: track_locations, track_whitespace: track_whitespace, analyze: analyze, strict: strict, action_view_helpers: action_view_helpers, transform_conditionals: transform_conditionals, strict_locals: strict_locals, prism_nodes: prism_nodes, prism_nodes_deep: prism_nodes_deep, prism_program: prism_program, arena_stats: arena_stats)
--- a/lib/herb.rb
+++ b/lib/herb.rb
@@ -79,7 +79,7 @@ module Herb
 
     # rubocop:disable Metrics/ParameterLists
 
-    #: (String path, ?track_locations: bool, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
+    #: (String path, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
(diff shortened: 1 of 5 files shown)


marcoroth (commented):
Awesome, thank you so much @elizamarcum, this is exactly the right approach in my option. A few things:

1. I think we also need to match the signatures for `parse` and `lex` in [`sig/herb_c_extension.rbs`](https://github.com/marcoroth/herb/blob/2de7b5d4e86702c81899a2a1ec586ca9c7604581/sig/herb_c_extension.rbs) (or maybe it already perfectly matches?).
2. In order to get the RBS signarures in `sig/*` re-generated you need to run `bundle exec rake rbs_inline`, so that it looks at all the RBS inline comments in the Ruby files and writes the matching `*.rbs` file for it.
3. Great catch about the `Brewfile` changes, I guess I never caught these, since I somehow already had a matching Python+Ruby version installed already, feel free to keep the changes in here. 🙌🏼 

Thank you so much for the contribution! 🙏🏼

sig/herb.rbs
```diff
@@ -4,6 +4,7 @@ module Herb
   # : (String path, ?arena_stats: bool) -> LexResult
   def self.lex_file: (String path, ?arena_stats: bool) -> LexResult
 
+  # rubocop:disable Metrics/ParameterLists
```
elizamarcum: @marcoroth I ran `bundle exec rake rbs_inline` as you suggested, and it added in this `rubocop:disable` without a matching enable.  Does that seem right to you?

lib/herb.rb
```diff
@@ -73,14 +73,16 @@
 module Herb
   class << self
     #: (String path, ?arena_stats: bool) -> LexResult
-    def lex_file(path, **)
-      lex(File.read(path), **)
+    def lex_file(path, arena_stats: ParserOptions::DEFAULT_CAPTURE_ARENA_STATS)
+      lex(File.read(path), arena_stats: arena_stats)
     end
 
+    # rubocop:disable Metrics/ParameterLists
```
marcoroth, line 80: @elizamarcum regarding the RuboCop line, I think if you add a newline between the comment and the RBS Inline signature it doesn't bring it over to the RBS file
```suggestion
    # rubocop:disable Metrics/ParameterLists

```

sig/herb.rbs
marcoroth: See: https://github.com/marcoroth/herb/pull/2233#discussion_r3865259091

lib/herb.rb
```diff
@@ -73,14 +73,16 @@
...
     #: (String path, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
-    def parse_file(path, **)
-      parse(File.read(path), **)
+    def parse_file(path, track_whitespace: ParserOptions::DEFAULT_TRACK_WHITESPACE, analyze: ParserOptions::DEFAULT_ANALYZE, strict: ParserOptions::DEFAULT_STRICT, action_view_helpers: ParserOptions::DEFAULT_ACTION_VIEW_HELPERS, transform_conditionals: ParserOptions::DEFAULT_TRANSFORM_CONDITIONALS, strict_locals: ParserOptions::DEFAULT_STRICT_LOCALS, prism_nodes: ParserOptions::DEFAULT_PRISM_NODES, prism_nodes_deep: ParserOptions::DEFAULT_PRISM_NODES_DEEP, prism_program: ParserOptions::DEFAULT_PRISM_PROGRAM, arena_stats: ParserOptions::DEFAULT_CAPTURE_ARENA_STATS)
```
marcoroth: We added `track_locations` in the meantime (via https://github.com/marcoroth/herb/pull/2199), which I think is why this is failing in CI now

sig/herb.rbs
elizamarcum: Thanks!

lib/herb.rb
elizamarcum: Makes sense- I've added that in now :)

elizamarcum: I spoke a little too soon on that and forgot to run the tests before I pushed up the final change 🤦‍♀️

lib/herb.rb
```diff
@@ -73,14 +73,17 @@
...
+    def lex_file(path, arena_stats: ParserOptions::DEFAULT_CAPTURE_ARENA_STATS)
+      lex(File.read(path), arena_stats: arena_stats)
     end
 
-    #: (String path, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
-    def parse_file(path, **)
-      parse(File.read(path), **)
+    # rubocop:disable Metrics/ParameterLists
+
+    #: (String path, ?track_locations: bool, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
```
marcoroth: Looks like we got `track_locations` listed twice here, which lead to the kwargs being generated all as `untyped` in the  `.rbs` file. 
```suggestion
    #: (String path, ?track_whitespace: bool, ?track_locations: bool, ?analyze: bool, ?strict: bool, ?action_view_helpers: bool, ?transform_conditionals: bool, ?strict_locals: bool, ?prism_nodes: bool, ?prism_nodes_deep: bool, ?prism_program: bool, ?arena_stats: bool) -> ParseResult
```
I feel like RBS Inline should warn about this.

marcoroth:
@elizamarcum we need one more `bundle exec rake rbs_inline` and then we are good to merge this. Sorry for the long back and forth 🙈

elizamarcum:
@marcoroth I think we've got it this time! 🙈🤞

marcoroth (approved):
Looks great, thank you so much @elizamarcum! 🙏🏼

PR #2234, opened 2026-08-14
Ruby: Intern token and AST node type strings

joelhawksley:
## What

Intern the type strings that the Ruby C extension attaches to every `Herb::Token` and AST node. Token and node type names come from closed enums, so identical names are now shared frozen strings instead of being allocated for every object.

Token types are cached by enum value, while each generated AST node builder caches its constant type. Both paths avoid repeated fstring-table probes after their first use.

Token values are unchanged: each value remains independently allocated and mutable, with the existing read-only `Token#value` API and RBS contract.

## Why

Type interning captures most of the allocation reduction without introducing frozen-value compatibility concerns or changing token value semantics.

## Benchmark

Measured over the [herb-corpus](https://github.com/marcoroth/herb-corpus) (36,989 `.erb` files), building both sides from source and taking the median of 5 runs. Allocations are `GC.stat(:total_allocated_objects)` deltas.

| | allocations | delta | wall clock (median) | delta |
| --- | --- | --- | --- | --- |
| `Herb.parse`, main | 61,372,804 | | 6.226s | |
| `Herb.parse`, this branch | 52,596,949 | **−14.3%** | 6.024s | −3.2% |
| `Herb.lex`, main | 85,981,356 | | 4.516s | |
| `Herb.lex`, this branch | 73,724,726 | **−14.3%** | 4.034s | −10.7% |

**21,032,485 fewer allocations (−14.3%)** across the corpus.

Lexing gains more wall clock because a token type string is a larger share of the work per token than a node type string is per node.

## Testing

- Built the Ruby native extension after regenerating templates and vendoring Prism.
- Added focused coverage that token and AST node type strings are frozen and shared, while token values remain mutable and independently allocated.
- Ran the interning, diagnostics, HTML-safe assertions visitor, and engine test files.


marcoroth (commented):
Thanks @joelhawksley! 

I was curious where the win actually comes from, so I built three variants off this branch and ran parse + lex over all 37,433 files in the corpus:

| variant | allocations | vs baseline | median |
| --- | --- | --- | --- |
| no interning | 147,906,914 | - | 11.266s |
| types only | 126,795,736 | -14.3% | 10.729s |
| types + short values | 110,954,929 | -25.0% | 10.523s |

Interning only the type names accounts for 21.1M of the 37.0M saved allocations, so 57% of the win. The short values add the remaining 15.8M.

That makes me lean towards keeping just the type interning. All four frozen-value breakages were on `token.value`, never on `token.type`. Type names come from a closed enum and nothing outside the extension writes to them, so it needs no `attr_accessor :value` and no frozen-string contract to document. It also avoids `token.value.frozen?` depending on how long the expression is.

Would you be up for splitting it, landing the type interning and then let's see what we can do about the short values as their own PR?

Thank you! 🙏🏼

joelhawksley:
@marcoroth done: https://github.com/marcoroth/herb/pull/2328

templates/ext/herb/nodes.c.erb
```diff
@@ -32,7 +32,14 @@ static VALUE rb_<%= node.human %>_from_c_struct(<%= node.struct_type %>* <%= nod
 
   AST_NODE_T* node = &<%= node.human %>->base;
 
-  VALUE type = rb_string_from_hb_string(ast_node_type_to_string(node));
+  /* Each builder handles a single node type, so its interned type name is
+     constant: intern once, pin as a GC root, then reuse by direct reference
+     instead of probing the fstring table on every node. */
+  static VALUE type = 0;
```
marcoroth:
```suggestion
  static VALUE type = 0;

```

ext/herb/extension_helpers.c
```diff
@@ -79,6 +80,44 @@ VALUE rb_string_from_hb_string(hb_string_T string) {
...
+  return rb_enc_interned_str(string.data, string.length, rb_utf8_encoding());
+}
+
+/*
+ * Token *type* names come from a fixed enum, so cache the interned String for
+ * each type keyed by its enum value. Every type name is interned once and then
+ * returned by an O(1) array lookup, avoiding a global fstring-table probe on
+ * every token. The lazy fill is race-free because token/node builders always
+ * run with the GVL held, and each cached String is pinned as a GC root.
+ */
```
marcoroth:
```suggestion
```

ext/herb/extension_helpers.c
```diff
@@ -79,6 +80,44 @@ VALUE rb_string_from_hb_string(hb_string_T string) {
   return rb_utf8_str_new(string.data, string.length);
 }
 
+/*
+ * Returns a shared, frozen, deduplicated String for the given hb_string.
+ *
+ * Backed by Ruby's global fstring table, so identical bytes always map to the
+ * same object without allocating a new String. This is ideal for type names,
+ * which are drawn from a small, fixed vocabulary and recur on every token/node.
+ */
```
marcoroth:
```suggestion
```

marcoroth (approved):
Thank you @joelhawksley! 🙏🏼

PR #2235, opened 2026-08-14
Engine: Use `String#match?` for boolean-context regexp guards in the compiler

joelhawksley:
## What

Switch three boolean-context regexp guards in `Engine::Compiler` from `=~` to
`String#match?`:

- `trailing_whitespace?` — `token[1] =~ /\s\z/`
- `leading_whitespace?` — `token[1] =~ /\A\s/`
- `at_line_start?` — `last_value =~ WHITESPACE_ONLY` and
  `last_value =~ TRAILING_INDENTATION`

## Why

These match a token value against a regexp purely for a truthy/falsy result, but
`=~` allocates a `MatchData` and assigns `$~` on every call. They run in hot
paths while scanning and optimizing the token stream, so `String#match?` returns
the same boolean without allocating.

The `=~` sites that actually consume `Regexp.last_match` (e.g.
`extract_leading_space`) are intentionally left as-is.

## Benchmark

Compiling a large real-world `.erb` view corpus (~6,000 templates, from the
github.com monolith) with `Herb::Engine` on Ruby 4.0.5: about **−0.7%**
(−152,875) total objects allocated per full-corpus compile. Small but free.

Full test suite passes (`2386 runs, 5462 assertions, 0 failures, 0 errors`).

### DIFF
--- a/lib/herb/engine/compiler.rb
+++ b/lib/herb/engine/compiler.rb
@@ -522,11 +522,11 @@ module Herb
       def trailing_whitespace?(token)
         return false unless token
 
-        token[0] == :whitespace || (token[0] == :text && token[1] =~ /\s\z/)
+        token[0] == :whitespace || (token[0] == :text && token[1].match?(/\s\z/))
       end
 
       def leading_whitespace?(token)
-        token && token[0] == :text && token[1] =~ /\A\s/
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Nice, another small change for a great win! Thank you @joelhawksley! 🙏🏼

PR #2252, opened 2026-08-15
Linter: Don't report duplicate strict locals diagnostics

### LINKED ISSUE #2236: Linter: Emits Rails-specific linter strict locals warnings/errors
I gave herb-tools/linter another shot on Ubicloud. One issue that came up were many pairs of warnings and errors like this:

```
[warning] Strict locals declarations are only supported in partials.
This file is not a partial. (actionview-strict-locals-partial-only)

[error] Strict locals (locals:) only work in partials (files starting with _).
This declaration will be ignored. (erb-strict-locals-comment-syntax)
```

Each of these pairs was for the same file and line.

First, having a both warning and an error for the same reason and same code location doesn't make sense. I realize these are different checks (`actionview-strict-locals-partial-only` and `erb-strict-locals-comment-syntax`), but if the same location has both a warning and an error for the same reason, the warning should be suppressed.

Second, this warning and error are Rails specific. Tilt supports strict locals and does not have a restriction that they only be used in partials (it doesn't even have the concept of partials), it allows them for any template. Based on discussion in #480, these warnings and errors should not be emitted by default, only if `framework: actionview`.

AG0708:
A strict locals declaration in a non-partial file was reported twice. `erb-strict-locals-comment-syntax` carried its own partial-only check, and `actionview-strict-locals-partial-only` reports the same thing for Action View projects, so a single `<%# locals: (user:) %>` in `app/views/users/show.html.erb` produced two offenses that say the same thing in different words.

`erb-strict-locals-comment-syntax` is about the syntax of the declaration, so the partial-only check is dropped from it and the Rails-specific diagnostic stays in `actionview-strict-locals-partial-only`, where it is already framework-gated and autocorrectable.

The same overlap existed one level down. `<% # locals: (user:) %>` was flagged by both `erb-comment-syntax`, for the generic "use `<%#` instead of `<% #`" reason, and by `erb-strict-locals-comment-syntax`, which has a dedicated message explaining that only ERB comment syntax is recognized for strict locals. `erb-comment-syntax` now skips content that looks like a locals declaration and leaves it to the more specific rule. The two helpers that recognize such a declaration, `extractRubyCommentContent` and `looksLikeLocalsDeclaration`, move to a shared `strict-locals-utils.ts` so both rules use the same definition.

Resolves #2236

### DIFF
--- a/javascript/packages/linter/docs/rules/erb-strict-locals-comment-syntax.md
+++ b/javascript/packages/linter/docs/rules/erb-strict-locals-comment-syntax.md
@@ -43,30 +43,43 @@ Required keyword argument:
 
 ```erb
 <%# locals: (user:) %>
+
+<%= user %>
 ```
 
 Keyword argument with default value:
 
 ```erb
 <%# locals: (user:, admin: false) %>
+
+<%= user %>
+<%= admin %>
 ```
 
 Complex default values:
 
 ```erb
 <%# locals: (items: [], config: {}) %>
+
+<%= items %>
+<%= config %>
 ```
 
 No locals (empty):
 
 ```erb
 <%# locals: () %>
+
+<p>Static content only</p>
 ```
 
 Double-splat for optional keyword arguments:
 
 ```erb
 <%# locals: (message: "Hello", **attributes) %>
+
+<%= message %>
+<%= attributes %>
 ```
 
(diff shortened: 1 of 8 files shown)


marcoroth (approved):
Thank you @AG0708!

PR #2253, opened 2026-08-16
Linter: Don't flag external scripts in `html-require-script-nonce`

### LINKED ISSUE #2237: Linter: `html-require-script-nonce` rule should ignore tags with `src` attribute
The `html-require-script-nonce` linter rule errors for `<script src="/a.js"></script>`:

```
Missing a `nonce` attribute on `<script>` tag. Use `request.content_security_policy_nonce`.
```

First, this doesn't make sense. If a `script` tag has a `src` attribute, you don't want to use a CSP nonce, you instead include the domain (or `'self'`) in the CSP `script-src`.  This error should not be issued for `script` tags with a `src` attribute.

Second, the proposed fix recommends what looks like a Rails-specific method. Unless Rails support is enabled via `framework: actionview`, the recommended fix should be generic, per discussion in #480.

https://herb-tools.dev/playground/?tab=diagnostics#DwZwxgTglgDgLgAhBMBeARAegIYDoBWI6AfMJuNPMUA

manrods:
This pull request updates `html-require-script-nonce` so it no longer flags external scripts, and tailors the recommendation in the offense message to the configured framework.

The rule reported every `<script>` element without a `nonce`, including ones that do nothing but load an external file:

```erb
<script src="/application.js"></script>
```

```
Missing a `nonce` attribute on `<script>` tag. Use `request.content_security_policy_nonce`.
```

A nonce is not what governs that tag under most policies. External scripts are covered by the `script-src` source list, so a project that allowlists `'self'` has nothing to fix here and no way to quiet the rule short of disabling it. Elements carrying a `src` attribute are now skipped.

That also covers `javascript_include_tag`, which always renders an external script. The parser synthesizes a real `src` attribute node from the first positional argument, so the same guard catches it. `<%= javascript_include_tag "application" %>` is no longer an offense, and the docs move it out of the Bad examples.

Resolves #2237
Related #480


marcoroth (approved):
Thank you @manrods! 🙏🏼

PR #2260, opened 2026-08-17
Ruby: Add RuboCop support for Ruby in ERB templates

joelhawksley:
This adds `rubocop-herb`, a RuboCop plugin that runs a project’s configured RuboCop cops against Ruby embedded in ERB templates.

The plugin uses Herb to build a position-preserving Ruby representation of the entire template and passes it through RuboCop’s native `ruby_extractors` API. This preserves control flow and variable usage across ERB tags while mapping offenses back to their original template locations. Existing configuration continues to work, including custom cops, inherited configuration, exclusions, target Ruby versions, `--only`, and safe or unsafe autocorrection.

Autocorrections are restricted to edits fully contained within one original Ruby region, preventing cops from modifying HTML or ERB delimiters. Cops that depend on physical Ruby file contents or obscured template layout are excluded for ERB templates by default. The integration lives under `gems/rubocop-herb` and is enabled through the standard RuboCop plugin configuration:

```sh
bundle add rubocop-herb
```

```yaml
plugins:
  - rubocop-herb
```

This PR implements responsibility (2) from the review discussion. Running Herb’s own linter rules as part of RuboCop remains separate future work for the `herb-linter` integration.


marcoroth (commented):
Thanks for getting this started!

I reserved the `rubocop-herb` gem name to run the Herb Linter as part of a regular RuboCop run, but I could also see that gem to have tighter RuboCop integration in the future for what you are proposing here.

I think if we'd want to land something like this feature, it probably belongs into `rubocop-herb`.

joelhawksley:
> I reserved the `rubocop-herb` gem name to run the Herb Linter as part of a regular RuboCop run, but I could also see that gem to have tighter RuboCop integration in the future for what you are proposing here.

I am suggesting the opposite (I think?): we should run RuboCop as part of Herb, not the other way around. I don't know if the semantics matter in this case though.

> I think if we'd want to land something like this feature, it probably belongs into `rubocop-herb`.

@marcoroth my mental model of the Herb linter has been that it is meant to replace ERB Lint out of the box, and ERB Lint supports RuboCop as one of its linters: https://github.com/Shopify/erb_lint#linters. I think dropping ERB Lint (which doesn't get much attention these days) is a good value proposition for adopting Herb.

I'd like to drop our use of ERB Lint at GitHub, but we use the ERB Lint RuboCop linter, so this is a blocker. Which means I'm motivated to get `rubocop-herb` working, if you're looking for someone to take the work on.

marcoroth:
Yeah, I think that `rubocop-herb` would have two responsibilities: 

1) to run Herb's own Linter rules as part of a RuboCop run (which is most likely going to be powered by the `herb-linter` gem)
2) to run the configured RuboCop rules (minus the ones that don't make sense) on the Ruby inside the ERB templates

This PR is providing the work for 2). So, if you'd want to migrate the mechanics of this pull request into `gems/rubocop-herb/*` and cover the 2) aspect I'd be more than happy to get that going. Thank you! 🙏🏼

joelhawksley:
Alright @marcoroth, I've gotten this working to the point of catching RuboCop errors in `.html.erb` files in other projects: https://github.com/discourse/discourse/pull/43828. Let me know what you think ❤️

PR #2263, opened 2026-08-17
Ruby: Release the GVL during native parsing

joelhawksley:
## Summary

Release Ruby's GVL while `herb_parse` processes a template so independent parses can run concurrently in Ruby threads.

This came out of profiling Action View template precompilation with ReActionView. Native parsing was a significant hotspot, but threaded precompilation could not parallelize it while the extension held the GVL.

## Implementation

- run `herb_parse` through `rb_thread_call_without_gvl`
- temporarily lock mutable Ruby strings while the native parser reads their buffers
- skip temporary locking for `nil` and frozen inputs
- use `rb_ensure` to unlock the source and free the parser arena if parsing or Ruby object conversion raises
- add a regression test that observes the parser thread outside the GVL and confirms mutable input is unlocked afterward

Only the native parse runs without the GVL. Converting the C AST into Ruby objects still runs with the GVL.


joelhawksley:
@marcoroth thoughts on getting this landed? I'm seeing rough a 2x speedup in per-thread performance with this change.

marcoroth (approved):
Thank you @joelhawksley! 🙏🏼

PR #2265, opened 2026-08-17
Linter: Respect `framework` when parsing Action View helpers

### LINKED ISSUE #2258: Linter: Invalid unknown HTML tag linter warning
With erb code such as this (simplified example from Ubicloud):

```erb
<tr id="<%= tag.ubid %>"></tr>
```

The Herb linter gives the following warning:

```
Unknown HTML tag `<ubid>`. This is not a standard HTML element.
```

This seems like a bug. That warning should be emitted for `<ubid></ubid>`, not for calling the `ubid` method as part of an attribute value.

Note that this erb code also results in the additional warning

```
Attribute `id` must not be empty.
Either provide a meaningful value or remove the attribute entirely.
```

This warning is also invalid, since the id attribute will have the value (assuming the `ubid` method returns it).

https://herb-tools.dev/playground/?tab=diagnostics&linterOptions=%7B%22framework%22%3A%22ruby%22%7D#DwFwTgBAxgNghgZwQXgETAKTIiOBzAOgFcAjASwBMIMA+VG4AenBqA

marcoroth:
This pull request stops the linter from reading Action View helpers into a template that isn't rendered by Action View.

On a project configured as `framework: ruby`, this template:

```erb
<tr id="<%= tag.ubid %>"></tr>
```

produced two offenses, neither of which describes anything wrong with the code:

```
Unknown HTML tag `<ubid>`. This is not a standard HTML element.
Attribute `id` must not be empty. Either provide a meaningful value or remove the attribute entirely.
```

Both come from the same cause. A lot of framework-agnostic rules ask for the `action_view_helpers` parser option so they can see a helper as the element it renders, which is what lets `html-img-require-alt` catch an `image_tag` without an `alt`. The option was requested unconditionally, so the Action View tag proxy was applied everywhere.

A rule asking for `action_view_helpers` is saying it wants to see helpers as the elements they render. Whether those helpers exist depends on the project, not the rule configuration. 

So now, `Linter#parserOptionsFor` answers that question once for every rule. Rules keep their Action View awareness on Action View projects, and see the ERB they were actually given anywhere else.

The same fix covers the plainer version of the bug, where `<%= image_tag "logo.png" %>` was reported by `html-img-require-alt` in a project with no Action View at all.

The template from the issue still reports under `framework: actionview`, and deliberately so. In Rails `tag.ubid` really is the tag proxy and really does render `<ubid></ubid>`, so Herb cannot tell that apart from a record that happens to be called `tag`.

Resolves #2258

### DIFF
--- a/javascript/packages/linter/src/linter.ts
+++ b/javascript/packages/linter/src/linter.ts
@@ -15,7 +15,7 @@ import { DEFAULT_RULE_CONFIG } from "./types.js"
 import { resolveSeverity, ALL_RULES_KEY } from "@herb-tools/config/schema"
 
 import type { RuleClass, ParserRuleClass, LexerRuleClass, SourceRuleClass, Rule, ParserRule, LexerRule, SourceRule, LintResult, LintOffense, UnboundLintOffense, LintContext, AutofixResult, RuleVersion, LinterMode, Framework } from "./types.js"
-import type { ParseResult, LexResult, HerbBackend } from "@herb-tools/core"
+import type { ParseResult, LexResult, HerbBackend, ParserOptions } from "@herb-tools/core"
 import type { RuleConfig, Config } from "@herb-tools/config"
 
 export interface LinterOptions {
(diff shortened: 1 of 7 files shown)


jeremyevans:
> The template from the issue still reports under `framework: actionview`, and deliberately so. In Rails `tag.ubid` really is the tag proxy and really does render `<ubid></ubid>`, so Herb cannot tell that apart from a record that happens to be called `tag`.

I would assume with prism you can see if `tag` happens to be a local variable and not a method call, and in that case, guess that it probably isn't a tag proxy. In the Ubicloud template, the (simplified) code is:

```erb
<% @tags.each do |tag| %>
  <tr id="<%= tag.ubid %>">
```

marcoroth:
@jeremyevans great catch, thanks! v0.11.0 will address this probably via https://github.com/marcoroth/herb/pull/1900 and https://github.com/marcoroth/herb/pull/1897.

PR #2328, opened 2026-08-20
Ruby: Intern short token values

joelhawksley:
## What

Follow-up split from #2234, as requested in [the maintainer review](https://github.com/marcoroth/herb/pull/2234#pullrequestreview-4978693074).

Intern token values of 16 bytes or fewer with `rb_enc_interned_str`. These common structural values are shared frozen strings; values over 16 bytes remain freshly allocated, independent, mutable strings. This PR intentionally does not include token or AST node type interning.

Because short token values are frozen, this also:

- makes the two compiler trim paths copy-on-write;
- makes `Token#tree_inspect` change encoding on a copy rather than the token value;
- adds a `Token#value=` writer (and RBS API) for visitors that replace values;
- updates the HTML-safe assertions visitor and diagnostics rewrite example to replace token values through that writer.

## Benchmarks

Re-benchmarked after merging the latest `main` (parse + lex over 36,989 corpus files, `GC.stat[:total_allocated_objects]`):

| variant | allocations | vs baseline |
| --- | ---: | ---: |
| no interning (pre-#2234) | 147,354,286 | - |
| main (types only) | 126,321,851 | -14.3% |
| **types + short values (this PR)** | **110,507,528** | **-25.0%** |

Short-value interning saves an incremental 15.8M allocations beyond the types-only variant on `main`.

## Testing

- Regenerated templates, re-vendored Prism, and compiled the Ruby native extension.
- Added focused coverage for the 16-byte boundary, shared/frozen short values, mutable/independent long values, mutation failure, writer replacement, and non-mutating inspection.
- Ran focused interning, diagnostics, and HTML-safe assertions visitor tests.
- Ran the full Ruby test suite and RBS generation check.
- Ran RuboCop on all changed Ruby files and clang-format verification on the changed C file.
- Resolved the conflict in `ext/herb/extension_helpers.c` against #2234 by keeping the new `rb_interned_string_from_hb_string` / `rb_token_type_value` helpers and routing `rb_token_value_from_hb_string` through them.


joelhawksley:
I'm going to close this for now as re-benchmarking has shown that it slows compilation by about a second on a benchmark I'm working on.

PR #2351, opened 2026-08-21
Linter: Extend `herb:disable` with per-rule counts

joelhawksley:
Extend the existing `herb:disable` directive with an optional per-rule count or `all` suffix so a single comment can pin the number of offenses of a given rule in a file (a ratchet), without needing a separate directive syntax.

Based on @marcoroth's [review suggestion](https://github.com/marcoroth/herb/pull/2351) in this PR.

## Syntax

```erb
<%# herb:disable rule-name %>       line-scoped (unchanged behavior)
<%# herb:disable rule-name N %>     file-scoped, expects exactly N offenses of `rule-name`
<%# herb:disable rule-name all %>   file-scoped, suppresses every offense of `rule-name`
```

Multiple entries can mix line-scoped and file-scoped forms in one comment:

```erb
<%# herb:disable rule-a, rule-b 3, rule-c all %>
```

`rule-a` is line-scoped (existing behavior); `rule-b` is a file-scoped count; `rule-c` is fully waived for the file.

## Semantics

For a file-scoped entry `rule N` with `E` = the actual number of offenses of `rule` in the file (after herb:disable line-scope filtering):

| Case | Underlying offenses | `herb-disable-comment-out-of-date` |
|------|---------------------|------------------------------------|
| `N == E` | all suppressed | not emitted |
| `N > E` | **all reported** (still surface) | emitted, autofix rewrites `N`→`E` |
| `0 < N < E` | first `N` suppressed | emitted, autofix rewrites `N`→`E` |
| `N > 0`, `E == 0` | nothing to suppress | emitted, message: "Remove the count (or the entry)" |
| `N == 0`, `E == 0` | nothing | not emitted |
| `all` | all suppressed | never emitted |

The `N > E` branch (report offenses *and* the drift) is an intentional divergence from erb_lint's counter behavior: keeping the underlying offenses visible when the file has degraded surfaces the actual problem, not just the drift.

## Meta-rules

One new meta-rule replaces the five herb-counter-comment-* rules from earlier iterations of this PR:

- **`herb-disable-comment-out-of-date`** — flags any file-scoped entry whose count disagrees with reality. Autofix rewrites the count in place.

The existing herb-disable meta-rules already cover the rest:

- **`herb-disable-comment-valid-rule-name`** — now also validates file-scoped rule names.
- **`herb-disable-comment-no-duplicate-rules`** — now also considers file-scoped entries when detecting duplicates within one comment.
- **`herb-disable-comment-malformed`** — already fires on parse failure of the whole comment.

## CLI

- `--ignore-disable-comments` now disables **both** line-scoped ignoring and file-scoped counter suppression.
- `--update-disable-counts` (renamed from `--update-counters`) rewrites file-scoped entries in place across the tracked file set: bumps counts to match reality, or removes the entry when `E == 0`.

## Design notes

- `all` is the escape hatch when you want to silence a rule for a file entirely without ratcheting. `herb-disable-comment-out-of-date` never fires on `all`.
- `<%# herb:disable rule N %>` is not accepted with a negative or non-integer `N`; the parser rejects those entries via the existing `herb-disable-comment-malformed` path.
- Line-scoped and file-scoped semantics don't mix in one entry: a rule token followed by a count/`all` is always file-scoped, regardless of where the comment appears.
- The parser exposes line-scoped and file-scoped entries as separate lists (`ruleNames` / `ruleNameDetails` and `fileScopedEntries`), so existing meta-rules that iterate line-scoped entries don't need to know about counts and vice versa.

## Related

- Original spec: https://github.com/marcoroth/herb/blob/main/javascript/packages/linter/spec/counter-based-suppression.md
- Precedent: [`erb_lint`'s `# rubocop:todo`-style file-scoped counters](https://github.com/Shopify/erb_lint#disabling-linters)


marcoroth (commented):
Thanks for working on this @joelhawksley!

I haven't seen this approach, so this is new to me. But I think I like the approach in this pull request more over the approach in https://github.com/marcoroth/herb/pull/685.

That said, I think we need to change the `<%# herb:counter %>` directive to something else. Especially since we also recently introduced `<%# herb:state %>` which deals with client-side state, so `counter` reads a lot like a client-side concept.

With that out of the way, what is speaking against extending the already existing `<%# herb:disable %>` syntax?

So you could say:
```erb
<%# herb:disable html-tag-name-lowercase 1 %>
<%# herb:disable html-tag-name-lowercase all %>
```

I think since the existing `<%# herb:disable %>` requires commas for separating the rules, we could make this work. What do you think?

joelhawksley:
@marcoroth I like it. Let me see what I can come up with 👍🏻

joelhawksley:
@marcoroth I adopted your API suggestion, I think it's nice!

joelhawksley:
@marcoroth I got this back to green ❤️

javascript/packages/linter/docs/rules/herb-disable-comment-out-of-date.md
```diff
@@ -0,0 +1,54 @@
+# herb-disable-comment-out-of-date
```
marcoroth: This needs a description of the rule as the title to match the format of the others:

```suggestion
# Linter Rule: Detect file-scoped `herb:disable rule N` entries whose count no longer matches the file.
```

javascript/packages/linter/README.md
```diff
@@ -536,6 +536,68 @@ You can disable linting for an entire file by adding the `ignore` directive anyw
 The `<%# herb:linter ignore %>` directive must be an exact match. Extra text or spacing will prevent it from working.
 :::
 
+### File-scoped counter suppression <Badge type="info" text="v0.10.4+" />
```
marcoroth:
```suggestion
### File-scoped counter suppression <Badge type="info" text="v0.11.0+" />
```

marcoroth (approved):
Thank you @joelhawksley! 🙏🏼

PR #2395, opened 2026-08-23
Only configure VS Code extension when `.vscode` exists

andyw8:
## Summary

For developers not using VS Code, `herb --init` pollutes the repository with an unwanted file.

This change makes is so `.vscode/extensions.json` is only touched when the project already has a `.vscode/` directory.


marcoroth:
Hey @andyw8, thanks for the pull request!

I'm going to close this, as I think it still helps with discovery in teams to suggest the extension. If you don't want it, you can always just delete the generated file and not commit it.

andyw8:
👍

PR #2406, opened 2026-08-25
Linter: Let user `files.include` override default excludes

joelhawksley:
Alternative fix for #1508.

When a file matches both an `include` and an `exclude` pattern, the more specific pattern now wins, comparing the leading path segments of each pattern that contain no glob characters. This lets users opt files back in (e.g. a specific `vendor/` subdirectory) without disabling the whole `vendor/**/*` default exclude.

### Behavior

- An `include` pattern overrides an `exclude` pattern when it names the same directory as that exclude or one below it. `vendor/keep/**/*` beats `vendor/**/*`; a broad `**/*.html.erb` beats nothing.
- An `exclude` with no literal prefix (`**/*.generated.html.erb`) is never overridable, since it selects files by name and not by location.
- A file has to out-specify every `exclude` pattern it matches to be kept.
- The rule is the same at every level, so `linter.include` and `formatter.include` follow it too. Previously a match on either bypassed parent excludes unconditionally, so a broad `linter.include` switched off the default excludes.
- Tool-level excludes (`linter.exclude`, `formatter.exclude`) take part in the same comparison, so they are no longer applied unconditionally on top.
- No change when no `include` out-specifies an exclude, so default excludes continue to filter as before.

### Example

With the default `vendor/**/*` exclude, this now works:

```yaml
files:
  include:
    - "vendor/keep/**/*.html.erb"
```

Files under `vendor/keep/` are linted; other `vendor/` and `node_modules/` files remain excluded.

### Changes

- `lib/herb/configuration.rb`: add `include_overrides_exclude?` and `literal_prefix_segments`, and give `path_excluded?` the include patterns so it can apply them.
- `javascript/packages/config/src/config.ts` and `rust/herb-config/src/{config,glob}.rs`: the same rule, so all three implementations agree on a given `.herb.yml`.
- `lib/herb/dev/watcher.rb`: pass the include patterns through, so the dev server watches the files `find_files` indexes.
- `sig/herb/configuration.rbs`: signatures for the above.
- Tests: 10 new in `test/configuration_test.rb`, 10 in `javascript/packages/config/test/config.test.ts`, 9 in `rust/herb-config/tests/config_test.rs`.
- `docs/docs/configuration.md`: add an "Include Precedence" section and rewrite the "Including Previously Excluded Files" tip.


joelhawksley:
@marcoroth as an aside, https://github.com/marcoroth/reactionview/pull/115 did not work for us, as we do not use Bundler at boot time.

marcoroth (commented):
Thank you @joelhawksley!

PR #2411, opened 2026-08-26
Linter: Fix SVG false positives in `html-no-block-inside-inline`

FrancescoK:
- allow SVG embedded content inside elements classified as inline
- skip SVG descendants rather than classifying them using HTML block/inline sets
- add regression coverage and document SVG as a valid example

`svg` is valid embedded phrasing content, but the rule currently classifies it as unknown. SVG descendants also belong to the SVG namespace and should not be checked against HTML element categories.

This preserves the rule's existing handling of block, unknown, custom, and anchor elements.

Addresses #267.
Related #291

### DIFF
--- a/javascript/packages/linter/src/rules/html-no-block-inside-inline.ts
+++ b/javascript/packages/linter/src/rules/html-no-block-inside-inline.ts
@@ -7,6 +7,7 @@ import type { HTMLOpenTagNode, HTMLElementNode, ParseResult } from "@herb-tools/
 
 const SVG_HTML_INTEGRATION_POINTS = new Set(["foreignobject", "desc", "title"])
 
+// TODO: refactor using ElementStackVisitor
 class BlockInsideInlineVisitor extends BaseRuleVisitor {
   private inlineStack: string[] = []
   private insideSVG = false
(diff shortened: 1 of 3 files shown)


FrancescoK:
@codex can you review again?

javascript/packages/linter/src/rules/html-no-block-inside-inline.ts
```diff
@@ -52,6 +66,25 @@ class BlockInsideInlineVisitor extends BaseRuleVisitor {
...
+        this.insideSVG = true
+        return
+      }
+
+      super.visitHTMLElementNode(node)
+      return
+    }
+
+    // SVG is valid embedded phrasing content. Traverse its descendants without
+    // classifying SVG-native elements using HTML block/inline element sets.
```
marcoroth:
```suggestion
```

marcoroth (approved):
Thanks @FrancescoK, I'm happy to merge this in the meantime, but I think long-term we want to deprecate this rule in favor of a rule like the one proposed in #1030.

Thank you! 🙏🏼

javascript/packages/linter/src/rules/html-no-block-inside-inline.ts
```diff
@@ -5,8 +5,11 @@ import { isHTMLOpenTagNode } from "@herb-tools/core"
 import type { UnboundLintOffense, LintContext, FullRuleConfig } from "../types.js"
 import type { HTMLOpenTagNode, HTMLElementNode, ParseResult } from "@herb-tools/core"
 
+const SVG_HTML_INTEGRATION_POINTS = new Set(["foreignobject", "desc", "title"])
+
```
marcoroth, line 9:
```suggestion

// TODO: refactor using ElementStackVisitor
```

FrancescoK:
@marcoroth I agree, though the work in #1030 is not quite in sync with the codebase by now, and a single rule would be pretty massive and cover/replace/extend existing work. Would you mind if I open a tracking issue with a proposal for a family of smaller rules?

PR #2437, opened 2026-08-28
Highlighter: Color `<%#` and its content as a comment

### LINKED ISSUE #2434: Highlighter: Color `<%#` as its content as as a comment
Currently they are treated the same way as `<%` and `<%=`. This should include both the TypeScript and Rust implementation.

ethanhawkes-gif:
Closes #2434

## Problem

ERB comment tags are highlighted exactly like `<%` and `<%=`. The opening tag gets the ERB tag color, and the body is run through the Ruby keyword highlighter — so in `<%# if end %>` the words `if` and `end` light up as Ruby keywords inside what is only a comment.

## Approach

`<%#` is already its own `TOKEN_ERB_START` value — it's listed in the lexer's `erb_open_patterns` (`src/lexer.c`), so the token stream for `<%# note %>` is:

```
TOKEN_ERB_START    "<%#"
TOKEN_ERB_CONTENT  " note "
TOKEN_ERB_END      "%>"
```

That means the renderer can key off the opening tag without any lexer or parser change. It tracks whether the current ERB tag is a comment and then colors the opening tag, the content and the closing `%>` with the comment color, skipping Ruby keyword highlighting for the body.

One detail worth calling out: the flag is re-decided on **every** `TOKEN_ERB_START`, so it never has to be cleared on `TOKEN_ERB_END`. That is what keeps the closing `%>` inside the comment, while a regular ERB tag that follows a comment still highlights normally — `<%# note %><% if x %>` colors the first tag as a comment and the second as ordinary ERB with `if` as a keyword.

Applied to both the TypeScript and the Rust implementation, as the issue asks.

## Color choice

This reuses the existing `TOKEN_HTML_COMMENT_START` color rather than introducing a new theme key, so all five bundled themes and any custom theme work unchanged — every theme already defines a muted comment color. If you'd rather ERB comments be themable separately, I'm happy to follow up with an optional `TOKEN_ERB_COMMENT` key that falls back to this one.

## Tests

Three new cases in each implementation, covering the comment tag itself, keyword suppression inside the body, and the "regular tag after a comment" reset. They use explicit assertions rather than snapshots so they document the expected colors directly.

- `javascript/packages/highlighter/test/syntax-renderer.test.ts`
- `rust/herb-highlighter/tests/syntax_renderer_test.rs`

Rust suite run locally — **186 passed, 0 failed** for `cargo test -p herb-highlighter`, including every pre-existing `insta` snapshot test, so nothing was invalidated:

```
test highlights_an_erb_comment_tag_and_its_content_as_a_comment ... ok
test does_not_highlight_ruby_keywords_inside_an_erb_comment ... ok
test still_highlights_a_regular_erb_tag_that_follows_a_comment ... ok
...
test result: ok. 13 passed; 0 failed  (syntax_renderer_test)
```

For the TypeScript side I couldn't build the wasm backend locally (no Emscripten on this machine), so I verified the renderer against the real lexer by driving `SyntaxRenderer` with the published `@herb-tools/node-wasm` token stream, and ran the new tests plus the existing `comment state tracking` tests — all passing. I also checked that no committed snapshot contains `<%#`, so none of them should shift.

### DIFF
--- a/javascript/packages/highlighter/src/syntax-renderer.ts
+++ b/javascript/packages/highlighter/src/syntax-renderer.ts
@@ -8,6 +8,8 @@ import type { ColorScheme } from "./themes.js"
 const HIGHLIGHTED_METHODS = ["raise"]
 const HIGHLIGHTED_WORDS = new Set([...RUBY_KEYWORDS, ...HIGHLIGHTED_METHODS])
 
+const ERB_COMMENT_TAG_OPENING = "<%#"
+
 type SyntaxRenderState = {
   inTag: boolean
   inQuotes: boolean
@@ -17,6 +19,7 @@ type SyntaxRenderState = {
   expectingAttributeName: boolean
   expectingAttributeValue: boolean
   inComment: boolean
+  inERBComment: boolean
 }
 
 export class SyntaxRenderer {
@@ -97,6 +100,7 @@ export class SyntaxRenderer {
       expectingAttributeName: false,
       expectingAttributeValue: false,
       inComment: false,
+      inERBComment: false,
     }
 
     for (let i = 0; i < tokens.length; i++) {
@@ -114,7 +118,7 @@ export class SyntaxRenderer {
 
       const color = this.getContextualColor(state, token, tokenText)
 
-      if (token.type === "TOKEN_ERB_CONTENT") {
+      if (token.type === "TOKEN_ERB_CONTENT" && !state.inERBComment) {
         const highlightedRuby = this.highlightRubyCode(tokenText)
         highlighted += highlightedRuby
       } else if (color !== undefined) {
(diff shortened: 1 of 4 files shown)


javascript/packages/highlighter/src/syntax-renderer.ts
```diff
@@ -204,6 +208,12 @@ export class SyntaxRenderer {
       case "TOKEN_HTML_COMMENT_END":
         state.inComment = false
         break
+
+      // Every ERB tag re-decides this, so the flag never has to be cleared on
+      // `TOKEN_ERB_END` — which keeps the closing `%>` inside the comment.
+      case "TOKEN_ERB_START":
+        state.inERBComment = tokenText === ERB_COMMENT_TAG_OPENING
```
marcoroth, line 215: @ethanhawkes-gif when do we set `state.inERBComment` back to `false`?

javascript/packages/highlighter/src/syntax-renderer.ts
```diff
@@ -212,6 +222,15 @@ export class SyntaxRenderer {
     token: Token,
     tokenText: string,
   ): Color | null {
+    if (
+      state.inERBComment &&
+      (token.type === "TOKEN_ERB_START" ||
+        token.type === "TOKEN_ERB_CONTENT" ||
+        token.type === "TOKEN_ERB_END")
+    ) {
+      return this.colors.TOKEN_HTML_COMMENT_START
```
marcoroth, line 231: I think we want a different color for ERB comments, usually HTML comments are green and ERB comments gray.

marcoroth:
Closing in favor of https://github.com/marcoroth/herb/issues/2434, thanks @ethanhawkes-gif

PR #2440, opened 2026-08-28
Highlighter: Color `<%#` and its content as a comment

alexsmolya:
Relates to #2434

ERB comment tags now render the opening tag, content, and closing tag with the existing comment color in both the TypeScript and Rust highlighters. Ruby keyword highlighting remains unchanged for regular ERB blocks.

The Rust snapshot records the ANSI-colored output produced with `with_color()`.

Validation:
- `git diff --check` passed.
- Rust `rustfmt --check` passed.
- TS/Rust structural parity check passed.
- Focused Rust test could not complete because the project build script requires `bundle show prism` and Bundler is unavailable.
- Prettier check reports the two touched TypeScript files as not clean; no unrelated formatting churn was applied.

### DIFF
--- a/javascript/packages/highlighter/src/syntax-renderer.ts
+++ b/javascript/packages/highlighter/src/syntax-renderer.ts
@@ -213,7 +213,7 @@ export class SyntaxRenderer {
         break
 
       case "TOKEN_ERB_START":
-        state.inErbComment = tokenText.startsWith("<%#")
+        state.inErbComment = state.inErbComment || tokenText.startsWith("<%#")
         break
     }
   }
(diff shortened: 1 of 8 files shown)


marcoroth (commented):
Thanks @alexsmolya!

Looks like there is one more bug:

Image


after a `<%#` comment, the start of the `<!--` the `<` is not colored in the same way as the rest of the HTML comment.

CleanShot 2026-08-28 at 12 58 01@2x

alexsmolya:
Thanks — I checked this against the current PR head. With the minimal transition `<%# comment %><!-- html comment -->`, the lexer emits the full `<!--` as a single HTML-comment-start token, and I can't reproduce the differently colored `<` from that case. Could you share the exact source snippet/context from the screenshot? There may be another tokenization or rendering path involved.

marcoroth (approved):
Thanks @alexsmolya!

PR #2553, opened 2026-09-05
VS Code: Stop charging Node startup to the Project Analysis timeout

### LINKED ISSUE #2552: Project Analysis: per-file 1 s spawn timeout counts Node/WASM startup, so most files show as Timed Out
## Summary

Project Analysis in the VS Code extension reports most files as **Timed Out** on a Rails app with ~460 templates, although none of them is slow to parse.

## What happens

`Project Analysis` spawns a fresh Node process per file:

```
process.execPath dist/parse-worker.js <file> <linter> <formatter> <indent> <maxLen> <rules> <root>
```

with `os.cpus().length` of them running concurrently, each under `execFile(..., { timeout: 1000 })`. A killed child is counted as a timeout.

Every spawn has to boot Node and load the ~3 MB WASM bundle (`@herb-tools/node-wasm`, Prism) before it parses anything. Measured on an Apple Silicon Mac (Node 26.5.1), one spawn of `parse-worker.js` on a 5 KB template takes:

| run | time |
|---|---|
| linter on, formatter off | 0.45–0.57 s |
| linter on, formatter on | 0.71 s |
| bare `node -e 0` | 0.08 s |

So the fixed cost is almost the whole 1 s budget before the file is even read. With a dozen of these booting at once, plus anything else on the CPU, a large share cross the second and are reported as timeouts. The count swings from run to run (452, 238, 275 out of 460 on the same tree) because it tracks machine load, not the files.

The templates themselves are fast: the native parser (herb gem 0.10.3, `Herb.parse` with `track_whitespace: true`) needs at most **24 ms** on the largest template (80 KB) and a median of 0.8 ms across all 484 files. `herb-lint` from the CLI checks all 460 files in about 18 s with no timeouts.

## Suggestions

- Start the timeout after the worker signals it is ready, so Node/WASM startup is not charged to the file.
- Or keep one worker (or a small pool) alive for the whole batch instead of one process per file.
- Or expose the limit as a setting; there is no `languageServerHerb.*` key for it today.

Raising the literal to 15 s in `dist/extension.js` makes every file pass on the same machine, which confirms the parse itself is not the problem.

## Environment

- Extension `marcoroth.herb-lsp` 0.10.3 (also the Cursor build, `0.10.3-universal`)
- Worker reports `@herb-tools/node-wasm@0.10.3, @herb-tools/core@0.10.3, libprism@1.9.0, libherb@0.10.3 (WebAssembly)`
- macOS 26 on Apple Silicon, Node v26.5.1, VS Code

assirims:
Fixes #2552.

## Problem

Project Analysis in the VS Code extension spawns a fresh Node process per file (`os.cpus().length` at once) and kills each after **1 s**. Every worker first has to boot Node and load the WASM bundle before it reads the file: ~0.45 s on an idle Apple Silicon Mac, ~0.7 s with the formatter enabled. Under any CPU contention most of a large project is reported as "Timed Out". On a 460-template Rails app the count was 452, 238 and 275 on the same tree, tracking machine load rather than the files; the native parser needs at most 24 ms on the largest template.

## Change

- **`analysis-service.ts`** — the process-level limit becomes a 30 s hang guard (`WORKER_TIMEOUT_MS`) with a comment on why it must not be tight. The parse itself is already bounded by `ParserOptions.timeout` (1 s by default), so runaway parses are still caught.
- **`parse-worker.js`** — reports `timedOut: true` when the parse result carries a `TIMEOUT_ERROR`, and the service maps that to the `timeout` status. The "Timed Out" group therefore keeps meaning "the parser gave up on this file" rather than "Node did not finish booting".

No new settings, no change to the worker's arguments.

## Verification

- With the bundled `extension.js` patched to the same larger timeout, all 460 files pass Project Analysis on the machine that reported the timeouts.
- `node --check` on the worker; the TypeScript change is a constant, one option, and one ternary.
- I could not run the monorepo build locally (it needs the native toolchain), so CI is the typecheck here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/javascript/packages/vscode/src/analysis-service.ts
+++ b/javascript/packages/vscode/src/analysis-service.ts
@@ -8,6 +8,23 @@ import type { AnalysisResult, FileStatus } from './types'
 
 const execFileAsync = promisify(execFile)
 
+/**
+ * How long a parse worker may run before it is killed and the file reported as
+ * timed out.
+ *
+ * This only has to catch a worker that hangs. The parse itself is already
+ * bounded: the parser enforces `ParserOptions.timeout` (1 s by default) and a
+ * parse that overruns it comes back with a `TIMEOUT_ERROR`, which the worker
+ * reports as `timedOut`.
+ *
+ * It must not be tight, because every worker is a fresh Node process that has
+ * to boot and load the WASM bundle before it reads the file. That costs about
+ * half a second on an idle Apple Silicon Mac, more when `cpus()` workers start
+ * at once. With a 1 s limit most of a large project was reported as "Timed Out"
+ * on startup cost alone (#2552).
+ */
+const WORKER_TIMEOUT_MS = 30_000
+
 export class AnalysisService {
   private workerPath: string
   private onVersionUpdate?: (version: string) => void
@@ -56,7 +73,7 @@ export class AnalysisService {
         JSON.stringify(linterRules),
         workspaceRoot,
         formatterIndentStyle
-      ], { timeout: 1000 })
+      ], { timeout: WORKER_TIMEOUT_MS })
 
       const result = JSON.parse(stdout.trim())
       const failed = result.errors > 0 || result.lintErrors > 0
@@ -66,7 +83,7 @@ export class AnalysisService {
       }
 
       return {
-        status: failed ? 'failed' : 'ok',
+        status: result.timedOut ? 'timeout' : failed ? 'failed' : 'ok',
         errors: result.errors as number,
         lintWarnings: result.lintWarnings as number || 0,
         lintErrors: result.lintErrors as number || 0,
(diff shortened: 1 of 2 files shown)


javascript/packages/vscode/src/analysis-service.ts
```diff
@@ -8,6 +8,23 @@ import type { AnalysisResult, FileStatus } from './types'
...
+ * bounded: the parser enforces `ParserOptions.timeout` (1 s by default) and a
+ * parse that overruns it comes back with a `TIMEOUT_ERROR`, which the worker
+ * reports as `timedOut`.
+ *
+ * It must not be tight, because every worker is a fresh Node process that has
+ * to boot and load the WASM bundle before it reads the file. That costs about
+ * half a second on an idle Apple Silicon Mac, more when `cpus()` workers start
+ * at once. With a 1 s limit most of a large project was reported as "Timed Out"
+ * on startup cost alone (#2552).
+ */
```
marcoroth:
```suggestion
```

marcoroth (commented):
Thanks @assirims!

PR #2614, opened 2026-09-10
Linter: Apply the strict locals rules to every template

### LINKED ISSUE #2501: Linter Rule: strict locals for non-partial templates
**Rule:** `erb-strict-locals-required`, `actionview-strict-locals-first-line`, and others.
`actionview-strict-locals-partial-only` should probably be removed.

In terms of best practices, maybe we should have a linter rule that ensures all (non-partial) templates take no locals (and declare as such)? (rationale: instance variables should be used instead for regular templates, `content_for` should be used for layouts).

## Description

The "strict locals" feature of Rails does not apply solely to partials, but to all templates. Any template can be rendered with locals.

The strict-locals-related linters and their documentation assume it only applies to partials, it should be possible to ensure all templates use strict locals for example. And the linter checking that strict locals are not used in non-partial templates seems unnecessary, at the very least its documentation shouldn't say that strict locals have no effect.

## Rationale

Strict locales also apply to non-partial templates, including layouts.

However the locals are shared between the layout rendering and the template rendering: if a local is passed to a template, it also needs to be accepted by the layout (e.g. `locals: (**)` to allow templates using that layout to receive any local). Conversely if a layout needs an optional local, all templates need to accept it, or it won't be possible to render the layout with that template.

## Solution

I don't have a strong opinion on what linters there should be to handle these, but I just wanted to report the issue with the linters here. I feel like "templates/layouts should take no locals" is a good rule, it should be possible to enforce these while still using strict locals.

ActionView should probably also have a feature that allows to assume empty strict locals for templates/layouts, without requiring `<%# locals: () %>` at the top of every of those files. Then in that case the `actionview-strict-locals-partial-only` could be useful.

marcoroth:
This pull request updates the strict locals rules to match how Action View actually treats them, by widening `actionview-strict-locals-first-line` to any template that carries a declaration and rebuilding `actionview-strict-locals-partial-only` as an opt-in convention rule.

Resolves https://github.com/marcoroth/herb/issues/2501


etiennebarrie:
Sorry I just got to testing it. This is missing changes to erb-strict-locals-required, in the sense that it's still not causing errors or fixing my non-partial templates.

Issue #1972, opened 2026-08-03
Linter: `erb-no-unused-literals` false positives on `case`/`when` and `case`/`in` conditions


siemensikkema:
Since `0.10.3`, `erb-no-unused-literals` flags the comparison literals of a `case` statement as unused. They aren't unused — they're the operands the `case` subject is compared against.

## Reproduction

`a.html.erb`:

```erb
<% case value
   when "a"
     result = 1
   end %>
```

```
$ herb-lint --simple a.html.erb
a.html.erb:
  2:8  ✗ Avoid using silent ERB tags for literals. `"a"` is evaluated but never used or output. (erb-no-unused-literals)
```

No `.herb.yml` and no other flags needed — exits `1`.

**Expected:** no offense. `"a"` is the `when` condition.

**Actual:** one offense per literal condition.

## Versions

```
@herb-tools/linter@0.10.3
@herb-tools/printer@0.10.3
@herb-tools/node-wasm@0.10.3
@herb-tools/core@0.10.3
```

Node.js v26.5.0. Reproduces with `--only erb-no-unused-literals`, so it's not a config interaction.

## Regression range

`0.10.2` reports **0 offenses** on the exact same file. `0.10.3` reports 1.

The cause looks like #1885 ("Core: Fix `PrismVisitor` skipping node list fields", merged Jul 29, first released in 0.10.3). Its description notes that visitors previously "never descended into `when`/`in` branch conditions" and that fixing this "fixes false negatives in ... `erb-no-unused-literals`".

But `LiteralCollector` in `javascript/packages/linter/src/rules/erb-no-unused-literals.ts` is a blocklist-based visitor — it overrides `visitCallNode` to skip arguments, and stops traversal into every assignment node type plus `visitReturnNode` / `visitBreakNode` / `visitNextNode`. It has no `visitWhenNode`, `visitInNode`, `visitCaseNode` or `visitCaseMatchNode`. So once the visitor started descending into node list fields, `when` conditions began landing in `literals` for the first time. The rule bug was always there; #1885 just made it reachable.

## Affected forms

| Input | Offenses | Correct? |
|---|---|---|
| `case v; when "a" then r = 1; end` | 1 | ✗ false positive |
| `case v; when "a", "b" then r = 1; end` | 2 | ✗ false positive |
| `case v; when 1 then r = 1; end` | 1 | ✗ false positive |
| `case v; when :sym then r = 1; end` | 1 | ✗ false positive |
[...]

PR #2637, opened 2026-09-13
Linter: Don't flag conditional predicates in `erb-no-unused-literals`

### LINKED ISSUE #2636: Linter: `erb-no-unused-literals` flags literals in `if`/`unless`/`while`/`until` conditions
`erb-no-unused-literals` reports a literal as "never used or output" when the literal sits in the condition of an `if`, `unless`, `while` or `until`. The literal is used: it is the receiver of the call the condition evaluates.

This is the same shape as #1972, one level up. `LiteralCollector` has overrides for `CaseNode`, `CaseMatchNode`, `WhenNode` and `InNode`, so `case` subjects and `when` operands are skipped, but it has no override for `IfNode`, `UnlessNode`, `WhileNode` or `UntilNode`. Default traversal reaches their predicate, and `visitCallNode` then collects the call's receiver.

Collecting a receiver is right when the whole tag is that expression, which is what `<% "hello".upcase %>` and `<% [1, 2, 3].length %>` already assert. It is wrong when the call's value is consumed by a conditional.

## Reproduction

`a.html.erb`:

```erb
<% level = "info" unless %w[debug info warn].include?(level) %>
```

```
$ herb-lint --only erb-no-unused-literals --simple a.html.erb
a.html.erb:
  1:25 ✗ Avoid using silent ERB tags for literals. `%w[debug info warn]` is evaluated but never used or output. (erb-no-unused-literals)
```

Exits `1`. No `.herb.yml` and no other flags needed.

**Expected:** no offense. `%w[debug info warn]` is the receiver of `.include?`, and the result is the `unless` condition.

**Actual:** one offense per literal in a conditional's predicate.

## Shapes

Checked against `main` at 9d87e97, each as its own `expectNoOffenses` case:

| ERB | Offense | Correct |
| --- | --- | --- |
| `<% l = "i" unless %w[a b].include?(l) %>` | yes | no |
| `<% l = "i" if %w[a b].include?(l) %>` | yes | no |
| `<% l = "i" while %w[a b].include?(l) %>` | yes | no |
| `<% if %w[a b].include?(l)` … `end %>` | yes | no |
| `<% unless %w[a b].include?(l)` … `end %>` | yes | no |
| `<% while %w[a b].include?(l)` … `end %>` | yes | no |
| `<% until %w[a b].include?(l)` … `end %>` | yes | no |
| `<% f = %w[a b].include?(l) %>` | no | yes |
| `<% if %w[a b].include?(l) %>` … `<% end %>` | no | yes |

The second-to-last row is the same call inside an assignment, where traversal already stops. The last is the condition on ERB tags rather than inside one silent tag. Those two are what keep this out of most templates: it surfaces on a modifier conditional, or on a conditional written inside a single silent tag.

## Versions

Reproduces on `main` at 9d87e97, and on the current release:

```
@herb-tools/linter@0.10.4
@herb-tools/printer@0.10.4
@herb-tools/node-wasm@0.10.4
@herb-tools/core@0.10.4
libprism@1.9.0
libherb@0.10.4 (WebAssembly)
```

Node.js v22.22.0.

viktorianer:
`LiteralCollector` in `erb-no-unused-literals` has no handling for `IfNode`, `UnlessNode`, `WhileNode` or `UntilNode`, so the default traversal descends into their predicate. `visitCallNode` then visits the call's receiver, and a literal receiver inside a condition is collected and reported as unused.

```erb
<% level = "info" unless %w[debug info warn].include?(level) %>
```

```
a.html.erb:
  1:25 ✗ Avoid using silent ERB tags for literals. `%w[debug info warn]` is evaluated but never used or output. (erb-no-unused-literals)
```

`%w[debug info warn]` isn't unused, it's the receiver of `.include?` and the result is the `unless` condition.

This is the same shape as the `case` operands in #1972, one level up: the collector already skips `case` subjects and `when`/`in` operands, and the four conditional node types were the remaining consumed positions. The fix follows the same pattern, visiting the branches and body while skipping the predicate.

Collecting a literal receiver stays correct where the whole tag is that expression, so `<% "hello".upcase %>`, `<% 1 + 2 %>` and `<% [1, 2, 3].length %>` are unaffected. A ternary is an `IfNode`, so its branches are still collected.

Resolves #2636


javascript/packages/linter/src/rules/erb-no-unused-literals.ts
```diff
@@ -59,6 +59,25 @@ class LiteralCollector extends PrismVisitor {
   visitBreakNode(): void {}
   visitNextNode(): void {}
 
+  // Skip predicates while continuing traversal into conditional branches and loop bodies.
```
marcoroth:
```suggestion
```

marcoroth (approved):
Sweet, thanks for catching this @viktorianer! 🙏🏼

PR #2655, opened 2026-09-14
Lexer: Recognize Ruby string literals and `#` comments in ERB content

joelhawksley:
`Herb::Engine` currently rejects templates like `<% x = '<%' %>` with `NestedERBTagError` because the ERB lexer treats any `<%` inside tag content as a new tag opener, including `<%` inside a Ruby string literal.

This teaches `lexer_parse_erb_content` (in `src/lexer.c`) to skip past Ruby-source constructs while scanning ERB content, matching Erubi and Ruby's stdlib ERB on this class of template. The nested-tag recovery still fires for genuine unclosed tags.

Fixes the last failing test from #1548.

### What's covered

- Strings: `'...'`, `"..."`, `` `...` `` — escapes, `#{ ... }` interpolation with recursion into nested strings.
- `# ... \n` line comments — terminated by newline, EOF, or `%>` (whichever comes first, so single-line `<% # foo %>` still terminates as before).
- Heredocs: `<<IDENT`, `<<'IDENT'`, `<<"IDENT"`, `` <<`IDENT` ``, plus the `<<~` (squiggly) and `<<-` (dashed) variants. Single-quoted / bareword-single heredocs are literal; the others honor `\` escapes and `#{ ... }` interpolation. Squiggly and dashed variants accept an indented terminator. A small per-tag queue supports multiple heredocs stacked on one line (`foo(<<A, <<B)`).
- `%`-literals: `%q`, `%Q`, `%w`, `%W`, `%i`, `%I`, `%s`, `%x`, and bare `%{}`, `%[]`, `%()`. Paired delimiters (`{} [] () <>`) nest; other non-alphanumeric delimiters are mirrored. Escapes always honored; interpolation applies to `%Q`, `%W`, `%I`, `%x`, and bare forms.
- Regex literals `/.../` with `\` escapes, `#{ ... }` interpolation, and trailing flags `[imxoesun]*`.
- Character literals `?a`, `?\n`, `?\x1b`, `?\u1234`, `?\u{1F600}`, `?\C-a`, `?\M-a`, `?\M-\C-a`.

Unterminated strings, %-literals, and regexes fall back to the existing recovery path via a state snapshot / restore.

### Heuristics

Ambiguous constructs are gated on the last significant character seen in the current ERB tag, which acts as a lightweight expression-vs-value discriminator:

- **`/` is a regex** only after `= ( , [ { ; & | ! ? : < > + - * / % ^ ~ \n \r` or at the start of the ERB content. Otherwise it's division. This keeps `<% x = a / b / c %>` intact.
- **`?` is a character literal** in the same positions. After an identifier, number, or `) ] }` it's a ternary — `<% n = 1 ? a : b %>` lexes normally.
- **Bare `%<delim>`** is restricted to `{`, `[`, `(`. The `<>` form is only accepted with an explicit type letter (`%Q<>` etc.) because `%<` collides with the `-%>`, `=%>`, `%%>` close-tag error-recovery variants (e.g. `<%= x -%<h1>`).

When ambiguous, the lexer errs on the side of *not* skipping, so misdetection cannot swallow legitimate `%>`.

### Snapshot audit

Full suite is green with zero pre-existing snapshot regressions from this commit. The one pre-existing snapshot that changed across the whole PR is `test/snapshots/analyze/render_test/test_0011_render_with_inline_*` — the `<%= render inline: "<%= name %>" %>` template that previously errored with `NestedERBTagError` / `StrayERBClosingTagError` now correctly parses as a single `ERBRenderNode` with `inline_template: "<%= name %>"`. Error-recovery snapshots for genuine unclosed tags (e.g. `<%<% %>` regression) are unchanged.

### Verification

`bundle exec rake test:all`: **3916 runs, 10430 assertions, 0 failures, 0 errors, 58 skips**.


marcoroth (commented):
@joelhawksley I honestly don't think we should be supporting this. I have been looking at supporting it too (to close #1548), but it's just super error prone and pretty much impossible to get right at the lexer level, without making the lexer super complicated.

I feel like wanting to use a literal `<%` in your template is kind of an edge case and sacrificing that for worse diagnostics and way more lexer complexity is not worth it, imho.

marcoroth:
Update: Found a simpler way to achieve the same result: https://github.com/marcoroth/herb/pull/2656

joelhawksley:
@marcoroth great! Glad to see we've closed the gaps with Erubi ahead of the upcoming Rails release. That being said, I think we can expect to find a few more undocumented compatibility edge cases as more people adopt Herb.

marcoroth:
Yes, definitely! We still got some more intentional divergences, but I'd argue that the way Herb handles them is better in all ways. 

That said, any render-time difference is still considered a bug.

PR #2659, opened 2026-09-18
Dev Tools: Fix outlines breaking and littering the page

eglitobias:
Three things go wrong once **View outlines** are on. They sit in the same pair of methods, so they are here together, one commit each.

## The page goes blank

```erb
<html>
  <body><div id="content"><%= yield %></div></body>
</html>
```

```css
#content { position: absolute; top: 40px; bottom: 20px; left: 0; right: 0 }
```

`#content` drops to zero height as soon as the outlines go on. In the app I hit this in, 805px to 4px.

To anchor its label, `createOverlayLabel` sets `position: relative` on the element it outlines. For a layout that element is `<html>`, which then becomes the containing block for everything positioned absolutely below it, and `<html>` has no height in a layout like this one.

`anchorLabel` leaves `<html>` and `<body>` alone. Both already share their origin with the initial containing block, so the label lands in the same place.

## Switching the outlines off doesn't undo it

The inline `position` stayed on every element the overlay had touched. `releaseLabel` takes back what the overlay set, and leaves a position the page set itself alone.

## A label stays on screen

Labels were looked up with `element.querySelector('.herb-overlay-label')`, which searches the whole subtree. With nested outlined elements the outer one removed a descendant's label and kept its own. The lookups are scoped to direct children now, which is what `overlay.css` already assumed with its `> .herb-overlay-label` rules.

---

Five tests in `test/view-outlines.test.ts`, added with the commit they cover. Each commit is green on its own, and undoing any one of the three changes fails its tests.


javascript/packages/dev-tools/src/overlay/overlay.ts
```diff
@@ -837,7 +837,9 @@ export class HerbOverlay {
   }
 
   private createOverlayLabel(element: HTMLElement, type: 'view' | 'partial' | 'component') {
-    if (element.querySelector('.herb-overlay-label')) {
+    // Scoped to direct children: an unscoped lookup finds a nested element's label instead, so
+    // removal takes the wrong one and the outer element keeps its own.
```
marcoroth:
```suggestion
```

javascript/packages/dev-tools/src/overlay/overlay.ts
```diff
@@ -920,18 +920,35 @@ export class HerbOverlay {
...
-    if (window.getComputedStyle(element).position === 'static') {
-      element.style.position = 'relative';
-    }
+    this.anchorLabel(element);
     element.appendChild(label);
   }
 
+  // The label is positioned inside its host, but `<html>` and `<body>` must stay unpositioned:
+  // making either a containing block collapses layouts that place elements against the initial
+  // containing block. Both share its origin, so the label lands in the same spot regardless.
```
marcoroth:
```suggestion
```

javascript/packages/dev-tools/src/overlay/overlay.ts
```diff
@@ -920,18 +920,35 @@ export class HerbOverlay {
...
+  private anchorLabel(host: HTMLElement) {
+    if (host === document.documentElement || host === document.body) return;
+    if (window.getComputedStyle(host).position !== 'static') return;
+
+    host.style.position = 'relative';
+    host.setAttribute('data-herb-debug-anchored', 'true');
+  }
+
+  // Only drop the position the overlay added.
```
marcoroth:
```suggestion
```

javascript/packages/dev-tools/test/view-outlines.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const overlayOf = () => (HerbDevTools.instance as any).overlay
+
+const markAsView = (element: HTMLElement) => {
+  element.setAttribute("data-herb-debug-outline-type", "view")
+  element.setAttribute("data-herb-debug-file-name", "show.html.erb")
+}
+
+describe("view outlines", () => {
+  // A layout whose root box is shorter than the viewport because everything in it is out of flow.
```
marcoroth:
```suggestion
```

PR #2664, opened 2026-09-19
Formatter: Add `--quiet` flag to `herb-format`

arzezak:
Every `herb-format` run printed two lines to stderr with no way to turn them off: the "Using Herb config file" notice from the config loader and the "Experimental Preview" banner. Tools that wrap the formatter (git hooks, editor integrations, CI logs) have no reason to repeat them on every invocation.

`-q, --quiet` skips the banner and loads the config with `silent: true`. Everything else is unchanged: the "Forcing formatter run" warning, the unformatted-file list, the summary, and errors still print.


javascript/packages/formatter/src/cli.ts
```diff
@@ -216,7 +220,7 @@ export class CLI {
         process.exit(0)
       }
 
-      const config = await Config.loadForCLI(configFile || this.projectPath, version)
+      const config = await Config.load(configFile || this.projectPath, { version, silent: isQuietMode })
```
arzezak, line 223: Switched from `Config.loadForCLI` to `Config.load` because `loadForCLI` hardcodes `silent: false` and has no way to pass the quiet flag through. The other options are the same: `createIfMissing` and `exitOnError` both default to `false` in `load`, which is what `loadForCLI` passed.

This also matches how `herb-lint` loads its config: it calls `Config.load` directly with its own `silent` value.

Happy to add a `silent` option to `loadForCLI` instead if you'd rather keep the CLI-specific entry point.

PR #2695, opened 2026-09-20
Ruby: Restore Prism default for starting line count

radanskoric:
## Overview 

This fixes a bug where Prism nodes inside Herb parsed AST have its source location line numbers be 0-based instead of the expected 1-based.

## Background

The `line` option sets the number with which Prism parser will start counting lines. As can be seen here, the default is 1:
https://docs.ruby-lang.org/en/master/Prism.html#method-c-parse

However, when herb initialises the options structure it sets the value to all 0s and then sets only the `partial_script` option.

Since line count of 0 is a valid value, Prism treats it as if the option is set to 0 instead of using the default. Since Herb itself counts lines from 0, this leads to a different line count between the Herb node and it's attached Prism node.

This commit explicitly sets it to 1 to restore the default value.

This also means that it no longer has to correct line counts when formatting location.

## Reproduction

Minimal reproduction of the issue:
```
# frozen_string_literal: true

require 'herb'
require 'prism'

source = File.read(File.join(__dir__, 'herb_line_bug.html.erb'))
result = Herb.parse(source, prism_nodes: true, prism_nodes_deep: true)
node = result.value.children.find { |child| child.is_a?(Herb::AST::ERBContentNode) }

puts "Herb line: #{node.location.start.line}" # 2
puts "Prism line: #{node.prism.location.start_line}" # 1 in old, 2 in new 
puts 'Expected both to be 2.'
```

With the following content in `herb_line_bug.html.erb` next to the script:
```
Test
<%= User.count %>
```


test/parser/prism/prism_nodes_test.rb
```diff
@@ -134,6 +134,33 @@ class PrismNodesTest < Minitest::Spec
...
+        source = "<% if true %>\n<%= User %>\n<% end %>"
+        result = Herb.parse(source, prism_nodes: true, prism_nodes_deep: false)
+        location = result.value.children.first.deserialized_prism_node.location
+
+        assert_equal 1, location.start_line
+        assert_equal 3, location.end_line
+        assert_equal 3, location.start_offset
+        assert_equal 32, location.end_offset
+        assert_equal 3, location.start_column
+        assert_equal 6, location.end_column
```
marcoroth: @radanskoric would you mind switching the two tests for `assert_parsed_snapshot`

radanskoric: Sorry, I forgot about that convention! I've updated it now.

marcoroth (approved):
Thank you @radanskoric! 🙏🏼

